PR #20, opened 2019-08-29
Address failing unit test

cgriggs01:
Hey Team,

This should be the final PR and we can release. 

The reason for this code chance is becuase we run `make test` in the release process which was failing due to env variable checks. I've moved these checks to be inside the TestAcc check call so they are not called during the simple unit test.


arvindsree:
Looks like make testacc works as before

PR #51, opened 2019-09-18
(10735): Update error message when terraform version does not match controller version

zhiqiangjin:
Point to the version construct on Terraform's website for instructions.


arvindsree:
add provider version to controller version specific information

PR #59, opened 2019-09-24
(10832): Fix the issue which is unable to create oci vpn-gw with no elb-name specified

zhiqiangjin:
Also fix 0010833: OCI import issue with "allocate_new_eip"

### DIFF
--- a/aviatrix/resource_aviatrix_gateway.go
+++ b/aviatrix/resource_aviatrix_gateway.go
@@ -615,6 +615,9 @@ func resourceAviatrixGatewayCreate(d *schema.ResourceData, meta interface{}) err
 			ElbName:         d.Get("elb_name").(string),
 			SaveTemplate:    "no",
 		}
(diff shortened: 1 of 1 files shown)


aviatrix/resource_aviatrix_gateway.go
```diff
@@ -615,6 +615,9 @@ func resourceAviatrixGatewayCreate(d *schema.ResourceData, meta interface{}) err
 			ElbName:         d.Get("elb_name").(string),
 			SaveTemplate:    "no",
 		}
+		if sTunnel.ElbName == "" {
```
arvindsree, line 618: Better check is if ELB is enabled

arvindsree (approved):

aviatrix/resource_aviatrix_gateway.go
zhiqiangjin: No, this is for the case that elb is not enabled, then gw_name will be used.

PR #60, opened 2019-09-24
(10757): Increase sleep time from 3min to 4min for vpn user to be active in vpn_user_accelerator

zhiqiangjin:


arvindsree:
Can you do a retry in a for loop instead of a single retry?

PR #100, opened 2019-10-09
(10434): Separate subnets to public and private subnets

zhiqiangjin:
Still keep subnets.


website/docs/r/aviatrix_vpc.html.markdown
```diff
@@ -45,6 +45,15 @@ In addition to all arguments above, the following attributes are exported:
   * `cidr` - CIDR block.
   * `name` - Name of this subnet.
   * `subnet_id` - ID of this subnet.
+* `public_subnets` - List of public subnet of the VPC to be created.
```
arvindsree, line 48: can you add the same info for firenet VPC which index is egress,  management etc

arvindsree (approved):

PR #132, opened 2019-10-23
(11276): Fix the issue where it is unable to disable gateway 'split_tunnel'

zhiqiangjin:


aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1210,21 +1210,20 @@ func resourceAviatrixGatewayUpdate(d *schema.ResourceData, meta interface{}) err
 	if d.HasChange("split_tunnel") || d.HasChange("additional_cidrs") ||
 		d.HasChange("name_servers") || d.HasChange("search_domains") {
 		splitTunnel := d.Get("split_tunnel").(bool)
+		sTunnel := &goaviatrix.SplitTunnel{
+			VpcID:   d.Get("vpc_id").(string),
+			ElbName: d.Get("elb_name").(string),
+		}
+		if sTunnel.ElbName == "" {
```
arvindsree, line 1217: This should be based on gateway ELB status and not name

PR #153, opened 2019-10-29
(11330): Add support of attaching/detaching FireNet gateway to/from TGW Firewall Domain

zhiqiangjin:


goaviatrix/aws_tgw_vpc_attachment.go
```diff
@@ -28,32 +30,76 @@ type DomainListResp struct {
 func (c *Client) CreateAwsTgwVpcAttachment(awsTgwVpcAttachment *AwsTgwVpcAttachment) error {
```
arvindsree: Not sure if this is the right way.
 goaviatrix should have two API's for TGQ VPC attachment and firenet.
 and terraform should call the right API depending on the IsFirewallSecurityDomain output

website/docs/r/aviatrix_controller_config.html.markdown
```diff
@@ -19,9 +19,7 @@ resource "aviatrix_controller_config" "test_controller_config" {
   fqdn_exception_rule        = false
   security_group_management  = true
 }
-```
```
arvindsree, line 22: Don't remove, it is better as separate config

PR #158, opened 2019-11-05
Add multiple retries for 'save' or 'synchronize' in data_source_firenet_vendor_integration

zhiqiangjin:
Try every 5 minutes, and fail it if still failing after six tries.

### DIFF
--- a/aviatrix/data_source_aviatrix_firenet_vendor_integration.go
+++ b/aviatrix/data_source_aviatrix_firenet_vendor_integration.go
@@ -103,7 +103,8 @@ func dataSourceAviatrixFireNetVendorIntegrationRead(d *schema.ResourceData, meta
 		return fmt.Errorf("can't do 'save' and 'synchronize' at the same time for vendor integration")
 	}
 
-	if vendorInfo.Save {
+	isRetry := true
+	if vendorInfo.Save && isRetry {
 		var err error
 		for i := 0; ; i++ {
 			err = client.EditFireNetFirewallVendorInfo(vendorInfo)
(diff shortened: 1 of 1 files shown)


aviatrix/data_source_aviatrix_firenet_vendor_integration.go
```diff
@@ -103,18 +104,34 @@ func dataSourceAviatrixFireNetVendorIntegrationRead(d *schema.ResourceData, meta
...
 
 	if vendorInfo.Save {
-		err := client.EditFireNetFirewallVendorInfo(vendorInfo)
-		if err != nil {
-			d.SetId("")
-			return fmt.Errorf("failed to 'save' FireNet Firewall Vendor Info: %s", err)
+		var err error
+		for i := 0; ; i++ {
+			err = client.EditFireNetFirewallVendorInfo(vendorInfo)
+			if err == nil {
```
arvindsree, line 111: We should only retry for particular errors

aviatrix/data_source_aviatrix_firenet_vendor_integration.go
```diff
@@ -103,18 +104,34 @@ func dataSourceAviatrixFireNetVendorIntegrationRead(d *schema.ResourceData, meta
...
+				break
+			}
+			if i <= 5 {
+				time.Sleep(300 * time.Second)
```
arvindsree, line 115: 5 minute * 5 is too much. Maybe 1 minute * 10

arvindsree:
The retry count should be a parameter. Default should be to not retry

I dont think we should do this. They can add a delay in tf file after the firewall instance creation.

PR #183, opened 2019-11-26
Update test-infra to skip testing for oci, spoke_vpc, and transit_vpc

zhiqiangjin:


arvindsree:
Let the skip variables be no. The end user can set alternately

arvindsree (approved):

PR #231, opened 2020-01-14
(0012247) Implement firewall image version parameter into firewall_instance resource

zhiqiangjin:
This will be available for controller 5.3.

### DIFF
--- a/aviatrix/resource_aviatrix_firewall_instance.go
+++ b/aviatrix/resource_aviatrix_firewall_instance.go
@@ -60,6 +60,12 @@ func resourceAviatrixFirewallInstance() *schema.Resource {
 				ForceNew:    true,
 				Description: "Management Interface Subnet.",
 			},
+			"firewall_image_version": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Version of firewall image.",
+			},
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_firewall_instance.go
```diff
@@ -165,6 +172,7 @@ func resourceAviatrixFirewallInstanceRead(d *schema.ResourceData, meta interface
 	d.Set("firenet_gw_name", fI.GwName)
 	d.Set("firewall_name", fI.FirewallName)
 	d.Set("firewall_image", fI.FirewallImage)
+	d.Set("firewall_image", fI.FirewallImage)
```
arvindsree: firewall_image_version?

arvindsree:
would it cause diff in existing deployments?

aviatrix/resource_aviatrix_firewall_instance.go
zhiqiangjin: Redundant set. Will remove it.

zhiqiangjin:
> would it cause diff in existing deployments?

backend only returns firewall image version when it was provided in creation. So, it will not cause diff in existing deployments.

arvindsree (approved):

PR #254, opened 2020-01-25
Deprecate dnat and snat features in spoke gateway resource

zhiqiangjin:
Also rename "enable_snat" to "single_ip_snat".


arvindsree:
I think we need the tf state migration to set the current value of single_ip_snat from enable_snat

arvindsree (approved):

arvindsree (approved):

PR #266, opened 2020-01-29
(12732) Fix issue where aws_tgw import and them apply fails

zhiqiangjin:


goaviatrix/utils.go
```diff
@@ -74,6 +74,43 @@ func DifferenceSlice(a, b [][]string) [][]string {
 	return ab
 }
 
+// DifferenceSliceAttachedVPC returns the one-dimension elements in two-dimension slice a that aren't in two-dimension b
```
arvindsree: Can you explain more in the comment? 
How is it different from DifferenceSlice

PR #267, opened 2020-01-30
(0012715) Fix the issue where spoke gateway does not handle 'included_advertised_spoke_routes' when HA is created

zhiqiangjin:


zhiqiangjin:
Both primary and HA have to be up for updating "included_advertised_spoke_routes". Solution in Terraform is to add retries to wait for primary and HA to be up instead of returning an error right away.

arvindsree (approved):

PR #284, opened 2020-02-12
(0013097) Add check for cloud_type when incorrectly using ha_subnet for three types if gateway

zhiqiangjin:


arvindsree:
is it possible to add this in the parameter validation function?

arvindsree (approved):

PR #299, opened 2020-02-21
(0013061) Implement add/delete spoke to transit firenet inspection by creating a new resource

zhiqiangjin:


arvindsree:
Isn't it better as a flag in transit gateway?

PR #311, opened 2020-03-02
(13392) Implement support for arm provider in aviatrix_vpc

zhiqiangjin:


arvindsree:
Add an acceptance test case as well

PR #316, opened 2020-03-04
(13438) Implement support of username and password in firewall_instance resource for azure provider

zhiqiangjin:


arvindsree:
We need read for the username for import

arvindsree (approved):

PR #346, opened 2020-03-26
(0013647) Add support for learned_cidrs_approval for aws_tgw_vpn_conn

zhiqiangjin:


aviatrix/resource_aviatrix_aws_tgw_vpn_conn.go
```diff
@@ -203,10 +220,52 @@ func resourceAviatrixAwsTgwVpnConnRead(d *schema.ResourceData, meta interface{})
...
+			err := client.DisableVpnConnectionLearnedCidrsApproval(awsTgwVpnConn)
+			if err != nil {
+				return fmt.Errorf("failed to disable learned cidrs approval: %s", err)
+			}
+		}
+		d.SetPartial("enable_learned_cidrs_approval")
+	}
+
+	d.SetId(awsTgwVpnConn.TgwName + "~" + awsTgwVpnConn.VpnID)
+	return resourceAviatrixAwsTgwVpnConnRead(d, meta)
```
arvindsree, line 266: use the defered read to handle failures

arvindsree (approved):

aviatrix/resource_aviatrix_aws_tgw_vpn_conn.go
zhiqiangjin: Only one thing, which is to create an aws_tgw_vpn_conn, is done in create function. Deferred read is not needed for this case.

arvindsree: What if the disable actions fails but gives a warning and the state is actually false?

You can avoid the additional terraform refresh

arvindsree: It's a good practice to follow everywhere

PR #348, opened 2020-03-27
(0013779) Enhance read back of 'attached_aviatrix_transit_gateway' to cover multiple transit gateways created on the same vpc

zhiqiangjin:


arvindsree (approved):

arvindsree:
Do we need to change the attach code?  Does it use the vpc_id?

PR #361, opened 2020-04-01
Update acceptance test timeout to 10 hours

zhiqiangjin:


arvindsree:
I think you can remove the timeout completely. There is no section that would indefinitely wait

PR #365, opened 2020-04-02
Implement logrus to introduce actual log level distinctions

### LINKED ISSUE #364: Implement logging by level
**Problem**

During an inspection of our logging I ran into sensitive data being logged, which I could trace back to this go-aviatrix package:

```
2020/04/02 01:00:30 [TRACE] POST https://<address>/v1/api Body: action=login&password=<redacted>&username=<redacted>
```
The login and password are logged plaintext, everyone that has access to the logging could use these credentials. 

The current logging implementation tags log lines with a `[LOG_LEVEL]` prepended to the message. This is cool from a visual aspect but this level can't be used by log aggregation tools to filter. So a message logged as TRACE isn't actually logged as trace, it just looks like a trace message.

**Proposed solution**

Using a third-party implementation for handling logs is widely accepted and offers more control over log levels. Using https://github.com/sirupsen/logrus for this, the log level is logged as a separate field, which can be picked up by log aggregation tools to filter out sensitive messages. 

Using `log.SetLevel(log.InfoLevel)` wil limit devoid the logging output from all TRACE and DEBUG messages. While trace logging can be useful while testing, using this approach developers can turn it off in production.

DerkSchooltink:
Fixes #364

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - Supported Terraform version: **v0.12.x**
 
 ### Features:
-1. Implemented support for ``enable_learned_cidrs_approval`` in **aviatrix_transit_gateway**, **aws_tgw_vpn_conn** and **aws_tgw_directconnect**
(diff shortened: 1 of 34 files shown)


DerkSchooltink:
Accidentally removed a bunch of vendor files. Fixed now.

arvindsree (approved):

PR #413, opened 2020-05-13
(0014613) Fix issue where transit external device conn is forced to recreate caused by connection_type not set correctly

zhiqiangjin:


arvindsree:
Use squash and merge

PR #421, opened 2020-05-22
(14717) Refactor to use constants for cloud provider id

CyrusJavan:
- Refactor all uses of integer cloud types in the `aviatrix` and `goaviatrix` package to use constants. This is an easy way to increase readability.

### DIFF
--- a/aviatrix/data_source_aviatrix_gateway.go
+++ b/aviatrix/data_source_aviatrix_gateway.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/terraform-providers/terraform-provider-aviatrix/cloud"
+	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix/cloud"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix"
--- a/aviatrix/data_source_aviatrix_gateway.go
+++ b/aviatrix/data_source_aviatrix_gateway.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/terraform-providers/terraform-provider-aviatrix/cloud"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix"
 )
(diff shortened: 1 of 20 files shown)


cloud/const.go
```diff
@@ -0,0 +1,31 @@
+// cloud package is used to hold cloud provider information that is needed
```
arvindsree: This should be moved under goaviatrix

CyrusJavan: My only issue with that was that `goaviatrix.AWS` is longer/uglier than `cloud.AWS`. But it does save on creating a whole new package just for constants. I will make the change.

arvindsree: What if you import it as goaviatrix/cloud/ ?

zhiqiangjin: Yes, I think so too. This is basically for our internal use.

CyrusJavan: I like that compromise. 

Moved cloud/const.go to goaviatrix/cloud/const.go, this avoids a new package in the base directory and lets us use `cloud.AWS` in the code. 👍

aviatrix/data_source_aviatrix_gateway.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix/cloud"
```
zhiqiangjin: There is no need to import "*/goaviatrix/cloud" since "/goaviatrix" is imported as a whole, right?

arvindsree (approved):

aviatrix/data_source_aviatrix_spoke_gateway.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix"
+	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix/cloud"
```
zhiqiangjin: Same

zhiqiangjin (changes requested):

aviatrix/data_source_aviatrix_gateway.go
CyrusJavan: We have to do it like this since `cloud` is still it's own package it's just located inside goaviatrix/ directory
```
package cloud

// Cloud provider ids
const (
	AWS    = 1
	GCP    = 4
	AZURE  = 8
	OCI    = 16
	AWSGOV = 256
)
```

zhiqiangjin (approved):

PR #429, opened 2020-06-03
(14138) Add route_tables and route_tables_filter to data_source_vpc

CyrusJavan:
- Add 2 new attributes to data source vpc
  - `route_tables` List of AWS route table ids associated with this VPC. Only populated for AWS vpc.
  - `route_tables_filter` Filters the route_tables list to contain only public or private route tables. Valid values are 'private' or 'public'. If not set then route_tables are not filtered.
- Updated docs

### DIFF
--- a/aviatrix/data_source_aviatrix_vpc.go
+++ b/aviatrix/data_source_aviatrix_vpc.go
@@ -5,6 +5,8 @@ import (
 	"log"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix"
 )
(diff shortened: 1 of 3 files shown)


aviatrix/data_source_aviatrix_vpc.go
```diff
@@ -189,6 +206,55 @@ func dataSourceAviatrixVpcRead(d *schema.ResourceData, meta interface{}) error {
...
+		if err != nil {
+			return fmt.Errorf("could not get public vpc route table ids: %v", err)
+		}
+
+		vpc.PublicRoutesOnly = false
+		allRtbs, err := client.GetVpcRouteTableIDs(vpc)
+		if err != nil {
+			return fmt.Errorf("could not get all vpc route table ids: %v", err)
+		}
+
```
zhiqiangjin: Default is to output all route tables. So, move getting public route tables to a function like "getPublicRouteTables" similar to "getPrivateRouteTables" below.

zhiqiangjin (changes requested):

aviatrix/data_source_aviatrix_vpc.go
CyrusJavan: Good idea. I ended up putting all 3 into functions since it made more sense to me that way

zhiqiangjin (approved):

PR #446, opened 2020-06-15
(15461) Fix vpc tracker unmarshal error

CyrusJavan:
The field "inst_num" in the JSON returned by the vpc tracker API is usually a number. However, sometimes it is set to a string `"n/a"` this would cause the unmarshalling error. So, I unmarshal into `interface{}` then check the type at runtime.


zhiqiangjin (approved):

arvindsree:
This should be fixed on the controller side

PR #476, opened 2020-07-10
Fix an overflow issue from HashiCorp side

zhiqiangjin:


arvindsree:
A better way is to make the constant an unsigned int

PR #486, opened 2020-07-17
(15350) Add fqdn_tag_rule resource

CyrusJavan:
- fqdn_domain_name resource
- Updated fqdn resouce domain_names to be Optional and Computed so that there is not an issue when using this new resource
- Updated docs on both fqdn and fqdn_domain_name resources

### DIFF
--- a/aviatrix/resource_aviatrix_fqdn.go
+++ b/aviatrix/resource_aviatrix_fqdn.go
@@ -91,12 +91,25 @@ func resourceAviatrixFQDN() *schema.Resource {
 					},
 				},
 			},
+			"manage_domain_names": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+				Description: "Enable to manage domain name rules in-line. If false, domain name rules must be managed " +
+					"using `aviatrix_fqdn_tag_rule` resources.",
+			},
 		},
 	}
 }
 
 func resourceAviatrixFQDNCreate(d *schema.ResourceData, meta interface{}) error {
 	client := meta.(*goaviatrix.Client)
+
+	_, hasSetDomainNames := d.GetOk("domain_names")
+	enabledInlineDomainNames := d.Get("manage_domain_names").(bool)
+	if hasSetDomainNames && !enabledInlineDomainNames {
+		return fmt.Errorf("manage_domain_names must be set to true to set in-line domain names")
+	}
 
 	fqdn := &goaviatrix.FQDN{
 		FQDNTag:  d.Get("fqdn_tag").(string),
@@ -122,7 +135,7 @@ func resourceAviatrixFQDNCreate(d *schema.ResourceData, meta interface{}) error 
 	flag := false
 	defer resourceAviatrixFQDNReadIfRequired(d, meta, &flag)
 
-	if _, ok := d.GetOk("domain_names"); ok {
+	if hasSetDomainNames && enabledInlineDomainNames {
 		names := d.Get("domain_names").([]interface{})
 		for _, domain := range names {
 			if domain != nil {
@@ -207,6 +220,7 @@ func resourceAviatrixFQDNRead(d *schema.ResourceData, meta interface{}) error {
 		id := d.Id()
 		log.Printf("[DEBUG] Looks like an import, no fqdn tag received. Import Id is %s", id)
 		d.Set("fqdn_tag", id)
+		d.Set("manage_domain_names", true)
 		d.SetId(id)
 	}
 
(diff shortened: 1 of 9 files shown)


aviatrix/resource_aviatrix_fqdn.go
```diff
@@ -62,6 +62,7 @@ func resourceAviatrixFQDN() *schema.Resource {
 			"domain_names": {
 				Type:        schema.TypeList,
 				Optional:    true,
+				Computed:    true,
```
arvindsree: Shouldn't make this computed. We should have a manage_domain_names flag similar to tgw resource

aviatrix/resource_aviatrix_fqdn_domain_name.go
```diff
@@ -0,0 +1,159 @@
...
+	return &schema.Resource{
+		Create: resourceAviatrixFQDNDomainNameCreate,
+		Read:   resourceAviatrixFQDNDomainNameRead,
+		Delete: resourceAviatrixFQDNDomainNameDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"fqdn_tag": {
```
arvindsree: rename to fqdn_tag_name

aviatrix/resource_aviatrix_fqdn_domain_name.go
```diff
@@ -0,0 +1,159 @@
...
+				ForceNew:    true,
+				Description: "FQDN Filter Tag Name to attach this domain.",
+			},
+			"fqdn": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "FQDN.",
+			},
+			"proto": {
```
arvindsree: rename to protocol

goaviatrix/fqdn.go
```diff
@@ -483,3 +488,74 @@ func (c *Client) GetGwFilterTagList(fqdn *FQDN) (*FQDN, error) {
...
+}
+
+func (c *Client) GetFQDNDomainName(fqdn *FQDN) (*FQDN, error) {
+	foundFQDN, err := c.ListDomains(&FQDN{FQDNTag: fqdn.FQDNTag})
+	if err != nil {
+		return nil, fmt.Errorf("could not list fqdn domains: %v", err)
+	}
+	domain := fqdn.DomainList[0]
+	found := false
+	for _, d := range foundFQDN.DomainList {
```
arvindsree, line 594: Same. Better to have a new controller API to do this processing

aviatrix/resource_aviatrix_fqdn.go
CyrusJavan: I'm curious, why is the manage_domain_names flag a better approach than Optional and Computed?

arvindsree: Making it computed will prevent managing domain_name any further on the fqdn resource. They will not be able to edit domain_names if they wish to do so.

goaviatrix/fqdn.go
CyrusJavan: Opened mantis [16320](https://mantis.aviatrix.com/mantisbt/view.php?id=16320) proposing the new APIs for controller to do this processing.

aviatrix/resource_aviatrix_fqdn.go
CyrusJavan: 👍

goaviatrix/fqdn.go
arvindsree: Sounds good. We can keep the current code till the API is available

CyrusJavan:
Renamed to `aviatrix_fqdn_tag_rule` per Sherry's request.

goaviatrix/fqdn.go
```diff
@@ -557,3 +562,74 @@ func (c *Client) DisableFQDNPassThrough(gw *Gateway) error {
...
+	}{
+		Action:   action,
+		CID:      c.CID,
+		TagName:  fqdn.FQDNTag,
+		Policies: string(policies),
+	}, func(action string, reason string, ret bool) error {
+		if !ret {
+			// Tried to delete a rule that did not exist, we don't need to fail the apply.
+			if strings.Contains(reason, "the following rules were not found") {
+				return nil
```
arvindsree: We should not probably return ErrNotFound here and ignore the error for destroy code.

The acceptance usually checks for it to confirm if it has been destroyed

CyrusJavan: > ...should not probably return ErrNotFound here...

Do you mean we _should_ return ErrNotFound then just ignore it when calling the function?

arvindsree: Yeah..that's what I meant. This current approach works as well.

arvindsree (approved):

PR #488, opened 2020-07-20
(15350) Add firewall_policy resource

CyrusJavan:
- firewall_policy resource
- Updated firewall resource policies to be Optional and Computed so that there is not an issue when using this new resource
- Updated docs on both firewall and firewall_policy resources

### DIFF
--- a/aviatrix/resource_aviatrix_firewall.go
+++ b/aviatrix/resource_aviatrix_firewall.go
@@ -36,6 +36,13 @@ func resourceAviatrixFirewall() *schema.Resource {
 				Default:     false,
 				Description: "Indicates whether enable logging or not. Valid values: true or false.",
 			},
+			"manage_firewall_policies": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+				Description: "Enable to manage firewall policies via in-line rules. If false, policies must be managed " +
+					"using `aviatrix_firewall_policy` resources.",
+			},
 			"policy": {
 				Type:        schema.TypeList,
 				Optional:    true,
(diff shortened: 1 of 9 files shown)


goaviatrix/firewall.go
```diff
@@ -164,3 +164,81 @@ func PolicyToMap(p *Policy) map[string]interface{} {
...
+			fw.PolicyList = []*Policy{p}
+			break
+		}
+	}
+	if !found {
+		return nil, ErrNotFound
+	}
+
+	return fw, nil
+}
```
arvindsree, line 249: Better to have an API on the controller to do this processing

aviatrix/resource_aviatrix_firewall.go
```diff
@@ -39,6 +39,7 @@ func resourceAviatrixFirewall() *schema.Resource {
 			"policy": {
 				Type:        schema.TypeList,
 				Optional:    true,
+				Computed:    true,
```
arvindsree: We should have a manage_firewall policies flag rather than making this computed

goaviatrix/firewall.go
```diff
@@ -164,3 +164,81 @@ func PolicyToMap(p *Policy) map[string]interface{} {
...
+
+func (c *Client) GetFirewallPolicy(fw *Firewall) (*Firewall, error) {
+	foundFirewall, err := c.GetPolicy(fw)
+	if err != nil {
+		return nil, fmt.Errorf("could not list firewall rules: %v", err)
+	}
+	rule := fw.PolicyList[0]
+	found := false
+
+	for _, p := range foundFirewall.PolicyList {
```
arvindsree, line 228: Same. better to do this processing on the controller

goaviatrix/firewall.go
CyrusJavan: Opened mantis [16320](https://mantis.aviatrix.com/mantisbt/view.php?id=16320) proposing the new APIs for controller to do this processing.

goaviatrix/firewall.go
```diff
@@ -164,3 +169,81 @@ func PolicyToMap(p *Policy) map[string]interface{} {
...
+	}{
+		Action: action,
+		CID:    c.CID,
+		GwName: fw.GwName,
+		Rules:  string(rules),
+	}, func(action string, reason string, ret bool) error {
+		if !ret {
+			// Tried to delete a rule that did not exist, we don't need to fail the apply.
+			if strings.Contains(reason, "Empty rules in Stateful Firewall to delete") {
+				return nil
```
arvindsree, line 216: Similar comment to one in the other PR .You could ignore the error where DeleteFirewallPolicy is being called

arvindsree (approved):

PR #546, opened 2020-08-27
0016823: Initial implementation of support for updating gw_name in aviatrix_firewall

zhiqiangjin:


aviatrix/resource_aviatrix_firewall.go
```diff
@@ -363,7 +383,7 @@ func resourceAviatrixFirewallUpdate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	d.Partial(false)
-	return nil
+	return resourceAviatrixFirewallRead(d, meta)
```
arvindsree, line 386: use the deferred read since you have other exit points above

aviatrix/resource_aviatrix_firewall.go
```diff
@@ -273,7 +282,18 @@ func resourceAviatrixFirewallUpdate(d *schema.ResourceData, meta interface{}) er
...
 	d.Partial(true)
 
-	log.Printf("[INFO] Creating Aviatrix firewall: %#v", firewall)
+	if d.HasChange("gw_name") {
+		_, gwNameNew := d.GetChange("gw_name")
+		gateway := &goaviatrix.Gateway{
+			GwName:         gwNameNew.(string),
+			GwOriginalName: d.Get("gw_original_name").(string),
+		}
+		err := client.IsGatewayNameUpdatable(gateway)
```
arvindsree, line 291: What does this function do? Could you add a short description at its definition?

zhiqiangjin: This function is to check what value the gw_name is changed to. If it is updated to the currently used alias, it is OK. Otherwise, it is not allowed. Will add a short description and figure out a better name for it.

aviatrix/resource_aviatrix_firewall.go
zhiqiangjin: Deferred function will be evaluated, maybe will need to be added to other resources as well.

arvindsree (approved):

PR #557, opened 2020-09-09
0017395: Implement support of TGW in AWS GOV cloud

zhiqiangjin:


aviatrix/resource_aviatrix_aws_tgw.go
```diff
@@ -141,6 +141,13 @@ func resourceAviatrixAWSTgw() *schema.Resource {
 					},
 				},
 			},
+			"cloud_type": {
```
CyrusJavan, line 144: Should this be `ForceNew`? If not, do we need to modify the Update function as well?

inc1t3Ful: If this was ForceNew, existing TGWs would be deleted; need to be mindful of backward-compatibility, though I totally understand why ForceNew would be useful here.

zhiqiangjin: Yes, I added a commit to block updating cloud_type.

aviatrix/resource_aviatrix_aws_tgw.go
```diff
@@ -141,6 +141,13 @@ func resourceAviatrixAWSTgw() *schema.Resource {
...
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      1,
+				Description:  "Type of cloud service provider, requires an integer value. Supported for AWS (1) and AWS GOV (256). Default value: 1.",
+				ValidateFunc: validation.IntInSlice([]int{1, 256}),
```
CyrusJavan:
```suggestion
				ValidateFunc: validation.IntInSlice([]int{goaviatrix.AWS, goaviatrix.AWSGOV}),
```

Could use the cloud type constants here.

CyrusJavan (approved):
Looks good to me.

PR #589, opened 2020-10-02
0017164: Implement a new resource to support attaching spoke to transit with advanced option

zhiqiangjin:


goaviatrix/spoke_transit_attachment.go
```diff
@@ -0,0 +1,97 @@
...
+	bodyString := buf.String()
+	bodyIoCopy := strings.NewReader(bodyString)
+	if err = json.NewDecoder(bodyIoCopy).Decode(&data); err != nil {
+		return errors.New("Json Decode attach_spoke_to_transit_gw failed: " + err.Error() + "\n Body: " + bodyString)
+	}
+	if !data.Return {
+		return errors.New("Rest API attach_spoke_to_transit_gw Post failed: " + data.Reason)
+	}
+	return nil
+}
```
CyrusJavan: This function could be simplified with the `PostAPI` function

```
action := "attach_spoke_to_transit_gw"
spokeTransitAttachment.CID = c.CID
spokeTransitAttachment.Action = action
return c.PostAPI(action, spokeTransitAttachment, BasicCheck)
```

goaviatrix/spoke_transit_attachment.go
```diff
@@ -0,0 +1,97 @@
...
+			spokeTransitAttachment.RouteTables = strings.Join(data.Results.RouteTables, ",")
+			return spokeTransitAttachment, nil
+		}
+	}
+
+	log.Errorf("Couldn't find Aviatrix gateway %s", spokeTransitAttachment.SpokeGwName)
+	return nil, ErrNotFound
+}
+
+func (c *Client) DeleteSpokeTransitAttachment(spokeTransitAttachment *SpokeTransitAttachment) error {
```
CyrusJavan, line 64: This function also could be simplified like the Create function using `PostAPI` function.

CyrusJavan:
We also need a new attribute on the spoke_gw, like `manage_transit_attachments`. Should that be in this PR or another one after?

inc1t3Ful:
Definitely should be in this PR, as it pertains the implementation of 17164

CyrusJavan (approved):
Looks good to me

PR #608, opened 2020-10-19
modify vpn config

ghost:


ghost:
@CyrusJavan, VPN config (idle timeout and renegotiation interval).

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -385,6 +385,30 @@ func resourceAviatrixGateway() *schema.Resource {
 				DiffSuppressFunc: DiffSuppressFuncString,
 				Description:      "A list of monitored instance ids separated by comma when 'monitor gateway subnets' feature is enabled.",
 			},
+			"idle_timeout": {
```
CyrusJavan, line 389: Why do we need both a boolean field and value field? Isn't the value field alone enough?
I see two situations:
1)  User supplies a value, assume boolean is true
2) User does not supply value, assume boolean is false

ghost:
The attribute is reorganized, only idle_timeout_value and renegotiation_interval_value are used for this feature.
These two attributes are integer type, and the default value is -1. If these attributes are set and the value is greater than 300, these feature are enabled. Otherwise, they will be disabled.

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -385,6 +386,20 @@ func resourceAviatrixGateway() *schema.Resource {
 				DiffSuppressFunc: DiffSuppressFuncString,
 				Description:      "A list of monitored instance ids separated by comma when 'monitor gateway subnets' feature is enabled.",
 			},
+			"idle_timeout_value": {
```
CyrusJavan: I think `idle_timeout` would be a better name. Adding `_value` to the end is unnecessary.

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -385,6 +386,20 @@ func resourceAviatrixGateway() *schema.Resource {
...
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Default:      -1,
+				ValidateFunc: validation.IntAtLeast(301),
+				Description:  "Typed value when modifying idle_timeout. If it's -1, this feature is disabled.",
+			},
+			"renegotiation_interval_value": {
```
CyrusJavan: Same as above, let's drop the `_value`. Remember to update the documentation too

website/docs/r/aviatrix_gateway.html.markdown
```diff
@@ -211,6 +211,10 @@ The following arguments are supported:
...
 * `monitor_exclude_list` - (Optional) A list of monitored instance IDs separated by comma when Monitor Gateway Subnets feature is enabled. Default value is "". Available as of provider version R2.18+.
 
+### Modify VPN Configuration
+* `idle_timeout_value` - (Optional) It sets the value (seconds) of the [idle timeout](https://docs.aviatrix.com/HowTos/openvpn_faq.html#how-do-i-fix-the-aviatrix-vpn-timing-out-too-quickly). This idle timeout feature is enable only if this attribute is set, otherwise it is disabled. The entered value must be an integer number greater than 300.  Available as of provider version R2.18+.
+* `renegotiation_interval_value` - (Optional) It sets the value (seconds) of the [renegotiation interval](https://docs.aviatrix.com/HowTos/openvpn_faq.html#how-do-i-fix-the-aviatrix-vpn-timing-out-too-quickly). This renegotiation interval feature is enable only if this attribute is set, otherwise it is disabled. The entered value must be an integer number greater than 300. Available as of provider version R2.18+.
```
CyrusJavan: Move these under `### VPN Access` above

goaviatrix/gateway.go
```diff
@@ -1161,3 +1173,71 @@ func (c *Client) DisableMonitorGatewaySubnets(gateway *Gateway) error {
...
+func (c *Client) DisableVPNConfig(gateway *Gateway, vpnConfig *VPNConfig) error {
+	action := "edit_vpn_config"
+	form := map[string]interface{}{
+		"CID":     c.CID,
+		"action":  action,
+		"command": "disable",
+		"vpc_id":  gateway.VpcID,
+		"lb_name": gateway.GwName,
+		"key":     vpnConfig.Name,
+		"value":   "600",
```
CyrusJavan: Why "600" is needed here? Can we leave this out?

goaviatrix/gateway.go
```diff
@@ -1161,3 +1173,71 @@ func (c *Client) DisableMonitorGatewaySubnets(gateway *Gateway) error {
...
+	if err = json.NewDecoder(bodyIoCopy).Decode(&data); err != nil {
+		return nil, errors.New("Json Decode edit_vpn_config(show) failed: " + err.Error() + "\n Body: " + bodyString)
+	}
+	if !data.Return {
+		return nil, errors.New("Rest API edit_vpn_config(show) Get failed: " + data.Reason)
+	}
+	return data.Results, ErrNotFound
+}
+
+func (c *Client) GetVPNConfig(vpnConfigName string, vpnConfigList []VPNConfig) *VPNConfig {
```
CyrusJavan: This helper function should not be an exported function of the `goaviatrix` package. The `goaviatrix` package is for functions that wrap API calls. If you just want a little helper function like this then define it in the `aviatrix` package in the same file where you use it.

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1156,6 +1203,46 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
...
+				if err != nil {
+					return fmt.Errorf("couldn't convert vpn config value (renegotiation interval)")
+				}
+				d.Set("renegotiation_interval_value", renegoIntervalValue)
+			} else {
+				d.Set("renegotiation_interval_value", -1)
+			}
+		} else {
+			return fmt.Errorf("couldn't find vpn config (renegotiation interval) for gateway: %s", gw.GwName)
+		}
```
CyrusJavan: This is just a nitpick not a big deal but these if statements can be simplified, check out this article: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 . Reducing the amount of indentation levels can increase readability of the code

goaviatrix/gateway.go
ghost: API to disable the feature requires the value input. This value is hard code. The updated RP use "-1" as value input, which will be more consistent to our attribute design.

CyrusJavan (approved):

PR #615, opened 2020-10-22
18216: Add firenet 2-Tuple forwarding algorithm support

arvindsree:
Add firenet 2-Tuple forwarding algorithm support in firenet resource and data source

### DIFF
--- a/goaviatrix/firenet.go
+++ b/goaviatrix/firenet.go
@@ -14,8 +14,9 @@ type FireNet struct {
 	VpcID            string `form:"vpc_id,omitempty" json:"vpc_id,omitempty"`
 	GwName           string `form:"gw_name,omitempty" json:"gw_name,omitempty"`
 	FirewallInstance []FirewallInstance
-	FirewallEgress   bool `form:"firewall_egress,omitempty" json:"firewall_egress,omitempty"`
-	Inspection       bool `form:"inspection,omitempty" json:"inspection,omitempty"`
+	FirewallEgress   bool   `form:"firewall_egress,omitempty" json:"firewall_egress,omitempty"`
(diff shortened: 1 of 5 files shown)


goaviatrix/firenet.go
```diff
@@ -365,3 +367,36 @@ func (c *Client) EditFireNetEgress(fireNet *FireNet) error {
...
+	}
+	editFireNet := url.Values{}
+	editFireNet.Add("CID", c.CID)
+	editFireNet.Add("action", "edit_firenet")
+	editFireNet.Add("vpc_id", fireNet.VpcID)
+	editFireNet.Add("firewall_hashing", fireNet.HashingAlgorithm)
+
+	Url.RawQuery = editFireNet.Encode()
+	resp, err := c.Get(Url.String(), nil)
+	if err != nil {
```
CyrusJavan: This function can be simplified, use a PostAPI function with a custom checkFunc. This way we are not repeating the same JSON decoding logic

```
func (c *Client) EditFireNetHashingAlgorithm(fireNet *FireNet) error {
	action := "edit_firenet"
	data := map[string]string{
		"action":           "edit_firenet",
		"CID":              c.CID,
		"vpc_id":           fireNet.VpcID,
		"firewall_hashing": fireNet.HashingAlgorithm,
	}
	checkFunc := func(act, reason string, ret bool) error {
		if !ret {
			if strings.Contains(reason, "configuration not changed") {
				return nil
			}
		}
		return nil
	}

	return c.PostAPI(action, data, checkFunc)
}
```

arvindsree: done

CyrusJavan (approved):
looks good to me

PR #617, opened 2020-10-23
18237: Change 'management_subnet' to optional for Check Point and Fortinet instances

zhiqiangjin:


CyrusJavan:
I think we also need to update the firewall_instance docs to change Required to Optional

CyrusJavan (approved):
Nice job 👍

Danny-Cooke:
thanks guys

PR #624, opened 2020-10-28
16265: [SAML Endpoint] Add support for Sign Authn Requests

arvindsree:
- Add support for signing AuthnRequests in saml_endpoint resource
 - Remove the double translation in the GetSamlEndpoint function

### DIFF
--- a/aviatrix/resource_aviatrix_saml_endpoint.go
+++ b/aviatrix/resource_aviatrix_saml_endpoint.go
@@ -49,6 +49,7 @@ func resourceAviatrixSamlEndpoint() *schema.Resource {
 				Optional:    true,
 				Default:     false,
 				Description: "Switch to differentiate if it is for controller login.",
+				ForceNew:    true,
 			},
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_saml_endpoint.go
```diff
@@ -151,6 +155,8 @@ func resourceAviatrixSamlEndpointRead(d *schema.ResourceData, meta interface{})
 				log.Printf("[WARN] Error setting 'rbac_groups' for (%s): %s", d.Id(), err)
 			}
 		}
+	} else {
+		d.Set("access_set_by", []string{})
```
CyrusJavan: "access_set_by" is TypeString but here you are trying to set it to slice of string. Is that intended?

arvindsree: Good catch.. It should be rbac_groups. corrected it

CyrusJavan (approved):

PR #629, opened 2020-10-29
Fixing return error issue

ghost:
Fixing return error issue when the enable_transit_firenet attribute is set to false


CyrusJavan:
Can we at least log it when we get an error?

CyrusJavan (approved):

PR #640, opened 2020-11-02
remove duplication in firenet markdown

ghost:
remove the 'manage_firewall_instance_association' duplication in markdown


CyrusJavan (approved):
Good find! Thanks for fixing this

PR #643, opened 2020-11-03
18404: Support IDP Metadata URLs for SAML endpoints

arvindsree:
Support IDP metadata URLs in the saml_endpoint resource

### DIFF
--- a/aviatrix/resource_aviatrix_saml_endpoint.go
+++ b/aviatrix/resource_aviatrix_saml_endpoint.go
@@ -41,11 +41,12 @@ func resourceAviatrixSamlEndpoint() *schema.Resource {
 				Description: "IDP Metadata.",
 			},
 			"idp_metadata_url": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Default:      "",
-				Description:  "IDP Metadata.",
-				ValidateFunc: validation.IsURLWithHTTPorHTTPS,
+				Type:          schema.TypeString,
+				Optional:      true,
+				Default:       "",
+				Description:   "IDP Metadata.",
+				ValidateFunc:  validation.IsURLWithHTTPorHTTPS,
+				ConflictsWith: []string{"idp_metadata"},
 			},
 			"custom_entity_id": {
 				Type:        schema.TypeString,
--- a/aviatrix/resource_aviatrix_saml_endpoint.go
+++ b/aviatrix/resource_aviatrix_saml_endpoint.go
@@ -47,6 +47,7 @@ func resourceAviatrixSamlEndpoint() *schema.Resource {
 				Description:   "IDP Metadata.",
 				ValidateFunc:  validation.IsURLWithHTTPorHTTPS,
 				ConflictsWith: []string{"idp_metadata"},
+				AtLeastOneOf:  []string{"idp_metadata", "idp_metadata_url"},
 			},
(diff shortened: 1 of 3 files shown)


CyrusJavan:
Need to rebase to `future` and change target merge branch to `future` as well.

arvindsree:
> Need to rebase to `future` and change target merge branch to `future` as well.

This is compatible with 6.2. Future is for 6.3  right?

CyrusJavan:
> > Need to rebase to `future` and change target merge branch to `future` as well.
> 
> This is compatible with 6.2. Future is for 6.3 right?

No, sorry confusing naming.

`future` is targeting 2.17.1/6.2-patch

`future6.3` is targeting 2.18/6.3

website/docs/r/aviatrix_saml_endpoint.html.markdown
```diff
@@ -19,7 +19,15 @@ For details on Aviatrix Controller Login with SAML authentication, please see do
...
-  idp_metadata      = "${var.idp_metadata}"
+  idp_metadata      = file("idp_metadata.xml")
+}
+```
+```hcl
+# Create an Aviatrix AWS SAML Endpoint using Metadata UDL
+resource "aviatrix_saml_endpoint" "test_saml_endpoint" {
+  endpoint_name     = "saml-test"
+  idp_metadata_type = "URL"
+  idp_metadata      = "https://dev-xyzz.okta.com/app/asdfasdfwfwf/sso/saml/metadata"
```
CyrusJavan: Should this be `idp_metadata_url` instead of `idp_metadata`?

aviatrix/resource_aviatrix_saml_endpoint.go
```diff
@@ -29,15 +29,24 @@ func resourceAviatrixSamlEndpoint() *schema.Resource {
...
+				ValidateFunc: validation.StringInSlice([]string{"URL", "Text"}, false),
 			},
 			"idp_metadata": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
+				Default:     "",
 				Description: "IDP Metadata.",
 			},
+			"idp_metadata_url": {
```
CyrusJavan, line 43:
```
	// ConflictsWith is a set of schema keys that conflict with this schema.
	// This will only check that they're set in the _config_. This will not
	// raise an error for a malfunctioning resource that sets a conflicting
	// key.
	ConflictsWith []string
```
You can define that idp_metadata ConflictsWith idp_metadata_url. So at _plan_ time you get an error.

CyrusJavan: AtLeastOneOf also looks like it would apply to idp_metadata and idp_metadata_url
```
	// AtLeastOneOf is a set of schema keys that, when set, at least one of
	// the keys in that list must be specified.
```

CyrusJavan (approved):
Besides the above comments, this looks good to me.

aviatrix/resource_aviatrix_saml_endpoint.go
arvindsree: Looks like AtLeastOneOf is a better one as one or the other is required

arvindsree: Actually both need to be used

PR #644, opened 2020-11-03
17443: fix gcp vpc update, use ForceNew

CyrusJavan:


CyrusJavan:
We need to discuss what behavior we want to see in the vpc resource. There are 2 options to choose from:
1. Current behavior. When you go to update subnets values Terraform will stop you with `Error: doesn't support update` 
2. Proposed behavior in this PR. When you go to update subnets values, Terraform will show you that you are performing a replacement operation, and would try to tear down the VPC and bring back up.

CyrusJavan:
As per discussion in today's meeting, we will merge this to make the resource's behavior uniform among all its attributes (all ForceNew).

PR #655, opened 2020-11-17
12272: implement remote syslog

shuoliugit:


aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,158 @@
...
+		Create: resourceAviatrixRemoteSyslogCreate,
+		Read:   resourceAviatrixRemoteSyslogRead,
+		Update: resourceAviatrixRemoteSyslogUpdate,
+		Delete: resourceAviatrixRemoteSyslogDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"index": {
```
CyrusJavan, line 28: You can utilize the ValidateFunc field and the `validation` helper library for attributes that need simple validation like this one. Do a Cmd-Shift-F search for `ValidateFunc` to see some examples. For this case you would use

```
ValidateFunc: validation.IntBetween(0, 9),
```

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,158 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Server Public Certificate: Public certificate of the controller signed by the same CA",
+			},
+			"private_key": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Server Private Key: Private key of the controller that pairs with the public certificate",
+			},
+			"protocol": {
```
CyrusJavan, line 61: You can use a ValidateFunc here to verify the string is TCP or UDP

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,158 @@
...
+					"List all excluded gateways.",
+			},
+			"excluded_gateway": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "List of excluded gateways.",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
```
CyrusJavan: These mean the same thing right? Also you should use a `schema.TypeSet` since we don't care about the order of the list

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,158 @@
...
+		CaCertificate:       d.Get("ca_certificate").(string),
+		PublicCertificate:   d.Get("public_certificate").(string),
+		PrivateKey:          d.Get("private_key").(string),
+		ExcludeGatewayInput: d.Get("exclude_gateway_list").(string),
+	}
+	if remoteSyslog.Index < 0 || remoteSyslog.Index > 9 {
+		return fmt.Errorf("please use an integer from 0 to 9 as the index")
+	}
+	if !((len(remoteSyslog.CaCertificate) > 0 && len(remoteSyslog.PublicCertificate) > 0 && len(remoteSyslog.PrivateKey) > 0) ||
+		(len(remoteSyslog.CaCertificate) == 0 && len(remoteSyslog.PublicCertificate) == 0 && len(remoteSyslog.PrivateKey) == 0)) {
```
CyrusJavan: Just use `remoteSyslog.CaCertificate == ""` or `remoteSyslog.CaCertificate != ""` instead of using the calls to `len`. Save `len` for when we are dealing with slices/arrays or when we actually care about the length of the string

goaviatrix/remote_syslog.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (c *Client) GetRemoteSyslogStatus(idx int) (*RemoteSyslogResp, error) {
+	params := map[string]string{
+		"action": "get_remote_syslog_logging_status",
+		"CID":    c.CID,
+		"index":  strconv.Itoa(idx),
+	}
+
+	type Resp struct {
+		Return  bool             `json:"return,omitempty"`
```
CyrusJavan: The `omitempty` tag is only for _encoding_ , in this case you are only using this struct for _decoding_ so the omitempty does nothing.

goaviatrix/remote_syslog.go
```diff
@@ -0,0 +1,121 @@
...
+			Path:      r.PublicCertificate,
+			ParamName: "public_certificate",
+		},
+		{
+			Path:      r.PrivateKey,
+			ParamName: "private_key",
+		},
+	}
+
+	resp, err := c.PostFile(c.baseURL, params, files)
```
CyrusJavan: Check out `PostFileAPI` function, it is a helper to avoid rewriting the response decoding and error checking

website/docs/r/aviatrix_remote_syslog.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+* `server` (Required) Server IP.
+* `port` (Required) Port number.
+* `protocol` (Optional) TCP or UDP. TCP by default.
+
+### Optional
+* `ca_certificate` (Optional) Certificate Authority (CA) certificate. `ca_certificate`, `public_certificate` and `private_key` must be used together.
+* `public_certificate` (Optional) Public certificate of the controller signed by the same CA.
+* `private_key` (Optional) Private key of the controller that pairs with the public certificate.
+* `template` (Optional) Optional custom template.
+* `exclude_gateway_list` (Optional) List of gateways to be excluded from logging. e.g.: "gateway01", "gateway01, gateway01-hagw". Use a comma delimited string.
```
CyrusJavan: Be sure to add all the attributes to the documentation, seems `excluded_gateway` is missing

website/docs/r/aviatrix_remote_syslog.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+The following arguments are supported:
+
+### Required
+* `index` (Optional) Profile index. An index from 0 to 9 is supported. 0 by default.
+* `server` (Required) Server IP.
+* `port` (Required) Port number.
+* `protocol` (Optional) TCP or UDP. TCP by default.
+
+### Optional
+* `ca_certificate` (Optional) Certificate Authority (CA) certificate. `ca_certificate`, `public_certificate` and `private_key` must be used together.
```
CyrusJavan: It needs to be more clear that these are file paths not the content of the certificates. Could we add a suffix like `_file_path` to the certificate attribute names. Also make sure the documentation is clear that this is a file path.

CyrusJavan (changes requested):

CyrusJavan:
Also, could you please rebase onto the latest `future6.3` branch? That should fix the linter CI issue

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,161 @@
...
+		excludedGateways = append(excludedGateways, v)
+	}
+	if err := d.Set("exclude_gateway", excludedGateways); err != nil {
+		return fmt.Errorf("could not set excluded_gateway: %v", err)
+	}
+
+	d.SetId("remote_syslog_" + remoteSyslogStatus.Index)
+	return nil
+}
+
```
CyrusJavan: Your `Read` function needs to `d.Set` every attribute except for the file path attributes since you cannot retrieve those values from the API. 

Also, in the `Read` we need to handle the case of `import` this current implementation failed to import correctly for me. Check out some other resource's `Read` functions, they will often have some logic at the top of the function to handle importing.

aviatrix/resource_aviatrix_remote_syslog_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix"
+)
+
+func TestAccAviatrixRemoteSyslog_basic(t *testing.T) {
```
CyrusJavan, line 14: Make sure to run the acceptance tests yourself, this test should have caught the `import` functionality not working.

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,161 @@
...
+				Default:      "TCP",
+				ValidateFunc: validation.StringInSlice([]string{"TCP", "UDP"}, false),
+				Description:  "Protocol: TCP or UDP (TCP by default)",
+			},
+			"template": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Optional Custom Template: Useful when forwarding to 3rd party servers like Datadog or Sumo",
+			},
+			"exclude_gateway": {
```
CyrusJavan: Since this is a collection of items the name should be plural. Please use something like `excluded_gateways`

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,179 @@
...
+		Create: resourceAviatrixRemoteSyslogCreate,
+		Read:   resourceAviatrixRemoteSyslogRead,
+		Update: resourceAviatrixRemoteSyslogUpdate,
+		Delete: resourceAviatrixRemoteSyslogDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"index": {
```
CyrusJavan, line 28: Shouldn't `index` be ForceNew? For example, if I change the `index` from 0 to 3. I would expect my config for index 0 would be removed, instead you end up with both 0 and 3 configured.

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,179 @@
...
+	d.Set("port", port)
+	d.Set("protocol", remoteSyslogStatus.Protocol)
+	d.Set("template", remoteSyslogStatus.Template)
+	d.Set("notls", remoteSyslogStatus.Notls)
+	d.Set("status", remoteSyslogStatus.Status)
+
+	var excludedGateways []interface{}
+	for _, v := range remoteSyslogStatus.ExcludedGateways {
+		excludedGateways = append(excludedGateways, v)
+	}
```
CyrusJavan: I don't think it is necessary to transform the type to `[]interface{}`, you should be able to `d.Set` a slice of string no problem.

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,179 @@
...
+
+func resourceAviatrixRemoteSyslogRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*goaviatrix.Client)
+
+	server := d.Get("server").(string)
+
+	if server == "" {
+		id := d.Id()
+		log.Printf("[DEBUG] Looks like an import. Import Id is %s", id)
+		index, _ := strconv.Atoi(id[len(id)-1:])
```
CyrusJavan, line 144: You should check this error, the `d.Id()` is a user provided value, there is no guarantee the user used the correct ID format from your documentation. We should return an error like: 
```
fmt.Errorf("invalid ID format, expected ID in format 'remote_syslog_{index} instead got %s", id)
```

Also should check the length of the ID before trying to extract the last character. If ID was an empty string, this would crash.

aviatrix/resource_aviatrix_remote_syslog_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+		resp, _ := client.GetRemoteSyslogStatus(index)
+		if !sliceMatch(resp.ExcludedGateways, input) {
+			return fmt.Errorf("excluded gateways don't match with the input")
+		}
+		return nil
+	}
+}
+
+func sliceMatch(a, b []string) bool {
```
CyrusJavan: Nit: could also use existing `func Equivalent(a, b []string) bool {`

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_remote_syslog.go
```diff
@@ -0,0 +1,180 @@
...
+func resourceAviatrixRemoteSyslogRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*goaviatrix.Client)
+
+	server := d.Get("server").(string)
+
+	if server == "" {
+		id := d.Id()
+		log.Printf("[DEBUG] Looks like an import. Import Id is %s", id)
+
+		match, _ := regexp.Match("\\bremote_syslog_[0-9]\\b", []byte(id))
```
CyrusJavan: This error should be checked if you want to use `regexp.Match`, however, the better way is to define a global variable like this 
```
var remoteSyslogMatcher = regexp.MustCompile("\\bremote_syslog_[0-9]\\b")
```
In Go, typically functions named `MustXXX` do not return error, instead they will panic if anything went wrong. In this case the function will panic if the regex wont compile, and since the variable is global scoped if you use an invalid regex it will crash on startup (making it easy to spot that the regex is wrong). 

Then with this `remoteSyslogMatcher` we can use it to check the ID format without having to worry about an error happening while compiling the regex. (also more efficient since we only compile the regex once)
```
match := remoteSyslogMatcher.Match([]byte(id))
```

CyrusJavan (changes requested):

CyrusJavan (approved):
Good job on the first PR!

PR #657, opened 2020-11-18
0018294: Add fix to cover the special user case where remote_gateway_ip is two IPs combined

zhiqiangjin:


CyrusJavan (approved):
Looks good to me!

PR #661, opened 2020-11-20
18662: implemented splunk logging

shuoliugit:


aviatrix/resource_aviatrix_splunk_logging.go
```diff
@@ -0,0 +1,146 @@
...
+				Optional:    true,
+				Description: "Server IP",
+			},
+			"port": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     -1,
+				Description: "Port number",
+			},
+			"cu_output_cfg_file_path": {
```
arvindsree: We should support file contents rather than file path. 
1. File contents being changed cannot be identified since there is no diff in filename.
2. They could store the configuration in locations like S3 and the file name cannot be passed here

Customers can use the terraform file function to read the file contents

aviatrix/resource_aviatrix_splunk_logging.go
```diff
@@ -0,0 +1,146 @@
...
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"server": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Server IP",
+			},
+			"port": {
```
CyrusJavan, line 28: What is the purpose of having a default value of `-1`?

aviatrix/resource_aviatrix_splunk_logging.go
shuoliugit: modified to support file contents, but currently there's bug in backend(exception when using configuration file), so will test later

aviatrix/resource_aviatrix_splunk_logging.go
shuoliugit: rest api will return err when port is 0, so I modified the code to simply check if the port is 0 or not.

aviatrix/resource_aviatrix_splunk_logging.go
```diff
@@ -0,0 +1,146 @@
...
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "Port number",
+			},
+			"cu_output_cfg": {
```
CyrusJavan: Is this `cu` short for `custom`, if so we should just not abbreviate it. Same with `cfg`, let's use the non-abbreviated words as much as we can

aviatrix/resource_aviatrix_splunk_logging_test.go
```diff
@@ -0,0 +1,95 @@
...
+		},
+	})
+}
+
+func testAccSplunkLoggingBasic() string {
+	return `
+resource "aviatrix_splunk_logging" "test_splunk_logging" {
+	server            = "1.2.3.4"
+    port              = 10
+	excluded_gateways = ["a", "b"]
```
CyrusJavan: formatting

aviatrix/resource_aviatrix_splunk_logging.go
```diff
@@ -0,0 +1,146 @@
...
+		splunkLogging = marshalSplunkLoggingInput(d, false)
+	}
+
+	var excludedGateways []string
+	for _, v := range d.Get("excluded_gateways").(*schema.Set).List() {
+		excludedGateways = append(excludedGateways, v.(string))
+	}
+	if len(excludedGateways) != 0 {
+		splunkLogging.ExcludedGatewaysInput = strings.Join(excludedGateways, ",")
+	}
```
CyrusJavan: Can move to marshal func

aviatrix/resource_aviatrix_splunk_logging.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func resourceAviatrixSplunkLoggingDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*goaviatrix.Client)
+
+	if err := client.DisableSplunkLogging(); err != nil {
+		return fmt.Errorf("could not disable remote syslog: %v", err)
+	}
+
+	d.SetId("")
```
CyrusJavan: Not necessary to set ID to empty string.

website/docs/r/aviatrix_splunk_logging.html.markdown
```diff
@@ -0,0 +1,49 @@
...
+  custom_output_configuration = file("/path/to/configuration.file")
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+### Required
+* `server` (Optional) Server IP. Either `server` and `port` combination or `cu_output_cfg` is required.
```
CyrusJavan: cu_output_cfg -> custom_output_configuration

CyrusJavan (changes requested):

website/docs/r/aviatrix_splunk_logging.html.markdown
```diff
@@ -0,0 +1,49 @@
+---
```
CyrusJavan, line 1: Need to document the Computed `status` attribute

PR #670, opened 2020-11-25
18666: implemented datadog agent

shuoliugit:


aviatrix/resource_aviatrix_datadog_agent.go
```diff
@@ -0,0 +1,114 @@
...
+
+	datadogAgent := marshalDatadogAgentInput(d)
+
+	var excludedGateways []string
+	for _, v := range d.Get("excluded_gateways").(*schema.Set).List() {
+		excludedGateways = append(excludedGateways, v.(string))
+	}
+	if len(excludedGateways) != 0 {
+		datadogAgent.ExcludedGatewaysInput = strings.Join(excludedGateways, ",")
+	}
```
CyrusJavan, line 67: Could be moved into the marshal func

aviatrix/resource_aviatrix_datadog_agent.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func resourceAviatrixDatadogAgentDelete(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*goaviatrix.Client)
+
+	if err := client.DisableDatadogAgent(); err != nil {
+		return fmt.Errorf("could not disable datadog agent: %v", err)
+	}
+
+	d.SetId("")
```
CyrusJavan: I don't think this is necessary. If the Delete function returns with a `nil` error then the resource  is removed from state, no need to set an empty ID

website/docs/r/aviatrix_datadog_agent.html.markdown
```diff
@@ -0,0 +1,41 @@
+---
```
CyrusJavan, line 1: Please document all the attributes of the resource, including `Computed` values. (`status` is missing)

CyrusJavan (changes requested):

PR #681, opened 2020-12-08
18797: implemented connection between azure vng and aviatrix tgw

shuoliugit:


aviatrix/resource_aviatrix_azure_vng_conn.go
```diff
@@ -0,0 +1,110 @@
...
+		log.Printf("[DEBUG] Looks like an import. Import Id is %s", id)
+
+		d.Set("connection_name", id)
+		connectionName = id
+	}
+
+	azureVngConnStatus, err := client.GetAzureVngConnStatus(connectionName)
+	if err != nil {
+		return fmt.Errorf("could not get azure vng conn status: %v", err)
+	}
```
CyrusJavan: From HashiCorp Terraform Provider Tutorial (https://learn.hashicorp.com/tutorials/terraform/provider-setup):

> When you create something in Terraform but delete it manually, Terraform should gracefully handle it. If the API returns an error when the resource doesn't exist, the read function should check to see if the resource is available first. If the resource isn't available, the function should set the ID to an empty string so Terraform "destroys" the resource in state. The following code snippet is an example of how this can be implemented; you do not need to add this to your configuration for this tutorial.
```
if resourceDoesntExist {
  d.SetID("")
  return
}
```

In our provider we use a specific error to signify that there was no actual error, just the resource was not found `goaviatrix.ErrNotFound`. For example in `dataSourceAviatrixFirewallRead`
```
	fw, err := client.GetPolicy(firewall)
	if err == goaviatrix.ErrNotFound {
		d.SetId("")
		return nil
	}
	if err != nil {
		return fmt.Errorf("error fetching firewall policy for gateway %s: %s", firewall.GwName, err)
	}
```


If you haven't already, please take the time to go through the entire Terraform provider tutorial.

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_azure_vng_conn.go
```diff
@@ -0,0 +1,114 @@
...
+	client := meta.(*goaviatrix.Client)
+
+	azureVngConn := marshalAzureVngConnInput(d)
+
+	if err := client.ConnectAzureVng(azureVngConn); err != nil {
+		return fmt.Errorf("could not connect to azure vng: %v", err)
+	}
+
+	d.SetId(azureVngConn.ConnectionName)
+	return resourceAviatrixAzureVngConnRead(d, meta)
```
CyrusJavan: Only need to return Read if >1 API is called

test-infra/cmdExportOutput.sh
```diff
@@ -51,7 +51,13 @@ TRANSIT_GATEWAY_NAME=$(terraform output TRANSIT_GATEWAY_NAME)
 ARM_RESOURCE_GROUP=$(terraform output ARM_RESOURCE_GROUP)
 ARM_HUB_NAME=$(terraform output ARM_HUB_NAME)
 AWS_TGW_NAME=$(terraform output AWS_TGW_NAME)
+<<<<<<< HEAD
 DATADOG_API_KEY=$(terraform output DATADOG_API_KEY)
+=======
+AZURE_VNG_VNET_ID=$(terraform output AZURE_VNG_VNET_ID)
+AZURE_VNG_SUBNET=$(terraform output AZURE_VNG_SUBNET)
+AZURE_VNG=$(terraform output AZURE_VNG)
+>>>>>>> 20d23c27 (18797: implemented connection between azure vng and aviatrix tgw)
```
CyrusJavan: Clean up

CyrusJavan (changes requested):

CyrusJavan (approved):

PR #687, opened 2020-12-09
18245: add guard duty features

CyrusJavan:
- Add resource `aviatrix_aws_guard_duty` that configures a single account/region pair
- Add `aws_guard_duty_scanning_interval` to controller_config resource


shuoliugit (commented):
Does the file aviatrix.erb need to be updated?

CyrusJavan:
> Does the file aviatrix.erb need to be updated?

@shuoliugit Actually I need to confirm. Now that I think about it, after the move the the Terraform Registry I don't think Hashicorp is still reading the .erb file to generate the sidebar

shuoliugit:
> > Does the file aviatrix.erb need to be updated?
> 
> @shuoliugit Actually I need to confirm. Now that I think about it, after the move the the Terraform Registry I don't think Hashicorp is still reading the .erb file to generate the sidebar

ok, if the .erb file is not needed anymore, maybe we can remove it later.

CyrusJavan:
> > > Does the file aviatrix.erb need to be updated?
> > 
> > 
> > @shuoliugit Actually I need to confirm. Now that I think about it, after the move the the Terraform Registry I don't think Hashicorp is still reading the .erb file to generate the sidebar
> 
> ok, if the .erb file is not needed anymore, maybe we can remove it later.

The docs are now totally generated from the static markdown files, this Ruby (.erb) file is no longer used. See https://www.terraform.io/docs/registry/providers/docs.html. I've opened a Mantis to work on this later https://mantis.aviatrix.com/mantisbt/view.php?id=18944

CyrusJavan:
They also released a tool for auto-generating documentation that we should look into. https://github.com/hashicorp/terraform-plugin-docs

shuoliugit (approved):

PR #693, opened 2020-12-15
18663: implemented filebeat forwarder

shuoliugit:


aviatrix/resource_aviatrix_filebeat_forwarder.go
```diff
@@ -0,0 +1,136 @@
...
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Trusted CA file path",
+			},
+			"config_file_path": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Configuration file path",
+			},
```
CyrusJavan: Why file path and not file content?

shuoliugit: trusted_ca_file_path and config_file_path won't be returned by the backend.

CyrusJavan: The backend does not return the file path either but that should not impact our choice of file path vs content

I thought we had discussed previously that using file content is better for customers since it makes working with remote data (files in s3 buckets for example) easier to work with.

shuoliugit: I see. Then the PostFileAPI needs to be modified a bit. 
Do we also need to maintain the content of trusted_ca_file?

CyrusJavan: Not sure what you mean by "maintain the content" There is nothing special you need to do with the content, it will be saved to state after Create

CyrusJavan (approved):
I like that new PostFileAPI for file content!

PR #700, opened 2020-12-17
18605: add ha_lan_interface_cidr to transit_gateway

CyrusJavan:
PR 7️⃣ 0️⃣ 0️⃣ 🎉 

This PR adds a Computed attribute to transit_gateway to get the LAN interface CIDR of the ha transit gateway. This is used in the Azure fqdn gateway use case.

### DIFF
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -325,6 +325,11 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 				Computed:    true,
 				Description: "Transit gateway lan interface cidr.",
 			},
+			"ha_lan_interface_cidr": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Transit gateway lan interface cidr for the HA gateway.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1142,6 +1147,11 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
 		d.Set("ha_cloud_instance_id", haGw.CloudnGatewayInstID)
 		d.Set("ha_gw_name", haGw.GwName)
 		d.Set("ha_private_ip", haGw.PrivateIP)
+		lanCidr, err := client.GetTransitGatewayLanCidr(haGw.GwName)
+		if err != nil && err != goaviatrix.ErrNotFound {
+			log.Printf("[WARN] Error getting lan cidr for HA transit gateway %s due to %s", haGw.GwName, err)
+		}
+		d.Set("ha_lan_interface_cidr", lanCidr)
```
zhiqiangjin, line 1154: What if err == goaviatrix.ErrNotFound? lanCidr == null?

zhiqiangjin (approved):

aviatrix/resource_aviatrix_transit_gateway.go
CyrusJavan: It’s a string, so it would be “”

PR #704, opened 2020-12-18
19020: Controller private oob implementation

zhiqiangjin:
Unique ID is set same as aviatrix_controller_config: controller ip.
Please also review if the name of the attribute and resource are OK.


aviatrix/resource_aviatrix_controller_private_oob.go
```diff
@@ -0,0 +1,103 @@
+package aviatrix
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
```
CyrusJavan: Needs to be `github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema`

goaviatrix/controller_private_oob.go
```diff
@@ -0,0 +1,103 @@
...
+	"strings"
+)
+
+type PrivateOobResp struct {
+	Return  bool   `json:"return"`
+	Results string `json:"results"`
+	Reason  string `json:"reason"`
+}
+
+func (c *Client) EnablePrivateOob() error {
```
CyrusJavan, line 18: Can use PostAPI with a custom CheckFunc instead. For example:
```golang
func (c *Client) UpdateAwsGuardDutyExcludedIPs(account *AwsGuardDutyAccount) error {
	data := map[string]string{
		"action":       "update_aws_guard_duty_excluded_ips",
		"CID":          c.CID,
		"account_name": account.AccountName,
		"region":       account.Region,
		"excluded_ips": strings.Join(account.ExcludedIPs, ","),
	}
	checkFunc := func(action, reason string, ret bool) error {
		if !ret && !strings.HasPrefix(reason, "No change in the exclude ip list for account") {
		}
		return nil
	}
	return c.PostAPI(data["action"], data, checkFunc)
}
```

goaviatrix/controller_private_oob.go
```diff
@@ -0,0 +1,103 @@
...
+	if !data.Return {
+		if strings.Contains(data.Results, "enable already") {
+			return nil
+		}
+		return errors.New("Rest API enable_private_oob Post failed: " + data.Reason)
+	}
+	return nil
+}
+
+func (c *Client) GetPrivateOobState() (bool, error) {
```
CyrusJavan, line 32: Can use GetAPI

goaviatrix/controller_private_oob.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if data.Results == "Enabled" || data.Results == "enabled" {
+		return true, nil
+	} else if data.Results == "Disabled" || data.Results == "disabled" {
+		return false, nil
+	}
+	return false, ErrNotFound
+}
+
+func (c *Client) DisablePrivateOob() error {
```
CyrusJavan, line 64: Can use PostAPI with custom CheckFunc

aviatrix/resource_aviatrix_controller_private_oob_test.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
```
CyrusJavan: Need to use `github.com/hashicorp/terraform-plugin-sdk/v2/`

CyrusJavan (changes requested):

goaviatrix/controller_private_oob.go
zhiqiangjin: Looks like this "checkFunc" can not cover the err of decoding?

goaviatrix/controller_private_oob.go
zhiqiangjin: Looks like this "checkFunc" can not cover the err of decoding?

goaviatrix/controller_private_oob.go
CyrusJavan: The decoding errors are caught by default.

The checkFunc is for once we decoded successfully and we decide on the error based on the the "Return" and "Reason"

CyrusJavan:
Hi I merged #707 which changes us to the new docs format, could you rebase onto latest master and move
```shell
website/docs/r/aviatrix_controller_private_oob.html.markdown
```
to 
```shell
docs/resources/aviatrix_controller_private_oob.md
```

CyrusJavan (approved):
👍

PR #705, opened 2020-12-18
19013: add single_tunnel in transit_gateway_peering

shuoliugit:


website/docs/r/aviatrix_transit_gateway_peering.html.markdown
```diff
@@ -49,6 +49,9 @@ The following arguments are supported:
 * `prepend_as_path1` - (Optional) AS Path Prepend customized by specifying AS PATH for a BGP connection. Applies on transit_gateway_name1. Available in provider version R2.17.2+.
 * `prepend_as_path2` - (Optional) AS Path Prepend customized by specifying AS PATH for a BGP connection. Applies on transit_gateway_name2. Available in provider version R2.17.2+.
 * `enable_peering_over_private_network` - (Optional) Enable peering over private network. ActiveMesh and Insane Mode is required on both transit gateways. Available in provider version R2.17.1+.
+* `enable_single_tunnel_mode` - (Optional) Enable peering with Single Tunnel mode. False by default.
+  
+~> **NOTE:** Basically the peering is **not** in "Single Tunnel" mode only when `enable_peering_over_private_network` is `true` and `enable_single_tunnel_mode` is `false`. When `enable_peering_over_private_network` is `false`, the "Single Tunnel" peering is public.
```
CyrusJavan: Don't need the word 'Basically', lets keep the docs more formal.

Also this note is a bit confusing to me, why do we say that "the peering is **not** in "Single Tunnel" mode only when `enable_peering_over_private_network` is `true` and `enable_single_tunnel_mode` is `false`."? Isn't the tunnel **not** in single tunnel mode whenever `enable_single_tunnel_mode` is set to False?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -89,6 +89,16 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
...
 				Description: "(Optional) Enable peering over private network. Insane mode is required on both transit gateways. Available as of provider version R2.17.1",
 			},
+			"enable_single_tunnel_mode": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				ForceNew: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return !d.Get("enable_peering_over_private_network").(bool)
+				},
```
CyrusJavan, line 99: Why is this DiffSuppressFunc needed?

website/docs/r/aviatrix_transit_gateway_peering.html.markdown
shuoliugit: Even the backend thinks this new "single_tunnel" feature probably will be confusing. For instance, if you peer two normal gateways (not insane mode), they are in fact connected by a "single tunnel". 

This new feature is for those who just want to peer two insane gateways over private network by a single tunnel. Maybe using the release note here is better.

"Private Transit Gateway Peering with Single-Tunnel Mode expands the existing Insane Mode Transit Gateway Peering Over Private Network to apply it to single IPSec tunnel. One use case is for low speed encryption between cloud networks."

aviatrix/resource_aviatrix_transit_gateway_peering.go
shuoliugit: If it's not peering over private network, there's only one tunnel anyway. So enabling the single_tunnel or not doesn't matter.

website/docs/r/aviatrix_transit_gateway_peering.html.markdown
CyrusJavan: Got it, in this case, we should block the use of this attribute if `enable_peering_over_private_network` is not true. An error message like "`enable_single_tunnel_mode` is only valid when `enable_peering_over_private_network` is set to true" should help the users understand what is going on.

We should then have the docs say that this attribute is only valid with  `enable_peering_over_private_network`.

Also please add "Available as of provider version R2.18+."

aviatrix/resource_aviatrix_transit_gateway_peering.go
CyrusJavan: 👍 Makes sense so if `enable_peering_over_private_network` is false then we don't care about the value of `enable_single_tunnel_mode`. I like this way of doing it, seems very clean!

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -124,6 +134,12 @@ func resourceAviatrixTransitGatewayPeeringCreate(d *schema.ResourceData, meta in
 		PrivateIPPeering:               d.Get("enable_peering_over_private_network").(bool),
 	}
 
+	if transitGatewayPeering.PrivateIPPeering {
+		transitGatewayPeering.SingleTunnel = d.Get("enable_single_tunnel_mode").(bool)
+	} else {
+		return fmt.Errorf("enable_single_tunnel_mode is only valid when enable_peering_over_private_network is set to true")
+	}
+
```
CyrusJavan: Whenever `enable_peering_over_private_network` is false we return an error?

goaviatrix/transit_gateway_peering.go
```diff
@@ -132,6 +138,10 @@ func (c *Client) GetTransitGatewayPeeringDetails(transitGatewayPeering *TransitG
 	transitGatewayPeering.PrependAsPath1 = data.Results.Site1.ConnBGPPrependAsPath
 	transitGatewayPeering.PrependAsPath2 = data.Results.Site2.ConnBGPPrependAsPath
 
+	if len(data.Results.Tunnels[0].LicenseId) == 1 {
```
CyrusJavan, line 141: Does `data.Results.Tunnels` always have at least 1 element in the slice? This would panic if the slice was empty

shuoliugit: There's at least one license assigned if two gateways are connected.

aviatrix/resource_aviatrix_transit_gateway_peering.go
shuoliugit: When `enable_peering_over_private_network` is false, because of `DiffSuppressFunc`, `enable_single_tunnel_mode` will be ignored. Do we still need to return an error?

CyrusJavan: Have you tested creating a peering with enable_peering_over_private_network set to False? The code looks like whenever enable_peering_over_private_network is false you return an error

CyrusJavan: Also this is in the Create function so the diffsuppressfunc does not mean anything

shuoliugit: The logic here is wrong for sure, this will be fixed. 
When I try terraform plan, if the private_network is false, it seems the single_tunnel attribute is ignored directly. So the diffsuppressfunc works in Create?



shuoliu@Shuos-MBP matis_0012272 % terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.


------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

  # aviatrix_transit_gateway_peering.foo will be created
  + resource "aviatrix_transit_gateway_peering" "foo" {
      + enable_peering_over_private_network = false
      + id                                  = (known after apply)
      + transit_gateway_name1               = "test1"
      + transit_gateway_name2               = "test2"
    }

Plan: 1 to add, 0 to change, 0 to destroy.

------------------------------------------------------------------------

CyrusJavan (approved):
👍

PR #711, opened 2021-01-05
18786: VPC with GWLB

CyrusJavan:


aviatrix/resource_aviatrix_vpc.go
```diff
@@ -407,6 +436,41 @@ func resourceAviatrixVpcRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	d.SetId(vC.Name)
+
+	firenetDetail, err := client.GetFireNet(&goaviatrix.FireNet{VpcID: vC.VpcID})
+	if err != nil {
+		d.Set("enable_native_gwlb", false)
```
shuoliugit, line 442: If GetFireNet fails, how do we know enable_native_gwlb is false? Should an error be returned here?

CyrusJavan: Good point. The intended logic is that if GetFireNet failed then it means the vpc is not a FireNet vpc, thus the VPC cannot have enable_native_gwlb. However, what I should really do is check for `ErrNotFound` and return the error if it is anything else.

CyrusJavan: Pushed the change 👍

shuoliugit (approved):

PR #713, opened 2021-01-06
19171: Monitor Gateway Subnets for Transit and Spoke gateways

CyrusJavan:
Modified `EnableMonitorGatewaySubnets` and `DisableMonitorGatewaySubnets` to be more generic for use in Transit and Spoke gateways
Add `enable_monitor_gateway_subnets` and `monitor_exclude_list` to transit_gateway and spoke_gateway

### DIFF
--- a/aviatrix/resource_aviatrix_spoke_gateway.go
+++ b/aviatrix/resource_aviatrix_spoke_gateway.go
@@ -211,6 +211,21 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 				Sensitive:   true,
 				Description: "Customer managed key ID.",
 			},
+			"enable_monitor_gateway_subnets": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+				Description: "Enable [monitor gateway subnets](https://docs.aviatrix.com/HowTos/gateway.html#monitor-gateway-subnet). " +
+					"Only valid for cloud_type = 1 (AWS) or 256 (AWSGOV). Valid values: true, false. Default value: false.",
+			},
+			"monitor_exclude_list": {
+				Type:     schema.TypeSet,
+				Optional: true,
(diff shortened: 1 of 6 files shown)


aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -370,6 +385,18 @@ func resourceAviatrixSpokeGatewayCreate(d *schema.ResourceData, meta interface{}
...
+	var excludedInstances []string
+	for _, v := range d.Get("monitor_exclude_list").(*schema.Set).List() {
+		excludedInstances = append(excludedInstances, v.(string))
+	}
+	if enableMonitorSubnets && gateway.CloudType != goaviatrix.AWS && gateway.CloudType != goaviatrix.AWSGOV {
+		return fmt.Errorf("'enable_monitor_gateway_subnets' is only valid for cloud_type = 1 (AWS) or 256 (AWSGOV)")
+	}
+	if !enableMonitorSubnets && len(excludedInstances) != 0 {
+		return fmt.Errorf("'monitor_exclude_list' must be empty if 'enable_monitor_gateway_subnets' is false")
+	}
```
shuoliugit, line 398: Will the list be used if 'enable_monitor_gateway_subnets' is false? If not, maybe the error message is too strict? I mean, the list won't be effective, but will being empty or not affect anything? How about just letting the user know the list won't be used?

CyrusJavan: The issue is that the next plan would give a diff because when `enable_monitor_gateway_subnets = false` then `monitor_exclude_list` will always read as empty. So, I think "failing fast" by giving the error message here is better than waiting for the user to notice they input an invalid config the next time the do a plan.

In general, my opinion is that we should let the user know they have an invalid config as soon as possible.

shuoliugit (approved):

PR #715, opened 2021-01-07
19019: Implementation of aviatrix_proxy_config resource

zhiqiangjin:


goaviatrix/client_proxy_config.go
```diff
@@ -0,0 +1,107 @@
...
+func (c *Client) GetClientProxyConfig() (*ClientProxyConfig, error) {
+	Url, err := url.Parse(c.baseURL)
+	if err != nil {
+		return nil, errors.New(("url Parsing failed for show_proxy_config") + err.Error())
+	}
+	showProxyConfig := url.Values{}
+	showProxyConfig.Add("CID", c.CID)
+	showProxyConfig.Add("action", "show_proxy_config")
+	Url.RawQuery = showProxyConfig.Encode()
+	resp, err := c.Get(Url.String(), nil)
```
CyrusJavan: Should use GetAPI as per the style guide

goaviatrix/client_proxy_config.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"net/url"
+	"strings"
+)
+
+type ClientProxyConfig struct {
+	HttpProxy          string `form:"http_proxy,omitempty" json:"http_proxy,omitempty"`
```
CyrusJavan: These form struct tags are unused and not necessary

goaviatrix/client_proxy_config.go
```diff
@@ -0,0 +1,107 @@
...
+	HttpsProxy         string `form:"https_proxy,omitempty" json:"https_proxy,omitempty"`
+	ProxyCaCertificate string `form:"server_ca_cert,omitempty" json:"server_ca_cert,omitempty"`
```
CyrusJavan: The JSON omitempty tag is also unnecessary since this struct is only used for decoding

goaviatrix/client_proxy_config.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type ClientProxyConfigResp struct {
+	Return  bool              `json:"return"`
```
CyrusJavan: These JSON tags are unnecessary

aviatrix/resource_aviatrix_client_proxy_config.go
```diff
@@ -0,0 +1,95 @@
...
+				ForceNew:    true,
+				Description: "http proxy URL.",
+			},
+			"https_proxy": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "https proxy URL.",
+			},
+			"proxy_ca_certificate": {
```
CyrusJavan: This should be file content instead of the file path

CyrusJavan (changes requested):

docs/resources/aviatrix_client_proxy_config.md
```diff
@@ -0,0 +1,38 @@
...
+The **aviatrix_client_proxy_config** resource allows management of an Aviatrix Controller's client proxy configurations.
+
+## Example Usage
+
+```hcl
+# Create an Aviatrix Controller Client Proxy Config
+resource "aviatrix_client_proxy_config" "test_proxy_config" {
+  http_proxy           = "172.31.52.145:3127"
+  https_proxy          = "172.31.52.145:3129"
+  proxy_ca_certificate = "/Documents/myCA.pem"
```
CyrusJavan: We want to use file content not the path. So this should be 
```
proxy_ca_certificate = file("/Documents/myCA.pem")
```

docs/resources/aviatrix_client_proxy_config.md
```diff
@@ -0,0 +1,38 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `http_proxy` - (Required) Http proxy URL.
+* `https_proxy` - (Required) Https proxy URL.
+* `proxy_ca_certificate` - (Optional) Server CA Certificate file.
```
CyrusJavan: We should add a line here 
```
Use the `file` function to read from a file.
```

goaviatrix/client_proxy_config.go
```diff
@@ -0,0 +1,92 @@
...
+			"http_proxy":  clientProxyConfig.HttpProxy,
+			"https_proxy": clientProxyConfig.HttpsProxy,
+		}
+
+		files := []File{
+			{
+				Path:      clientProxyConfig.ProxyCaCertificate,
+				ParamName: "server_ca_cert",
+			},
+		}
```
CyrusJavan: This will not work with file content, this is for reading a file path.
To use file content you would use the UseFileContent option. See Shuo's filebeat_forwarder resource as an example.
```

	var files []File

	if r.TrustedCAFile != "" {
		ca := File{
			ParamName:      "trusted_ca",
			UseFileContent: true,
			FileName:       "ca.pem", // fake name for ca
			FileContent:    r.TrustedCAFile,
		}
		files = append(files, ca)
	}
```

CyrusJavan (changes requested):

arvindsree:
This also needs to be implemented into the controller init module. This is usually set before the controller is initialized. Mostly the environments would not have internet access if the proxy is not set. So the controller would not be able to complete the initialization step without configuring proxy first

goaviatrix/client_proxy_config.go
```diff
@@ -0,0 +1,76 @@
...
+	} else {
+		data := map[string]interface{}{
+			"CID":         c.CID,
+			"action":      action,
+			"http_proxy":  clientProxyConfig.HttpProxy,
+			"https_proxy": clientProxyConfig.HttpsProxy,
+		}
+		return c.PostAPI(action, data, BasicCheck)
+	}
+}
```
CyrusJavan: We can clean this up a bit by only defining the data map once and just adding the file to the []File inline
```
	action := "apply_proxy_config"
	params := map[string]string{
		"CID":         c.CID,
		"action":      action,
		"http_proxy":  clientProxyConfig.HttpProxy,
		"https_proxy": clientProxyConfig.HttpsProxy,
	}
	if clientProxyConfig.ProxyCaCertificate != "" {
		files := []File{
			{
				ParamName:      "server_ca_cert",
				UseFileContent: true,
				FileName:       "ca.pem", // fake name for ca
				FileContent:    clientProxyConfig.ProxyCaCertificate,
			},
		}
		return c.PostFileAPI(params, files, BasicCheck)
	} else {
		return c.PostAPI(action, params, BasicCheck)
	}
```

CyrusJavan (approved):

PR #723, opened 2021-01-12
fix19245: add data source support

shuoliugit:


aviatrix/data_source_aviatrix_vpc.go
```diff
@@ -197,9 +202,32 @@ func dataSourceAviatrixVpcRead(d *schema.ResourceData, meta interface{}) error {
...
+			var subscriptionId string
+			if acc != nil {
+				subscriptionId = acc.ArmSubscriptionId
+			}
+
+			resourceGroup := strings.Split(vC.VpcID, ":")[1]
+			azureVnetResourceId := "/subscriptions/" + subscriptionId + "/resourceGroups/" + resourceGroup + "/providers/Microsoft.Network/virtualNetworks/" + vC.Name
+			d.Set("resource_group", resourceGroup)
+			d.Set("azure_vnet_resource_id", azureVnetResourceId)
+		}
```
CyrusJavan: This block does not need to be in the `else`. Less indentation the better.

CyrusJavan (changes requested):

aviatrix/data_source_aviatrix_vpc.go
shuoliugit: also refactored this part in resource_aviatrix_vpc

CyrusJavan (approved):

PR #730, opened 2021-01-13
18907: fix race condition when deploying spoke gateway

shuoliugit:


CyrusJavan (approved):
Wow I didn't even know we did retry logic like this in the provider, cool!

zhiqiangjin (approved):

PR #736, opened 2021-01-20
19549: OOB tgw (and peer hagw) and sgw (and peer hagw)

shuoliugit:


aviatrix/data_source_aviatrix_spoke_gateway.go
```diff
@@ -180,6 +180,22 @@ func dataSourceAviatrixSpokeGateway() *schema.Resource {
...
+			"oob_management_subnet": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "OOB management subnet.",
+			},
+			"oob_availability_zone": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "OOB subnet availability zone.",
+			},
```
CyrusJavan: Data source attributes should be `Computed`

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -92,9 +92,14 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
...
-				Optional:    true,
-				Computed:    true,
+				Type:     schema.TypeString,
+				Optional: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					if old != "" {
+						return new != old
+					}
+					return false
+				},
```
CyrusJavan: Could you explain the reasoning for this? We will never show a diff after the eip is set?

Also nitpick: could be one line `return old != "" && new != old`

aviatrix/data_source_aviatrix_transit_gateway.go
```diff
@@ -224,6 +224,22 @@ func dataSourceAviatrixTransitGateway() *schema.Resource {
...
+			"oob_management_subnet": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "OOB management subnet.",
+			},
+			"oob_availability_zone": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "OOB subnet availability zone.",
+			},
```
CyrusJavan: Data source attributes should be `Computed`

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_spoke_gateway.go
shuoliugit: For this feature, we don't allocate new eip for the tgw, so we have to set eip. But I found that no matter what you set, the backend will assign a different one (not a new eip), which will trigger a diff.

CyrusJavan: Have you asked Steve about that behavior? I think we can find a better solution than just never reading a diff for this attribute

shuoliugit: Now I stop reading any diff for `allocate_new_eip` if private oob is enabled.

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -672,7 +724,7 @@ func resourceAviatrixSpokeGatewayRead(d *schema.ResourceData, meta interface{})
 			d.Set("vpc_id", strings.Split(gw.VpcID, "~~")[0]) //AWS vpc_id returns as <vpc_id>~~<other vpc info> in rest api
 			d.Set("vpc_reg", gw.VpcRegion)                    //AWS vpc_reg returns as vpc_region in rest api
 
-			if gw.AllocateNewEipRead {
+			if gw.AllocateNewEipRead && !gw.EnablePrivateOob {
```
CyrusJavan, line 727: If `enable_private_oob == true` then `allocate_new_eip` will always be read as false. However,  `allocate_new_eip` is default `true`. Thus if someone does not explicitly set `allocate_new_eip` to false then they would have a diff after apply. To me this implies that setting `enable_private_oob == true && allocate_new_eip == true` is an invalid config and we should throw an error if that occurs.

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_spoke_gateway.go
shuoliugit: Now I use the following `DiffSuppressFunc` for `allocate_new_eip`. The idea is to completely ignore `allocate_new_eip` when oob is enabled.

```
DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
	return d.Get("enable_private_oob").(bool)
},
```

CyrusJavan: Ohhhh I see now, got it.

CyrusJavan (approved):

PR #740, opened 2021-01-22
19738: handle ErrNotFound properly in data sources

CyrusJavan:


aviatrix/data_source_aviatrix_account.go
```diff
@@ -58,6 +58,10 @@ func dataSourceAviatrixAccountRead(d *schema.ResourceData, meta interface{}) err
 
 	acc, err := client.GetAccount(account)
 	if err != nil {
+		if err == goaviatrix.ErrNotFound {
+			d.SetId("")
+			return nil
+		}
 		return fmt.Errorf("aviatrix Account: %s", err)
```
shuoliugit, line 65: If the resource is not ready, although the data source won't return an error, will the data source still be empty before the user runs plan or apply again?

CyrusJavan: In the state file, yes the data source would be empty. However, by default, Terraform will refresh the state prior to any plan/apply step. So, it should be transparent to the user that the data source was ever empty. (Which is better than the current behavior where any refresh will just fail because the data source returns an error)

zhiqiangjin (approved):

PR #744, opened 2021-01-26
Update aviatrix_firewall_instance.md

fkhademi:
Added version info for a few Azure specific attributes.  These were added in 2.17.1 provider AFAIK and the docs should highlight that


CyrusJavan (approved):
Thanks!

zhiqiangjin (approved):

PR #745, opened 2021-01-27
19804: Make firenet_gw_name Optional

CyrusJavan:
Within firewall_instance and firewall_instance_association change
firenet_gw_name from Required to Optional.

Also since GWLB FireNet firewall_instance requires a 'zone' parameter
we expand the use of the 'zone' parameter to allow use with AWS
GWLB. Previously, 'zone' was only for Azure.

Test:
Create firewall_instance with GWLB enabled VPC and no firenet_gw_name
PASS

Create firewall_instance without GWLB and a firenet_gw_name
PASS

Create firewall_instance_association with GWLB enabled VPC and no firenet_gw_name
PASS

Create firewall_instance_association without GWLB and a firenet_gw_name
PASS

### DIFF
--- a/aviatrix/resource_aviatrix_firewall_instance.go
+++ b/aviatrix/resource_aviatrix_firewall_instance.go
@@ -27,7 +27,7 @@ func resourceAviatrixFirewallInstance() *schema.Resource {
 			},
 			"firenet_gw_name": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				ForceNew:    true,
 				Description: "Name of the primary FireNet gateway.",
 			},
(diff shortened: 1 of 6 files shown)


aviatrix/resource_aviatrix_firewall_instance.go
```diff
@@ -249,22 +249,21 @@ func resourceAviatrixFirewallInstanceCreate(d *schema.ResourceData, meta interfa
 	// gateway name. If there is an issue, we will just continue on without the additional
 	// validation.
 	var cloudType int
-	gw, err := client.GetGateway(&goaviatrix.Gateway{GwName: firewallInstance.GwName})
-	if err != nil {
-		log.Printf("[WARN] Could not get cloud_type from firenet_gw_name: %v", err)
+	if firewallInstance.GwName == "" {
```
zhiqiangjin, line 276: If firewallInstance.GwName == "", it could be user forgetting to set the value. So, if it is empty, need to confirm it is AWS GWLB Firenet; on the other hand, if it is not empty, need to confirm it is not AWS GWLB.

zhiqiangjin (changes requested):

aviatrix/resource_aviatrix_firewall_instance.go
CyrusJavan: I can try to get cloud_type from the vpc id

> on the other hand, if it is not empty, need to confirm it is not AWS GWLB.

~~I don't think this is necessary. Backend will return an error if the VPC is not compatible~~
You are right, it is better user experience to catch the invalid config early.

zhiqiangjin (approved):

PR #757, opened 2021-02-03
18470: Gateway Certificate Import

CyrusJavan:
Add new resource `aviatrix_gateway_certificate_config`.

This resources has two required, force new attributes:
- ca_certificate
- ca_private_key

This resource also takes advantage of the Context aware
version of the CRUD functions that were made available
in v2 of the terraform provider SDK.


docs/resources/aviatrix_gateway_certificate_config.md
```diff
@@ -0,0 +1,40 @@
+---
+subcategory: "Settings"
+layout: "aviatrix"
+page_title: "Aviatrix: aviatrix_gateway_certificate_config"
+description: |-
+    Manages Aviatrix gateway certificate configuration
```
zhiqiangjin: Should be two indent, not four.

docs/resources/aviatrix_gateway_certificate_config.md
```diff
@@ -0,0 +1,40 @@
...
+The following arguments are supported:
+
+### Required
+* `ca_certificate` - (Required) CA Certificate in PEM format. To read certificate from a file please use the built-in `file` function.
+* `ca_private_key` - (Required/Sensitive) CA Private Key. To read the private key from a file please use the built-in `file` function.
+
+
+## Import
+
+!> **WARNING:** Due to security reasons the Aviatrix controller does not ever output your CA private key or CA certificate, thus when importing the provider cannot read your private key or certificate into the state file. After importing if you wish to not change the values of the CA private key or certificate you must set the attributes `ca_certificate` and `ca_private_key` to the empty string (""). If you do provide non-empty values for `ca_certificate` and `ca_private_key` after importing, then Terraform will see a diff between the given values and the values in the state file and force replacement of the resource.
```
zhiqiangjin: Make this paragraph shorter.

goaviatrix/client.go
```diff
@@ -254,28 +342,28 @@ func (c *Client) PostFile(path string, params map[string]string, files []File) (
 
 			file, err := os.Open(f.Path)
 			if err != nil {
-				return nil, err
+				return nil, "", err
```
zhiqiangjin, line 345: Is "PostFile" used anywhere else? Output format changing causes issue wherever else it is used, right?

zhiqiangjin (changes requested):

shuoliugit:
Are we going to use the Context aware version of the CRUD functions from now on?

goaviatrix/client.go
CyrusJavan: I did not change the output format of PostFile. This is actually a new function `func encodeMultipartFormData(params map[string]string, files []File) (*bytes.Buffer, string, error) {` that encapsulates some behavior from PostFile so that I can reuse it in PostFile and PostFileContext

CyrusJavan:
> Are we going to use the Context aware version of the CRUD functions from now on?

@shuoliugit In my opinion, yes we should.

1) The non-context aware CRUD functions are deprecated.
2) The context aware CRUD functions return a `diag.Diagnostics` instead of `error` which allows us to return Warnings and more detailed error messages to users.
3) By passing the context through the API calls it makes our provider more responsive to when users hit Ctrl-C during an apply since Terraform will cancel the context.

shuoliugit:
> > Are we going to use the Context aware version of the CRUD functions from now on?
> 
> @shuoliugit In my opinion, yes we should.
> 
> 1. The non-context aware CRUD functions are deprecated.
> 2. The context aware CRUD functions return a `diag.Diagnostics` instead of `error` which allows us to return Warnings and more detailed error messages to users.
> 3. By passing the context through the API calls it makes our provider more responsive to when users hit Ctrl-C during an apply since Terraform will cancel the context.

Good to know, thanks!

CyrusJavan:
@shuoliugit @zhiqiangjin I just noticed a bug actually when a user would hit Ctrl-C during an apply with this resource it would crash, this issue was fixed in a newer version of the SDK so I upgraded by doing:

```
go get github.com/hashicorp/terraform-plugin-sdk/v2
go mod tidy
go mod vendor
```

That is why there are all those deleted files now.

zhiqiangjin (approved):

PR #760, opened 2021-02-05
19914: Fix firewall_instance.key_name import

CyrusJavan:
Due to a bug in export we would not export
the key_name field. Then if a user imported
the resource while the key_name field was
blank then the key_name was not read into state.

From then on the key_name would never be read into
state, even if user added the key_name to the config
because of the blank value in the state file.

Also update the documentation as it is not clear that
the key_name is actually an AWS KeyPair name.


zhiqiangjin (approved):

CyrusJavan:
Hi @zhiqiangjin

I noticed that actually I should probably check the cloud type of the firewall before writing to state.

Also since I needed to add cloud types I fixed another piece of code that was not checking all the correct vendor names.

shuoliugit (approved):

PR #762, opened 2021-02-05
20043: remove egress+connected_transit check

CyrusJavan:
Prior to 6.3 having egress+connected_transit was allowed so there exists customer with egress+connected_transit already deployed. When those customers try to do any update to their transit_gateway they will hit this block, however, if they try to set their connected_transit=false and apply, they get the error `[AVXERR-TRANSIT-0018] Transit edit is not supported on Egress Transit gateway` so that leaves them with an unusable Terraform.

Instead we should not have this provider-side check at all in the Update function, if there is any issue with having egress+connected_transit then backend is responsible for returning an error.


inc1t3Ful (approved):
Approved 👍 

As discussed internally and with Felipe; should be a backend ticket as well, so future customers moving to 6.3 should not be in this strange limbo state

zhiqiangjin (approved):

PR #763, opened 2021-02-06
19867: Fix issue where terrafrom refresh/destroy does not work if vgw conn has been removed from UI

zhiqiangjin:
If connection does not exist, do not return error. To be consistent with aviatrix_site2cloud, aviatrix_transit_external_device_conn, and aviatrix_device_transit_gateway_attachment, and aviatrix_device_virtual_wan_attachment.

### DIFF
--- a/goaviatrix/vgw_connection.go
+++ b/goaviatrix/vgw_connection.go
@@ -182,6 +182,9 @@ func (c *Client) GetVGWConnDetail(vgwConn *VGWConn) (*VGWConn, error) {
 	var data VGWConnDetailResp
 	err := c.GetAPI(&data, params["action"], params, BasicCheck)
 	if err != nil {
+		if strings.Contains(data.Reason, "does not exist") {
+			return nil, ErrNotFound
+		}
 		return nil, err
 	}
 	if data.Results.Connections.ConnName[0] != "" {


CyrusJavan (approved):
You could also define a custom checkFunc, for example:
```
	checkFunc := func(action, reason string, ret bool) error {
		if !ret {
			if strings.Contains(reason, "does not exist") {
				return ErrNotFound
			}
			return errors.New("Rest API 'get_site2cloud_conn_detail' Get failed: " + reason)
		}
		return nil
	}
	var data ExternalDeviceConnDetailResp
	err := c.GetAPI(&data, params["action"], params, checkFunc)
```

arvindsree:
> You could also define a custom checkFunc, for example:
> 
> ```
> 	checkFunc := func(action, reason string, ret bool) error {
> 		if !ret {
> 			if strings.Contains(reason, "does not exist") {
> 				return ErrNotFound
> 			}
> 			return errors.New("Rest API 'get_site2cloud_conn_detail' Get failed: " + reason)
> 		}
> 		return nil
> 	}
> 	var data ExternalDeviceConnDetailResp
> 	err := c.GetAPI(&data, params["action"], params, checkFunc)
> ```

Maybe we could make this a common function where we pass "does not exist" or a list of strings that can be ignored in the reason if it fails as an argument

zhiqiangjin:
Sometimes it is "does not exist", sometimes it is something else.

CyrusJavan:
Next time we need to do something like this we could add a convenience check function like Arvind described
```golang
func ErrNotFoundCheck(notFoundStrings []string) func(string, string, bool) error {
    return func(action, reason string, ret bool) error {
        if !ret {
            if containsAnyString(reason, notFoundStrings) {
                return ErrNotFound
            }
        }
        return nil
    }
}
``` 

Then in this case you would have done 
```
err := c.GetAPI(&data, params["action"], params, ErrNotFoundCheck([]string{"does not exist"}))
```

PR #765, opened 2021-02-09
19591: add route_tables in aviatrix_vpc

shuoliugit:


CyrusJavan (commented):
Should we add `route_tables_filter` as well to resource vpc? Would be useful to the user.

shuoliugit:
> Should we add `route_tables_filter` as well to resource vpc? Would be useful to the user.

It's useful, but I'm not sure if there's a way to set it properly in Read(). And since `route_tables_filter` is already in the vpc data_source, user can get the information in the data_source.

CyrusJavan:
> It's useful, but I'm not sure if there's a way to set it properly in Read(). And since `route_tables_filter` is already in the vpc data_source, user can get the information in the data_source.

How do we set `route_tables_filter` in the data source?

Also if this is all already covered in the data_source, why add it to the resource at all?

shuoliugit:
> > It's useful, but I'm not sure if there's a way to set it properly in Read(). And since `route_tables_filter` is already in the vpc data_source, user can get the information in the data_source.
> 
> How do we set `route_tables_filter` in the data source?
> 
> Also if this is all already covered in the data_source, why add it to the resource at all?

In data source, it's an argument.
I think maybe the reporter doesn't know about the data source. Let me check with the reporter, maybe we can close this.

zhiqiangjin (approved):

shuoliugit:
CK is happy with data_source_aviatrix_vpc, but several improvements are still made here:
1. add support for Azure route tables in `data_source_aviatrix_vpc`
2. add `route_tables` in `resource_aviatrix_vpc`

PR #767, opened 2021-02-10
19543: aws_tgw cidrs

CyrusJavan:
image

TGW -> List -> Edit -> Edit TGW CIDR

This feature is a requirement for the upcoming TGW Connect resource.

### Unit Test:
```
Create aws_tgw with cidrs
PASS

Create aws_tgw without cidrs
PASS

Update aws_tgw cidrs
PASS

Delete aws_tgw with cidrs
PASS

Import aws_tgw with cidrs
PASS

Import aws_tgw without cidrs
PASS
```

### DIFF
--- a/aviatrix/resource_aviatrix_aws_tgw.go
+++ b/aviatrix/resource_aviatrix_aws_tgw.go
@@ -181,9 +181,19 @@ func resourceAviatrixAWSTgw() *schema.Resource {
 					"Valid values: true, false. Default value: true.",
 			},
 			"enable_multicast": {
-				Type:     schema.TypeBool,
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Enable Multicast.",
+			},
+			"cidrs": {
+				Type:     schema.TypeSet,
 				Optional: true,
-				Default:  false,
+				Elem: &schema.Schema{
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_aws_tgw.go
```diff
@@ -181,9 +181,19 @@ func resourceAviatrixAWSTgw() *schema.Resource {
 					"Valid values: true, false. Default value: true.",
 			},
 			"enable_multicast": {
-				Type:     schema.TypeBool,
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Enable Multicast.",
+			},
+			"cidrs": {
```
zhiqiangjin, line 189: 1. May need check if backend preserve the order of the input for "cidrs";
2. May also check if "cidrs" will be referenced somewhere. If yes, the order of the cidrs may matter.

CyrusJavan: 1. Order that is input into the UI is not preserved.
2. In the TGW Connect workflow, the CIDRs are not referenced.

shuoliugit (approved):

zhiqiangjin (approved):

PR #769, opened 2021-02-17
20100: Fix invalid Session Errors while using aviatrix_controller_config resource

zhiqiangjin:
Add client.Login() after every successful upgrade.


arvindsree:
Better to move the login() call inside the Upgrade() function

arvindsree (approved):

PR #774, opened 2021-02-19
20199: Added enable_jumbo_frame to gateway, transit gateway and spoke gateway

mlin-aviatrix:


aviatrix/resource_aviatrix_gateway.go
```diff
@@ -450,6 +450,12 @@ func resourceAviatrixGateway() *schema.Resource {
 				Optional:    true,
 				Description: "Whether to enforce Guard Duty IP blocking. Required when `enable_public_subnet_filtering` attribute is true. Valid values: true or false. Default value: true.",
 			},
+			"enable_jumbo_frame": {
+				Type:        schema.TypeBool,
+				Default:     true,
+				Optional:    true,
+				Description: "Enable jumbo frame support for Gateway. Valid values: true of false. Default value: true.",
```
CyrusJavan: Typo: true of false -> true or false

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -968,6 +974,18 @@ func resourceAviatrixGatewayCreate(d *schema.ResourceData, meta interface{}) err
 		}
 	}
 
+	if d.Get("enable_jumbo_frame").(bool) {
+		err := client.EnableJumboFrame(gateway)
+		if err != nil {
+			return fmt.Errorf("couldn't enable jumbo frames for Gateway: %s", err)
+		}
```
CyrusJavan: Why call EnableJumboFrame if it is already enabled by default?

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -278,6 +278,12 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 				Optional:    true,
 				Description: "OOB subnet availability zone.",
 			},
+			"enable_jumbo_frame": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Enable jumbo frame support for transit gateway. Valid values: true of false. Default value: true.",
```
CyrusJavan: Typo: true of false -> true or false

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -399,6 +399,12 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 				Optional:    true,
 				Description: "OOB subnet availability zone.",
 			},
+			"enable_jumbo_frame": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Enable jumbo frame support for spoke gateway. Valid values: true of false. Default value: true.",
```
CyrusJavan: Typo: true of false -> true or false

docs/resources/aviatrix_gateway.md
```diff
@@ -263,6 +263,7 @@ The following arguments are supported:
 * `tag_list` - (Optional) Tag list of the gateway instance. Only available for AWS, AWSGov and Azure gateways. Example: ["key1:value1", "key2:value2"].
 * `enable_vpc_dns_server` - (Optional) Enable VPC DNS Server for gateway. Currently only supported for AWS and AWSGov gateways. Valid values: true, false. Default value: false.
 * `zone` - (Optional) Availability Zone. Only available for Azure and Public Subnet Filtering gateway. Available for Azure as of provider version R2.17+.
+* `enable_jumbo_frame` - (Optional) Enable jumbo frames for the gateway. Default value is true.
```
CyrusJavan: Add documentation for spoke and transit as well.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -2291,6 +2314,20 @@ func resourceAviatrixTransitGatewayUpdate(d *schema.ResourceData, meta interface
 		}
 	}
 
+	if d.HasChange("enable_jumbo_frame") {
+		if d.Get("enable_jumbo_frame").(bool) {
+			err := client.EnableJumboFrame(gateway)
+			if err != nil {
+				return fmt.Errorf("could not enable jumbo frame for gateway: %v", err)
```
CyrusJavan: Error messages need to be unique strings for debugging/diagnosing issues. You have used the same string `could not enable jumbo frame for gateway` across the 3 gateway resources, if a customer reported a problem and all we had was this error string, we would have no clue which gateway resource had the problem. Also you reused the same string for Create and Update, which is also ambiguous.

Add more context to make the strings unique, for example, this is Transit Gateway update so the error should be
```
could not enable jumbo frame for transit gateway when updating
```

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_transit_gateway.go
mlin-aviatrix: I added unique messages for the error messages so you can tell if it's a gateway, spoke gateway or transit gateway. Also differentiated create and update error messages.

CyrusJavan (approved):

zhiqiangjin (approved):

PR #778, opened 2021-02-24
20092: Added tags to firewall_instance

mlin-aviatrix:


goaviatrix/firewall_instance.go
```diff
@@ -119,6 +121,16 @@ func (c *Client) CreateFirewallInstance(firewallInstance *FirewallInstance) (str
 	if firewallInstance.UserData != "" {
 		addFirewallInstance.Add("user_data", firewallInstance.UserData)
 	}
+	if len(firewallInstance.Tags) > 0 {
+		tagList := make([]string, 0, len(firewallInstance.Tags))
+		for key, val := range firewallInstance.Tags {
+			escapedKey := strings.ReplaceAll(key, ":", "\\:")
+			escapedVal := strings.ReplaceAll(fmt.Sprint(val), ":", "\\:")
```
CyrusJavan: Don't need fmt.Sprint

goaviatrix/firewall_instance.go
```diff
@@ -31,20 +32,21 @@ type FirewallInstance struct {
...
+	BootstrapStorageName string                 `form:"bootstrap_storage_name,omitempty" json:"bootstrap_storage_name,omitempty"`
+	StorageAccessKey     string                 `form:"storage_access_key,omitempty" json:"storage_access_key,omitempty"`
+	FileShareFolder      string                 `form:"file_share_folder,omitempty" json:"file_share_folder,omitempty"`
+	ShareDirectory       string                 `form:"share_directory,omitempty" json:"share_directory,omitempty"`
+	SicKey               string                 `form:"sic_key,omitempty" json:"sic_key,omitempty"`
+	ContainerFolder      string                 `form:"container_folder,omitempty" json:"container_folder,omitempty"`
+	SasUrlConfig         string                 `form:"sas_url_config,omitempty" json:"sas_url_config,omitempty"`
+	SasUriLicense        string                 `form:"sas_url_license,omitempty" json:"sas_url_license,omitempty"`
+	UserData             string                 `form:"user_data,omitempty" json:"user_data,omitempty"`
+	Tags                 map[string]interface{} `json:"usr_tags,omitempty"`
```
CyrusJavan: The omitempty here is not necessary. omitempty only applies to the encoding behavior of the field. Since this field is only decoded, it is not needed.

aviatrix/resource_aviatrix_firewall_instance.go
```diff
@@ -199,6 +199,14 @@ func resourceAviatrixFirewallInstance() *schema.Resource {
 				Computed:    true,
 				Description: "Management Public IP.",
 			},
+			"tags": {
+				Type:        schema.TypeMap,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Default:     nil,
+				ForceNew:    true,
```
CyrusJavan, line 206: Is it really ForceNew? Interesting, I would think tags would be mutable.

aviatrix/resource_aviatrix_firewall_instance.go
```diff
@@ -199,6 +199,14 @@ func resourceAviatrixFirewallInstance() *schema.Resource {
 				Computed:    true,
 				Description: "Management Public IP.",
 			},
+			"tags": {
+				Type:        schema.TypeMap,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Default:     nil,
```
CyrusJavan: This is unnecessary, Default is of type interface{} and the default value for interface{} is already nil.

aviatrix/resource_aviatrix_firewall_instance.go
mlin-aviatrix: Firewall instance has no update method. The API does not support updating/editing firewall instances and the UI doesn't support it either.

goaviatrix/firewall_instance.go
mlin-aviatrix: firewallInstance.Tags is map[string]interface{}, so val is of type interface{} and needs to be cast to a string.

aviatrix/resource_aviatrix_firewall_instance.go
```diff
@@ -334,6 +341,12 @@ func resourceAviatrixFirewallInstanceCreate(d *schema.ResourceData, meta interfa
 		}
 	}
 
+	if tags, ok := d.GetOk("tags"); ok && (cloudType == goaviatrix.AWS || cloudType == goaviatrix.AWSGOV || cloudType == goaviatrix.AZURE) {
+		firewallInstance.Tags = tags.(map[string]interface{})
```
CyrusJavan: Could we just use map[string]string instead? I think the map[string]interface{} is a bit confusing.

I think the best path forward is a function like `func getTags(d d *schema.ResourceData) map[string]string` that will pull out the "tags" attribute and convert to map[string]string. You could also put the colon escaping logic in this function so that we can reuse it later when we change all `tag_list` to `tags`

CyrusJavan (changes requested):

goaviatrix/firewall_instance.go
CyrusJavan: Ah I see okay. See my comment below about changing this to map[string]string

goaviatrix/firewall_instance.go
```diff
@@ -119,6 +120,16 @@ func (c *Client) CreateFirewallInstance(firewallInstance *FirewallInstance) (str
 	if firewallInstance.UserData != "" {
 		addFirewallInstance.Add("user_data", firewallInstance.UserData)
 	}
+	if len(firewallInstance.Tags) > 0 {
+		tagList := make([]string, 0, len(firewallInstance.Tags))
+		for key, val := range firewallInstance.Tags {
+			escapedKey := strings.ReplaceAll(key, ":", "\\:")
+			escapedVal := strings.ReplaceAll(val, ":", "\\:")
```
CyrusJavan: Don't need this anymore since we already escape in extractTags

CyrusJavan (changes requested):

aviatrix/utils.go
```diff
@@ -103,3 +104,29 @@ func intInSlice(needle int, haystack []int) bool {
...
+	if tags, ok := d.GetOk("tags"); ok && (cloudType == goaviatrix.AWS || cloudType == goaviatrix.AWSGOV || cloudType == goaviatrix.AZURE) {
+		tagsMap := tags.(map[string]interface{})
+		tagsStrMap := make(map[string]string, len(tagsMap))
+
+		for key, val := range tagsMap {
+			valStr := fmt.Sprint(val)
+			if strings.Contains(key, ",") {
+				return nil, fmt.Errorf("illegal character in tags: \",\"")
+			}
+			matched, err := regexp.MatchString("^[a-zA-Z0-9+\\-=._ :/@ ]*$", key+valStr)
```
CyrusJavan: Should compile the regex outside of the function 
```
var tagMatcher = regexp.MustCompile("^[a-zA-Z0-9+\\-=._ :/@ ]*$")
...

if tagMatcher.Match(key+valStr)

```

aviatrix/utils.go
```diff
@@ -103,3 +104,29 @@ func intInSlice(needle int, haystack []int) bool {
...
+
+func extractTags(d *schema.ResourceData, cloudType int) (map[string]string, error) {
+	if tags, ok := d.GetOk("tags"); ok && (cloudType == goaviatrix.AWS || cloudType == goaviatrix.AWSGOV || cloudType == goaviatrix.AZURE) {
+		tagsMap := tags.(map[string]interface{})
+		tagsStrMap := make(map[string]string, len(tagsMap))
+
+		for key, val := range tagsMap {
+			valStr := fmt.Sprint(val)
+			if strings.Contains(key, ",") {
+				return nil, fmt.Errorf("illegal character in tags: \",\"")
```
CyrusJavan: Wouldn't the regexp below catch this already?

aviatrix/resource_aviatrix_firewall_instance.go
```diff
@@ -334,6 +341,13 @@ func resourceAviatrixFirewallInstanceCreate(d *schema.ResourceData, meta interfa
 		}
 	}
 
+	tags, err := extractTags(d, cloudType)
+	if err != nil {
+		return fmt.Errorf("error creating tags for firewall instance: %v", err)
+	} else if tags != nil {
+		firewallInstance.Tags = tags
+	}
```
CyrusJavan: This else if is unnecessary, could just do: 
```
        tags, err := extractTags(d, cloudType)
	if err != nil {
		return fmt.Errorf("error creating tags for firewall instance: %v", err)
	}
        firewallInstance.Tags = tags
```

aviatrix/utils.go
```diff
@@ -103,3 +104,29 @@ func intInSlice(needle int, haystack []int) bool {
 	}
 	return false
 }
+
+func extractTags(d *schema.ResourceData, cloudType int) (map[string]string, error) {
```
CyrusJavan, line 108: Nitpick: It is more idiomatic Go style to keep the "happy path" (non-error case) aligned on the left, and leave the indented parts to the error handling. For this function it would look like
```golang
func extractTags(d *schema.ResourceData, cloudType int) (map[string]string, error) {
	tags, ok := d.GetOk("tags")
	 if ok && !intInSlice(cloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV,goaviatrix.AZURE} {
		return nil, fmt.Errorf("adding tags is only supported for AWS, AWSGOV and AZURE, cloud_type must be 1, 256 or 8")
	}
	 if !ok {
	 	 return nil, nil
	 }
	tagsMap := tags.(map[string]interface{})
	tagsStrMap := make(map[string]string, len(tagsMap))
	for key, val := range tagsMap {
		valStr := fmt.Sprint(val)
		if strings.Contains(key, ",") {
			return nil, fmt.Errorf("illegal character in tags: \",\"")
		}
		matched, err := regexp.MatchString("^[a-zA-Z0-9+\\-=._ :/@ ]*$", key+valStr)
		if err != nil || !matched {
			return nil, fmt.Errorf("illegal characters in tags")
		}
		escapedKey := strings.ReplaceAll(key, ":", "\\:")
		escapedVal := strings.ReplaceAll(valStr, ":", "\\:")
		tagsStrMap[escapedKey] = escapedVal
	}
	return tagsStrMap, nil
}
```

We end up with a less indented function that is easier to read.

CyrusJavan (changes requested):

CyrusJavan (approved):
Good job on this one!

zhiqiangjin (approved):

PR #782, opened 2021-02-27
20384: Add aviatrix_controller_bgp_max_as_limit_config resource

mlin-aviatrix:
Add new resource to manage controller config for BGP maximum AS limit for transit gateways.


docs/resources/aviatrix_controller_bgp_max_as_limit_config.md
```diff
@@ -0,0 +1,35 @@
...
+subcategory: "Settings"
+layout: "aviatrix"
+page_title: "Aviatrix: aviatrix_controller_bgp_max_as_limit_config"
+description: |-
+  Creates and manages an Aviatrix controller BGP max AS limit for transit gateways
+---
+
+# aviatrix_controller_bgp_max_as_limit_config
+
+The **aviatrix_controller_bgp_max_as_limit_config** resource allows management of an Aviatrix Controller's BGP max AS limit for transit gateways.
```
CyrusJavan: Please append a sentence `This resource is available as of provider version R2.18.1+.`

goaviatrix/controller_bgp_max_as_limit_config.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	var resp BgpMaxAsLimitResponse
+	err := c.GetAPIContext(ctx, &resp, data["action"], data, BasicCheck)
+	if err != nil {
+		return -1, err
+	}
+
+	if !resp.Return {
+		return -1, fmt.Errorf("rest API show_bgp_max_as_limit GET failed: %s", resp.Reason)
```
CyrusJavan: GetAPIContext will already check the `return` field with your provided check function

goaviatrix/controller_bgp_max_as_limit_config.go
```diff
@@ -0,0 +1,74 @@
...
+	data := map[string]string{
+		"action": "show_bgp_max_as_limit",
+		"CID":    c.CID,
+	}
+
+	type BgpMaxAsLimitResponse struct {
+		Return bool   `json:"return"`
+		Reason string `json:"reason"`
+		Result string `json:"results"`
+	}
```
CyrusJavan: You do not need the Return or Reason fields, GetAPIContext will already check those fields using your check function. 

Also if you name the `Result` field `Results` then you don't need the json struct tags either. By default, the `encoding/json` library will assume the json field name is the same as the struct field name in all lowercase.

goaviatrix/controller_bgp_max_as_limit_config.go
```diff
@@ -0,0 +1,74 @@
...
+
+	var resp BgpMaxAsLimitResponse
+	err := c.GetAPIContext(ctx, &resp, data["action"], data, BasicCheck)
+	if err != nil {
+		return -1, err
```
CyrusJavan: In Go, it is standard to just return the zero value for a type if there is an error. For `int` that would just be `0`

CyrusJavan (changes requested):
Please also add an acceptance test.

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config.go
```diff
@@ -0,0 +1,91 @@
+package aviatrix
+
+import (
+	"context"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"github.com/terraform-providers/terraform-provider-aviatrix/goaviatrix"
```
CyrusJavan: Our module path has been updated, please change to `github.com/AviatrixSystems/terraform-provider-aviatrix/v2/goaviatrix`

mlin-aviatrix: Rebased with master and added this change.

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config_test.go
```diff
@@ -0,0 +1,94 @@
...
+	if skipAcc == "yes" {
+		t.Skip("Skipping Controller BGP Max AS Limit Config test as SKIP_CONTROLLER_BGP_MAX_AS_LIMIT_CONFIG is set")
+	}
+	msgCommon := ". Set SKIP_CONTROLLER_BGP_MAX_AS_LIMIT_CONFIG to yes to skip Controller BGP Max AS Limit Config tests"
+	resourceName := "aviatrix_controller_bgp_max_as_limit_config.test_bgp_max_as_limit"
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+			preAccountCheck(t, msgCommon)
```
CyrusJavan: I don't think we need the preAccountCheck.

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config.go
```diff
@@ -0,0 +1,91 @@
...
+func resourceAviatrixControllerBgpMaxAsLimitConfigRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*goaviatrix.Client)
+
+	if d.Id() != strings.Replace(client.ControllerIP, ".", "-", -1) {
+		return diag.Errorf("ID: %s does not match controller IP. Please provide correct ID for importing", d.Id())
+	}
+
+	maxAsLimit, err := client.GetControllerBgpMaxAsLimit(ctx)
+	if err != nil {
+		if err == goaviatrix.ErrNotFound {
```
CyrusJavan, line 56: You are checking for ErrNotFound, but your function never returns ErrNotFound.

goaviatrix/controller_bgp_max_as_limit_config.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	var resp BgpMaxAsLimitResponse
+	err := c.GetAPIContext(ctx, &resp, data["action"], data, BasicCheck)
+	if err != nil {
+		return 0, err
+	}
+
+	if resp.Results == "" {
+		return -1, nil
```
CyrusJavan: Don't use -1 as some special value, communicate what exactly went wrong through the `error` not the value. In this case I think you would return ErrNotFound here

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+func testAccControllerBgpMaxAsLimitConfigBasic() string {
+	return `
+resource "aviatrix_controller_bgp_max_as_limit_config" "test_bgp_max_as_limit" {
+	max_as_limit = 1
+}
+`
+}
+
+func testAccCheckControllerBgpMaxAsLimitConfigExists(n string) resource.TestCheckFunc {
```
CyrusJavan, line 53: This function needs to actually call `client.GetControllerBgpMaxAsLimit(context.Background())` and verify that it is configured

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config_test.go
```diff
@@ -0,0 +1,94 @@
...
+			continue
+		}
+
+		maxAsLimit, err := client.GetControllerBgpMaxAsLimit(context.Background())
+		if err != nil {
+			return fmt.Errorf("could not retrieve controller bgp max as limit config status: %v", err)
+		}
+		if maxAsLimit != -1 {
+			return errors.New("controller bgp max as limit is still enabled")
+		}
```
CyrusJavan: You can just check if 
```
if err == nil {
  return errors.New("controller bgp max as limit is still enabled")
}
```

CyrusJavan (changes requested):

CyrusJavan (changes requested):
Also please add your `SKIP_*` env var to test-infra/cmdExportOutput.sh

goaviatrix/controller_bgp_max_as_limit_config.go
mlin-aviatrix: I'm using -1 to signify that the feature is currently disabled. This is for an edge case where the user created the resource and somehow disabled the feature somewhere else (maybe through the UI). If I return an error it wouldn't be possible for them to update the resource. They would have to destroy it first then create a new one.

CyrusJavan: You can use a specific error to communicate this case, like ErrNotFound, in you Read function you would check for ErrNotFound.

user would not need to destroy the resource first, in the Read if you d.SetID(“”) then terraform considers the resource already destroyed

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config_test.go
```diff
@@ -0,0 +1,95 @@
...
+	client := testAccProviderVersionValidation.Meta().(*goaviatrix.Client)
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aviatrix_controller_bgp_max_as_limit_config" {
+			continue
+		}
+
+		_, err := client.GetControllerBgpMaxAsLimit(context.Background())
+		if err == nil || err != goaviatrix.ErrNotFound {
+			return fmt.Errorf("could not retrieve controller bgp max as limit config status: %v", err)
```
CyrusJavan: This error message is not correct, should be something like `controller bgp max as limit configured when it should be destroyed` I believe

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config_test.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		client := testAccProviderVersionValidation.Meta().(*goaviatrix.Client)
+
+		maxAsLimit, err := client.GetControllerBgpMaxAsLimit(context.Background())
+		if err != nil {
+			return fmt.Errorf("failed to get controller bgp max as limit config status: %v", err)
+		} else if maxAsLimit != 1 {
+			return fmt.Errorf("API returned the wrong value for controller bgp max as limit: expected %d but got %d", 1, maxAsLimit)
+		}
```
CyrusJavan: You don't need to check the value of maxAsLimit, that is already covered with the `resource.TestCheckResourceAttr(resourceName, "max_as_limit", "1")` above.

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_controller_bgp_max_as_limit_config_test.go
mlin-aviatrix: Would it make sense to have two different error messages here? One where the err is nil and another where the err is not nil but is not ErrNotFound.

CyrusJavan: I don't think it's necessary. Just a general "this thing was supposed to be destroyed but it's not" is fine

CyrusJavan (approved):

zhiqiangjin (approved):

PR #790, opened 2021-03-05
19016: GCP Firewall Instance and Association

CyrusJavan:
Added support for GCP Firewall Instance, Firewall
Instance Association, and GCP Transit GW zone.

### DIFF
--- a/aviatrix/utils.go
+++ b/aviatrix/utils.go
@@ -110,8 +110,8 @@ func extractTags(d *schema.ResourceData, cloudType int) (map[string]string, erro
 	if !ok {
 		return nil, nil
 	}
-	if !intInSlice(cloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.AZURE}) {
-		return nil, fmt.Errorf("adding tags is only supported for AWS, AWSGOV and AZURE, cloud_type must be 1, 256 or 8")
+	if !intInSlice(cloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.GCP, goaviatrix.AZURE}) {
+		return nil, fmt.Errorf("adding tags is only supported for AWS, AWSGOV and AZURE, cloud_type must be 1, 256, 4 or 8")
 	}
 
 	tagsMap := tags.(map[string]interface{})
--- a/aviatrix/utils.go
+++ b/aviatrix/utils.go
@@ -95,6 +95,11 @@ func intInSlice(needle int, haystack []int) bool {
 	return false
 }
 
+var (
+	tagMatcher    = regexp.MustCompile(`^[a-zA-Z0-9+\-=._ :/@ ]*$`)
+	gcpTagMatcher = regexp.MustCompile(`^[\p{Ll}\p{Lo}\p{N}_-]*$`)
+)
+
 func extractTags(d *schema.ResourceData, cloudType int) (map[string]string, error) {
 	tags, ok := d.GetOk("tags")
 	if !ok {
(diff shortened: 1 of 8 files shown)


goaviatrix/firewall_instance.go
```diff
@@ -180,6 +197,13 @@ func (c *Client) GetFirewallInstance(firewallInstance *FirewallInstance) (*Firew
 		return nil, errors.New("Rest API get_instance_by_id Get failed: " + data.Reason)
 	}
 	if data.Results.InstanceID == firewallInstance.InstanceID {
+		// Only try to decode if tags are not empty string
+		if string(data.Results.TagsMessage) != `""` {
+			err = json.Unmarshal(data.Results.TagsMessage, &data.Results.Tags)
+			if err != nil {
+				return nil, fmt.Errorf("json Decode get_instance_by_id failed: %v", err)
+			}
+		}
```
mlin-aviatrix, line 206: I submitted a mantis for this: 20558. I think the backend should return an empty dict instead of an empty string so we don't have the error when it tries to unmarshal a string into a map.

aviatrix/utils.go
```diff
@@ -110,8 +110,8 @@ func extractTags(d *schema.ResourceData, cloudType int) (map[string]string, erro
 	if !ok {
 		return nil, nil
 	}
-	if !intInSlice(cloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.AZURE}) {
-		return nil, fmt.Errorf("adding tags is only supported for AWS, AWSGOV and AZURE, cloud_type must be 1, 256 or 8")
+	if !intInSlice(cloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.GCP, goaviatrix.AZURE}) {
+		return nil, fmt.Errorf("adding tags is only supported for AWS, AWSGOV and AZURE, cloud_type must be 1, 256, 4 or 8")
```
mlin-aviatrix: Missing GCP in supported cloud names.

goaviatrix/firewall_instance.go
CyrusJavan: I agree. I will keep this workaround for now until 20558 is fixed.

aviatrix/utils.go
```diff
@@ -110,8 +100,8 @@ func extractTags(d *schema.ResourceData, cloudType int) (map[string]string, erro
 	if !ok {
 		return nil, nil
 	}
-	if !intInSlice(cloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.AZURE}) {
-		return nil, fmt.Errorf("adding tags is only supported for AWS, AWSGOV and AZURE, cloud_type must be 1, 256 or 8")
+	if !intInSlice(cloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.GCP, goaviatrix.AZURE}) {
+		return nil, fmt.Errorf("adding tags is only supported for AWS, AWSGOV, GCP and AZURE, cloud_type must be 1, 256, 4 or 8")
 	}
 
 	tagsMap := tags.(map[string]interface{})
```
mlin-aviatrix, line 112: GCP needs its own regex for tag validation. Something like ```regexp.Compile(`^[\p{Ll}\p{Lo}\p{N}_-]*$`)```. I copied the regex from Google's page, but I removed the check for tag length. Checking for tag length for other cloud types requires much more logic which I think is better left in the backend.

mlin-aviatrix (changes requested):

mlin-aviatrix (approved):

shuoliugit (approved):

zhiqiangjin (approved):

PR #791, opened 2021-03-05
17261: add support for management access from onprem

shuoliugit:
- create vpc attachment with edge_attachment -> remove, import, plan -> no diff
- change edge attachment -> remove, import, plan -> no diff
- delete edge attachment -> remove, import, plan -> no diff
- create vpc attachment without edge_attachment -> remove, import, plan -> no diff
- add edge attachment -> remove, import, plan -> no diff
- destroy
- try a non firenet domain -> error

### DIFF
--- a/aviatrix/resource_aviatrix_aws_tgw_vpc_attachment.go
+++ b/aviatrix/resource_aviatrix_aws_tgw_vpc_attachment.go
@@ -80,6 +80,14 @@ func resourceAviatrixAwsTgwVpcAttachment() *schema.Resource {
 				Default:     false,
 				Description: "Advanced option. If set to true, it disables automatic route propagation of this VPC to other VPCs within the same security domain.",
 			},
+			"edge_attachment": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "",
+				Description: "Edge attachment ID. To allow access to the private IP of the MGMT interface of the " +
+					"Firewalls, set this attribute to enable Management Access From Onprem. This feature advertises " +
+					"the Firewalls private MGMT subnet to your Edge domain.",
+			},
 		},
 	}
 }
@@ -98,6 +106,7 @@ func resourceAviatrixAwsTgwVpcAttachmentCreate(d *schema.ResourceData, meta inte
 		RouteTables:                  d.Get("route_tables").(string),
 		CustomizedRouteAdvertisement: d.Get("customized_route_advertisement").(string),
 		DisableLocalRoutePropagation: d.Get("disable_local_route_propagation").(bool),
+		EdgeAttachment:               d.Get("edge_attachment").(string),
 	}
 
 	isFirewallSecurityDomain, err := client.IsFirewallSecurityDomain(awsTgwVpcAttachment.TgwName, awsTgwVpcAttachment.SecurityDomainName)
@@ -111,12 +120,23 @@ func resourceAviatrixAwsTgwVpcAttachmentCreate(d *schema.ResourceData, meta inte
 	log.Printf("[INFO] Attaching vpc: %s to tgw %s", awsTgwVpcAttachment.VpcID, awsTgwVpcAttachment.TgwName)
 
 	if isFirewallSecurityDomain {
-		err := client.CreateAwsTgwVpcAttachmentForFireNet(awsTgwVpcAttachment)
+		err = client.CreateAwsTgwVpcAttachmentForFireNet(awsTgwVpcAttachment)
 		if err != nil {
 			return fmt.Errorf("failed to create Aviatrix Aws Tgw Vpc Attach for FireNet: %s", err)
 		}
+
+		if awsTgwVpcAttachment.EdgeAttachment != "" {
+			err = client.UpdateFirewallAttachmentAccessFromOnprem(awsTgwVpcAttachment)
+			if err != nil {
+				return fmt.Errorf("failed to enable firewall attachment access from onprem: %s", err)
+			}
+		}
 	} else {
-		err := client.CreateAwsTgwVpcAttachment(awsTgwVpcAttachment)
+		if awsTgwVpcAttachment.EdgeAttachment != "" {
+			return fmt.Errorf("management access from onprem only works for FireNet")
+		}
+
+		err = client.CreateAwsTgwVpcAttachment(awsTgwVpcAttachment)
 		if err != nil {
 			return fmt.Errorf("failed to create Aviatrix Aws Tgw Vpc Attach: %s", err)
 		}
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_aws_tgw_vpc_attachment.go
```diff
@@ -245,6 +266,38 @@ func resourceAviatrixAwsTgwVpcAttachmentUpdate(d *schema.ResourceData, meta inte
...
+
+		if oldEA != "" && newEA != "" {
+			awsTgwVpcAttachment.EdgeAttachment = ""
+
+			err := client.UpdateFirewallAttachmentAccessFromOnprem(awsTgwVpcAttachment)
+			if err != nil {
+				return fmt.Errorf("failed to disable firewall attachment access from onprem while updating: %s", err)
+			}
+
+			awsTgwVpcAttachment.EdgeAttachment = d.Get("edge_attachment").(string)
```
CyrusJavan: Don't need `d.Get` can just use `newEA`

goaviatrix/aws_tgw_vpc_attachment.go
```diff
@@ -381,3 +390,47 @@ func (c *Client) EditTgwSpokeVpcCustomizedRouteAdvertisement(awsTgwVpcAttachment
...
+	}
+
+	type AccessFromEdge struct {
+		AccessFromEdge []string `json:"access_from_edge"`
+	}
+
+	type Resp struct {
+		Return  bool             `json:"return"`
+		Results []AccessFromEdge `json:"results"`
+		Reason  string           `json:"reason"`
```
CyrusJavan: Return and Reason are unnecessary.

Also the json struct tags are unnecessary.

goaviatrix/aws_tgw_vpc_attachment.go
```diff
@@ -170,6 +171,14 @@ func (c *Client) GetAwsTgwVpcAttachment(awsTgwVpcAttachment *AwsTgwVpcAttachment
 		aTVA.CustomizedRouteAdvertisement = ""
 	}
 
+	firenetManagementDetails, err := c.GetFirenetManagementDetails(awsTgwVpcAttachment)
+	if err != nil {
+		return nil, errors.New("could not get firenet management details: " + err.Error())
+	}
+	if len(firenetManagementDetails) != 0 {
+		aTVA.EdgeAttachment = firenetManagementDetails[1]
```
CyrusJavan, line 179: Why the `1` index not `0`? Will this slice always have >=2 values in it?

goaviatrix/aws_tgw_vpc_attachment.go
```diff
@@ -170,6 +171,14 @@ func (c *Client) GetAwsTgwVpcAttachment(awsTgwVpcAttachment *AwsTgwVpcAttachment
 		aTVA.CustomizedRouteAdvertisement = ""
 	}
 
+	firenetManagementDetails, err := c.GetFirenetManagementDetails(awsTgwVpcAttachment)
+	if err != nil {
+		return nil, errors.New("could not get firenet management details: " + err.Error())
```
CyrusJavan: Can use `fmt.Errorf` formatting errors

goaviatrix/aws_tgw_vpc_attachment.go
```diff
@@ -381,3 +390,47 @@ func (c *Client) EditTgwSpokeVpcCustomizedRouteAdvertisement(awsTgwVpcAttachment
...
+	}
+
+	var data Resp
+
+	err := c.GetAPI(&data, params["action"], params, BasicCheck)
+	if err != nil {
+		return nil, err
+	}
+
+	return data.Results[0].AccessFromEdge, nil
```
CyrusJavan, line 433: Will `Results` never be empty? What if the attachment is deleted on the GUI, would this panic?

aviatrix/resource_aviatrix_aws_tgw_vpc_attachment.go
```diff
@@ -245,6 +266,38 @@ func resourceAviatrixAwsTgwVpcAttachmentUpdate(d *schema.ResourceData, meta inte
...
+
+			err = client.UpdateFirewallAttachmentAccessFromOnprem(awsTgwVpcAttachment)
+			if err != nil {
+				return fmt.Errorf("failed to enable firewall attachment access from onprem while updating: %s", err)
+			}
+		} else {
```
CyrusJavan, line 325: What about the case where we are going from "" -> "someID", do we need to validate again that this is a FireNet attachment?

Or would the API return an error? If that's the case we probably don't need to validate again

aviatrix/resource_aviatrix_aws_tgw_vpc_attachment.go
```diff
@@ -80,6 +80,14 @@ func resourceAviatrixAwsTgwVpcAttachment() *schema.Resource {
 				Default:     false,
 				Description: "Advanced option. If set to true, it disables automatic route propagation of this VPC to other VPCs within the same security domain.",
 			},
+			"edge_attachment": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "",
```
CyrusJavan: Is this `Default` necessary?

aviatrix/resource_aviatrix_aws_tgw_vpc_attachment.go
shuoliugit: API will return an error, but we can add a validation here.

goaviatrix/aws_tgw_vpc_attachment.go
shuoliugit: I thought about this again. If the attachment is removed manually, then an error will be returned.

goaviatrix/aws_tgw_vpc_attachment.go
shuoliugit: `0` is the security domain name, `1` is the attachment name.

aviatrix/resource_aviatrix_aws_tgw_vpc_attachment.go
```diff
@@ -245,8 +283,57 @@ func resourceAviatrixAwsTgwVpcAttachmentUpdate(d *schema.ResourceData, meta inte
...
+			SecurityDomainName: d.Get("security_domain_name").(string),
+			EdgeAttachment:     d.Get("edge_attachment").(string),
+		}
+
+		isFirewallSecurityDomain, err := client.IsFirewallSecurityDomain(awsTgwVpcAttachment.TgwName, awsTgwVpcAttachment.SecurityDomainName)
+		if err != nil {
+			if err == goaviatrix.ErrNotFound {
+				return fmt.Errorf("could not find Security Domain: " + awsTgwVpcAttachment.SecurityDomainName)
+			}
+			return fmt.Errorf(("could not find Security Domain due to: ") + err.Error())
```
CyrusJavan: We should be consistent in our `fmt.Errorf` usage. This should just be:
```
return fmt.Errorf("could not find Security Domain due to: %v", err)
```

CyrusJavan (changes requested):

CyrusJavan (approved):

zhiqiangjin (approved):

PR #793, opened 2021-03-06
20353: Add tags to gateway, spoke gateway and transit gateway

mlin-aviatrix:


aviatrix/resource_aviatrix_gateway.go
```diff
@@ -460,6 +462,12 @@ func resourceAviatrixGateway() *schema.Resource {
 				Optional:    true,
 				Description: "Enable jumbo frame support for Gateway. Valid values: true or false. Default value: true.",
 			},
+			"tags": {
+				Type:        schema.TypeMap,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "A map of tags to assign to the gateway.",
+			},
```
shuoliugit, line 478: consider using ConflictsWith

goaviatrix/tags.go
```diff
@@ -122,7 +161,7 @@ func (c *Client) DeleteTags(tags *Tags) error {
 	return nil
 }
 
-func (c *Client) AzureUpdateTags(tags *Tags) error {
+func (c *Client) UpdateTags(tags *Tags) error {
 	tags.CID = c.CID
```
shuoliugit, line 109: I haven't tried it myself, but I remember `update_resource_tags` sends the `action=replace` to the backend cli. AWS and GCP don't support `replace`.

mlin-aviatrix: AWS now supports replace, I checked the code in the backend and tested it.  I haven't tried with GCP, I will check.

mlin-aviatrix: From my research, GCP actually has two kinds of "tagging". First are labels, which is a map of key value pairs just like the tags for other cloud types. Second are network tags, which is a list of strings. The backend supports GCP tags, not labels. The API does not support GCP tags.

shuoliugit (approved):

goaviatrix/tags.go
```diff
@@ -28,6 +29,15 @@ type TagAPIResp struct {
...
 }
 
+func (tags *Tags) ComputeTagList() {
+	tagList := make([]string, 0, len(tags.Tags))
+	for key, val := range tags.Tags {
+		tagList = append(tagList, key+":"+val)
+	}
+	tagListStr := strings.Join(tagList, ",")
+	tags.TagList = tagListStr
+}
```
CyrusJavan: Could we avoid having this mutator method? I find it difficult to understand, instead can we just have a pure function that converts tags map[string]string -> string, for example
```
func TagMapToString(tags map[string]string) string
```

and just use it like 
```
tags.Tags = tagsMap
tags.TagList = TagMapToString(tagsMap)
```

goaviatrix/tags.go
```diff
@@ -91,6 +104,32 @@ func (c *Client) GetTags(tags *Tags) ([]string, error) {
...
+	if !data.Return {
+		return fmt.Errorf("rest API list_resource_tags Post failed: %s", data.Reason)
+	}
+
+	if tagsMap, ok := data.Results["usr_tags"]; ok {
+		tags.Tags = tagsMap
+	}
+	return nil
+}
+
```
CyrusJavan: Any reason why we cannot use `GetAPI` here?

goaviatrix/tags.go
```diff
@@ -91,6 +104,32 @@ func (c *Client) GetTags(tags *Tags) ([]string, error) {
 	return nil, nil
 }
 
+func (c *Client) GetTagsMap(tags *Tags) error {
```
CyrusJavan: Again, can we avoid methods that only mutate an objects state? Instead just return the tags and leave the passed in object unmodified

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1343,24 +1363,33 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
...
 
-		var tagListStr []string
 		if _, ok := d.GetOk("tag_list"); ok {
+			tagList := make([]string, 0, len(tags.Tags))
+			for key, val := range tags.Tags {
+				str := key + ":" + val
+				tagList = append(tagList, str)
+			}
+
 			tagList1 := d.Get("tag_list").([]interface{})
```
CyrusJavan: Can we think of a better name than `tagList1`? Maybe `tagsFromUserConfig` or something more descriptive

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1727,59 +1756,40 @@ func resourceAviatrixTransitGatewayUpdate(d *schema.ResourceData, meta interface
...
 	}
 
-	if gateway.CloudType == goaviatrix.AWS || gateway.CloudType == goaviatrix.AWSGOV || gateway.CloudType == goaviatrix.AZURE {
-		if d.HasChange("tag_list") {
-			tags := &goaviatrix.Tags{
-				ResourceType: "gw",
-				ResourceName: d.Get("gw_name").(string),
-				CloudType:    gateway.CloudType,
-			}
+	if (d.HasChange("tag_list") || d.HasChange("tags")) && !intInSlice(gateway.CloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.AZURE}) {
```
CyrusJavan: Could we refactor this Update logic to all live under a conditional, I think it will be much easier to understand 
```
if d.HasChange("tag_list") || d.HasChange("tags") {
  // do validation
  // get user provided tags
  // call the update API
}
```

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1704,56 +1734,44 @@ func resourceAviatrixGatewayUpdate(d *schema.ResourceData, meta interface{}) err
...
+	if d.HasChange("tag_list") || d.HasChange("tags") {
+		if !intInSlice(gateway.CloudType, []int{goaviatrix.AWS, goaviatrix.AWSGOV, goaviatrix.AZURE}) {
+			return fmt.Errorf("failed to update gateway: adding tags is only supported for AWS, AWSGOV and AZURE, cloud_type must be set to 1, 256 or 8")
+		}
+
+		_, tagListOk := d.GetOk("tag_list")
+		_, tagsOk := d.GetOk("tags")
+		if tagsOk && tagListOk {
+			return fmt.Errorf("failed to update gateway: only one of tag_list and tags is allowed at the same time")
+		}
```
CyrusJavan: Is this check needed in the Create function as well? Or does `ConflictsWith` cover it?

docs/resources/aviatrix_gateway.md
```diff
@@ -260,10 +260,10 @@ The following arguments are supported:
...
 * `eip` - (Optional) Specified EIP to use for gateway creation. Required when `allocate_new_eip` is false.  Available in Controller version 3.5+. Only supported for AWS and GCP gateways.
-* `tag_list` - (Optional) Tag list of the gateway instance. Only available for AWS, AWSGov and Azure gateways. Example: ["key1:value1", "key2:value2"].
 * `enable_vpc_dns_server` - (Optional) Enable VPC DNS Server for gateway. Currently only supported for AWS and AWSGov gateways. Valid values: true, false. Default value: false.
 * `zone` - (Optional) Availability Zone. Only available for Azure and Public Subnet Filtering gateway. Available for Azure as of provider version R2.17+.
 * `enable_jumbo_frame` - (Optional) Enable jumbo frames for this gateway. Default value is true.
+* `tags` - (Optional) Map of tags to assign to the gateway. Only available for AWS, AWSGov and Azure gateway. Allowed characters are: letters, spaces, and numbers plus the following special characters: + - = . _ : @. Example: {"key1" = "value1", "key2" = "value2"}.
```
CyrusJavan: We should note that the requirements for the tags vary by cloud type.

goaviatrix/tags.go
```diff
@@ -50,81 +48,81 @@ func (c *Client) AddTags(tags *Tags) error {
...
+		"cloud_type":    strconv.Itoa(tags.CloudType),
+		"resource_type": tags.ResourceType,
+		"resource_name": tags.ResourceName,
 	}
-	if !data.Return {
-		return nil, errors.New("Rest API list_resource_tags Post failed: " + data.Reason)
+	var resp TagAPIResp
+	err := c.GetAPI(&resp, data["Action"], data, BasicCheck)
+	if err != nil {
+		return nil, fmt.Errorf("HTTP Get list_resource_tags failed: %v", err)
```
CyrusJavan: You can just return the naked err from GetAPI since it already nicely formats the error

goaviatrix/tags.go
```diff
@@ -50,81 +48,81 @@ func (c *Client) AddTags(tags *Tags) error {
...
+		"resource_type": tags.ResourceType,
+		"resource_name": tags.ResourceName,
 	}
-	resp, err := c.HTTPClient.Do(req)
+
+	var resp TagAPIResp
+	err := c.GetAPI(&resp, data["Action"], data, BasicCheck)
 	if err != nil {
-		return errors.New("HTTP Post delete_resource_tags failed: " + err.Error())
+		return nil, fmt.Errorf("HTTP Get list_resource_tags failed: %v", err)
```
CyrusJavan: Just return the naked err

goaviatrix/tags.go
```diff
@@ -50,81 +48,81 @@ func (c *Client) AddTags(tags *Tags) error {
...
 	}
-	var data APIResp
-	buf := new(bytes.Buffer)
-	buf.ReadFrom(resp.Body)
-	bodyString := buf.String()
-	bodyIoCopy := strings.NewReader(bodyString)
-	if err = json.NewDecoder(bodyIoCopy).Decode(&data); err != nil {
-		return errors.New("Json Decode delete_resource_tags failed: " + err.Error() + "\n Body: " + bodyString)
+
+	if !resp.Return {
```
CyrusJavan: Using `BasicCheck` with GetAPI already checks the Return boolean

goaviatrix/tags.go
```diff
@@ -50,81 +48,81 @@ func (c *Client) AddTags(tags *Tags) error {
...
+}
+
+func (c *Client) UpdateTags(tags *Tags) error {
 	tags.CID = c.CID
 	tags.Action = "update_resource_tags"
 
 	return c.PostAPI(tags.Action, tags, BasicCheck)
 }
+
+func TagsMapToString(tagsMap map[string]string) string {
```
CyrusJavan: Since this helper function is only used in the `aviatrix` package perhaps we should move it there

aviatrix/resource_aviatrix_gateway.go
mlin-aviatrix: ConflictsWith covers it. I forgot to remove it after I added ConflictsWith.

CyrusJavan (approved):

zhiqiangjin (approved):

PR #794, opened 2021-03-06
20560: Improve time for terraform refresh for gateway, spoke and transit

zhiqiangjin:
Backend added a new parameter for list_vpcs_summary to only get gateway info of the specified gateway. This avoids looping all the gateways and saves time.


arvindsree (changes requested):
we are still calling 7 or 8 APIs. All of this information needs to be consolidated into one API

CyrusJavan:
@arvindsree I implemented a prototype consolidated API for transit_gateway (See PR #13063 in the backend repo). If it looks good I will do the same for gateway and spoke_gateway.

arvindsree:
> @arvindsree I implemented a prototype consolidated API for transit_gateway (See PR #13063 in the backend repo). If it looks good I will do the same for gateway and spoke_gateway.

Sounds good. It could be a single API and can return the additonal specific information required for transit/spoke by checking whether it is a spoke/transit gateway

zhiqiangjin:
Backend is considering deprecating list_vpc_by_name since it is duplicate to list_vpcs_summary in next release. And they will need to merge output of list_vpc_by_name to list_vpcs_summary. I do not think we need a new API since everything can be incorporated into list_vpcs_summary. And list_vpcs_summary is already optimized, this fix can be available in 2.18.1/6.3-patch.

arvindsree (approved):

PR #875, opened 2021-04-02
20917: added support for fortigate vendor integration

shuoliugit:


aviatrix/data_source_aviatrix_firenet_vendor_integration.go
```diff
@@ -158,6 +176,8 @@ func dataSourceAviatrixFireNetVendorIntegrationRead(d *schema.ResourceData, meta
 		}
 	}
 
+	d.Set("password", nil)
+	d.Set("api_token", nil)
```
CyrusJavan: Why set these values at all?

shuoliugit: right, state file can have sensitive data... Will remove these.

CyrusJavan (approved):

zhiqiangjin (approved):

PR #877, opened 2021-04-02
21150: Fix issue where terraform apply to reorder security domains inside aviatrix_aws_tgw does do something sometimes

zhiqiangjin:
DifferenceSlice is used to find the difference between two list of connection policy pairs.
"connection policy" for aviatrix_aws_tgw is symmetric meaning if a is connected to b, b is also connected to a.
Example: a is [[domain1, domain2][domain3, domain4]], and b is [[domain1, domain2][domain4, domain3]]. There is no diff between a and b because domain1 and domain2, domain3 and domain4 are connected for both a and b.


CyrusJavan:
Could we change the name `DifferenceSlice` to something that better describes what this does? Maybe `DifferencePairSlice` or `DifferencePairs`. And also update the comment describing the function 
`// DifferencePairs returns all pairs in a that are not in b. If b contains any elements that are non-pairs (len != 2) then a is returned`

CyrusJavan (approved):
That is a good name!

PR #879, opened 2021-04-06
20888: implemented aviatrix_aws_tgw_security_domain_connection

shuoliugit:


aviatrix/resource_aviatrix_aws_tgw_security_domain_connection.go
```diff
@@ -0,0 +1,177 @@
...
+
+	if err := client.CreateDomainConnection(awsTgw, sourceDomainName, destinationDomainName); err != nil {
+		return diag.Errorf("could not create the security domain connection: %v", err)
+	}
+
+	if strings.Compare(sourceDomainName, destinationDomainName) == -1 {
+		d.SetId(awsTgw.Name + "~" + sourceDomainName + "~" + destinationDomainName)
+	} else {
+		d.SetId(awsTgw.Name + "~" + destinationDomainName + "~" + sourceDomainName)
+	}
```
CyrusJavan: Nit: From the implementation of strings.Compare, it seems we should just use `sourceDomainName < destinationDomainName`

```go
// Compare returns an integer comparing two strings lexicographically.
// The result will be 0 if a==b, -1 if a < b, and +1 if a > b.
//
// Compare is included only for symmetry with package bytes.
// It is usually clearer and always faster to use the built-in
// string comparison operators ==, <, >, and so on.
func Compare(a, b string) int {
	// NOTE(rsc): This function does NOT call the runtime cmpstring function,
	// because we do not want to provide any performance justification for
	// using strings.Compare. Basically no one should use strings.Compare.
	// As the comment above says, it is here only for symmetry with package bytes.
	// If performance is important, the compiler should be changed to recognize
	// the pattern so that all code doing three-way comparisons, not just code
	// using strings.Compare, can benefit.
	if a == b {
		return 0
	}
	if a < b {
		return -1
	}
	return +1
}
```

aviatrix/resource_aviatrix_aws_tgw_security_domain_connection.go
```diff
@@ -0,0 +1,177 @@
...
+
+		if sourceDomainName == destinationDomainName {
+			return diag.Errorf("two domains cannot be the same")
+		}
+
+		d.Set("tgw_name", tgwName)
+		d.Set("domain_name1", sourceDomainName)
+		d.Set("domain_name2", destinationDomainName)
+
+		if strings.Compare(sourceDomainName, destinationDomainName) == -1 {
```
CyrusJavan: Same as above

aviatrix/resource_aviatrix_aws_tgw_security_domain_connection.go
```diff
@@ -0,0 +1,177 @@
...
+		AwsTgwName: tgwName,
+	}
+
+	securityDomainDetails, err := client.GetSecurityDomainDetails(ctx, securityDomain)
+	if err == goaviatrix.ErrNotFound {
+		d.SetId("")
+		return nil
+	}
+	if err != nil {
+		return diag.Errorf("couldn't get the details of the security domain %s due to %v", sourceDomainName, err)
```
CyrusJavan: Let's stick with the same pattern for errors found in other parts of the code, using a colon to separate the error
```go
return diag.Errorf("could not get the details of the security domain %s: %v", sourceDomainName, err)
```

aviatrix/resource_aviatrix_aws_tgw_security_domain_connection.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	for _, sd := range securityDomainDetails.ConnectedDomain {
+		if sd == destinationDomainName {
+			if strings.Compare(sourceDomainName, destinationDomainName) == -1 {
```
CyrusJavan: Same as above

aviatrix/resource_aviatrix_aws_tgw_security_domain_connection_test.go
```diff
@@ -0,0 +1,154 @@
...
+    	aviatrix_aws_tgw_security_domain.Default_Domain,
+    	aviatrix_aws_tgw_security_domain.Shared_Service_Domain,
+    	aviatrix_aws_tgw_security_domain.Aviatrix_Edge_Domain
+  ]
+}
+resource "aviatrix_aws_tgw_security_domain_connection" "test" {
+	tgw_name = aviatrix_aws_tgw.test.tgw_name
+	domain_name1 = aviatrix_aws_tgw_security_domain.Default_Domain.name
+	domain_name2 = aviatrix_aws_tgw_security_domain.test.name
+}
```
CyrusJavan: Line up equals signs.

docs/resources/aviatrix_aws_tgw_security_domain.md
```diff
@@ -39,6 +39,30 @@ resource "aviatrix_aws_tgw_security_domain" "Aviatrix_Edge_Domain" {
...
+  tgw_name = aviatrix_aws_tgw.test.tgw_name
+  domain_name1 = aviatrix_aws_tgw_security_domain.Default_Domain.name
+  domain_name2 = aviatrix_aws_tgw_security_domain.Shared_Service_Domain.name
+}
+
+resource "aviatrix_aws_tgw_security_domain_connection" "default_sd_conn2" {
+  tgw_name = aviatrix_aws_tgw.test.tgw_name
+  domain_name1 = aviatrix_aws_tgw_security_domain.Aviatrix_Edge_Domain.name
+  domain_name2 = aviatrix_aws_tgw_security_domain.Default_Domain.name
+}
```
CyrusJavan: This one is repeat from above

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_aws_tgw_security_domain_connection.go
```diff
@@ -0,0 +1,177 @@
...
+		Name: d.Get("tgw_name").(string),
+	}
+
+	securityDomain := &goaviatrix.SecurityDomain{
+		Name:       sourceDomainName,
+		AwsTgwName: awsTgw.Name,
+	}
+
+	securityDomainDetails, err := client.GetSecurityDomainDetails(ctx, securityDomain)
+	if err != nil && err != goaviatrix.ErrNotFound {
```
CyrusJavan: What if err == ErrNotFound?

CyrusJavan (approved):

zhiqiangjin (approved):

PR #886, opened 2021-04-09
21111: Update case of the tunnel_protocol values in aviatrix_transit_external_device_conn to insensitive

zhiqiangjin:


aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -67,8 +67,8 @@ func resourceAviatrixTransitExternalDeviceConn() *schema.Resource {
 				Optional:     true,
 				Default:      "IPsec",
 				ForceNew:     true,
-				Description:  "Tunnel Protocol. Valid values: 'IPsec', 'GRE' or 'LAN'. Default value: 'IPsec'.",
-				ValidateFunc: validation.StringInSlice([]string{"IPsec", "GRE", "LAN"}, false),
+				Description:  "Tunnel Protocol. Valid values: 'IPsec', 'GRE' or 'LAN'. Default value: 'IPsec'. Case insensitive.",
```
CyrusJavan, line 70: Should we add a diff suppress func so that we get no diffs if someone uses `tunnel_protocol = "lan"` for example?

zhiqiangjin: Added.

shuoliugit (approved):

PR #900, opened 2021-04-16
21003: Only allow aws_gateway_role_app and aws_gateway_role_ec2 to be…

mlin-aviatrix:
… configured with AWS (1) cloud type and awsIam enabled


mlin-aviatrix:
I think we should prevent users from using the wrong attributes for the rest of the attributes as well. But that should be a different mantis.

zhiqiangjin (approved):

CyrusJavan (approved):

PR #901, opened 2021-04-16
21779: Give warning when aviatrix account fails audit in read

mlin-aviatrix:
Changed CRUD functions in aviatrix account to include context so that a warning can be given on audit failure. I only use the context in the new API call.


mlin-aviatrix:
I didn't change the account data source because terraform configurations using the data source will probably have resources using the account, which will error when credentials have issues.

CyrusJavan (dismissed):

CyrusJavan:
conflicts

zhiqiangjin (approved):

PR #873, opened 2021-04-01
20375: Site2Cloud event triggered HA

CyrusJavan:
Add Event Triggered HA option to all site2cloud/IPSec
based connections.

Updated these resources:
- site2cloud
- transit_external_device_conn
- vgw_conn
- device_transit_gateway_attachment

Later I need to also add to
- device_virtual_wan_attachment (currently not working on 6.4)
- cloudn_transit_gateway_attachment (not implemented yet)

PR #902, opened 2021-04-16
20985: Add single ip ha to site2cloud

mlin-aviatrix:


CyrusJavan (changes requested):
Can you check if this feature is specifically for the site2cloud connection only OR for all site2cloud (IPsec) type connections, which would include
- site2cloud
- transit_external_device_conn
- vgw_conn
- device_transit_gateway_attachment

Most likely it is for all these resources so this would need to be added to all of them. See #873 for example.

zhiqiangjin (dismissed):

shuoliugit (dismissed):

mlin-aviatrix:
This is only for site2cloud for now. Not the other IPSec type connections

zhiqiangjin (approved):

CyrusJavan (approved):

PR #904, opened 2021-04-16
20991: added tunnel ip parameters for s2c

shuoliugit:


aviatrix/resource_aviatrix_site2cloud.go
```diff
@@ -592,8 +638,12 @@ func resourceAviatrixSite2CloudRead(d *schema.ResourceData, meta interface{}) er
 	if tunnelName == "" || vpcID == "" {
 		id := d.Id()
 		log.Printf("[DEBUG] Looks like an import, no tunnel name or vpc id names received. Import Id is %s", id)
-		d.Set("connection_name", strings.Split(id, "~")[0])
-		d.Set("vpc_id", strings.Split(id, "~")[1])
+		parts := strings.Split(id, "~~")
```
CyrusJavan: I think we need to split on single tilde ~, not double ~~.

```
## Import
**site2cloud** can be imported using the `connection_name` and `vpc_id`, e.g.

$ terraform import aviatrix_site2cloud.test connection_name~vpc_id

```

CyrusJavan (changes requested):

CyrusJavan (approved):

zhiqiangjin (approved):

PR #910, opened 2021-04-21
21687: added support for OCI FireNet

shuoliugit:


CyrusJavan (approved):
Probably want to also check on firenet, firewall_instance, firewall_instance_association and data.firewall_vendor_integration resources, to see if they need any changes for OCI.

zhiqiangjin (approved):

PR #928, opened 2021-05-01
22021/22022: Allow route based connection for single IP HA and don't …

mlin-aviatrix:
…require backup gateway remote IP


mlin-aviatrix:
backup_gateway_remote_ip needs to be computed so there is no diff when single IP HA is enabled and the remote ip is not provided. The provider will still error when backup_gateway_remote_ip is empty and HA is enabled and single IP HA is disabled so this should not effect the way things work currently.

zhiqiangjin (approved):

PR #929, opened 2021-05-01
fix20592: check default domains when changing manage_security_domain from false to true

shuoliugit:
Problem:
For example, a TGW was created with manage_security_domain set to false, and not all default domains were created. 
If change manage_security_domain to true, it will crash.

Solution:
in update()
1. when dealing with existing domains in DB, add default domains to avoid crash
2. add a check for default domains, error out if one or more default domains are missing
3. if default domains were not created, create default domains first

### DIFF
--- a/aviatrix/resource_aviatrix_aws_tgw.go
+++ b/aviatrix/resource_aviatrix_aws_tgw.go
@@ -1043,8 +1043,19 @@ func resourceAviatrixAWSTgwUpdate(d *schema.ResourceData, meta interface{}) erro
 				}
 			}
 
+			defaultDomainsWithCreation := []string{"Aviatrix_Edge_Domain", "Default_Domain", "Shared_Service_Domain"}
+
+			// Since manage_security_domain is changed from false to true, domainsOld may not contain default domains.
+			// Missing default domains (fillers) are added here temporarily, otherwise ValidateAWSTgwDomains will crash.
+			fillersForDomainsOld := goaviatrix.Difference(defaultDomainsWithCreation, domainsOld)
+			if len(fillersForDomainsOld) != 0 {
+				domainsOld = append(domainsOld, fillersForDomainsOld...)
+			}
+
 			domainsToCreateOld, domainConnPolicyOld, domainConnRemoveOld, _ := client.ValidateAWSTgwDomains(domainsOld,
 				domainConnOld, attachedVPCOld)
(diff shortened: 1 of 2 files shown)


inc1t3Ful:
Could you explain what point 1 means?

What is the intended behavior if TGW manage is false. the default domains are created outside. Update TGW to manage true ? (where the TF file of the TGW resource does NOT have domains written in-line, just updating manage attr to true)

What is the intended behavior if TGW manage is false. no domains are created outside at all. Update TGW to manage true? (where TF file of TGW does NOT have domains written in-line, just updating manage attr to true)

shuoliugit:
> Could you explain what point 1 means?

`domainsToCreateOld, domainConnPolicyOld, domainConnRemoveOld, _ := client.ValidateAWSTgwDomains(domainsOld, domainConnOld, attachedVPCOld)`
The crash was caused by this line. If domainsOld doesn't contain any domain, ValidateAWSTgwDomains will crash. When ValidateAWSTgwDomains returns domainsToCreateOld, it will ignore default domains, so it's safe to add default domains to domainsOld to avoid crash. 

> What is the intended behavior if TGW manage is false. the default domains are created outside. Update TGW to manage true ? (where the TF file of the TGW resource does NOT have domains written in-line, just updating manage attr to true)

After changing manage_security_domains from false to true, if resource aws_tgw does not define domains inside, it will error out. Users have to define at least the three default domains inside, because they are using aws_tgw as before.

What users should do is: 
1. remove the state (only state, not resource) of all the security domain related resources outside 
2. change manage_security_domains from false to true 
3. define domains, connections, etc. inside. Domains must contain default domains. If the in-line domains don't match with domains created outside, only the in-line domains will be kept. 
4. apply

> What is the intended behavior if TGW manage is false. no domains are created outside at all. Update TGW to manage true? (where TF file of TGW does NOT have domains written in-line, just updating manage attr to true)

It will error out. As stated above, they are using aws_tgw in the old way, so at least default domains are required.

zhiqiangjin:
Customers are encouraged to use standalone resource aviatrix_aws_tgw_security_domain instead of inside aviatrix_aws_tgw in 6.4/2.19.0 and future releases. So, changing manage_security_domain from false to true should be a rare care. However, it is good to cover it.

zhiqiangjin (approved):

PR #942, opened 2021-05-06
22226: add phase1_remote_identifier to transit_external_device_conn

shuoliugit:


aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -297,6 +297,18 @@ func resourceAviatrixTransitExternalDeviceConn() *schema.Resource {
...
 			},
+			"phase1_remote_identifier": {
+				Type:     schema.TypeString,
+				Optional: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					ip := d.Get("remote_gateway_ip").(string)
+					o, n := d.GetChange("phase1_remote_identifier")
+					return n.(string) == ip && o.(string) == ip
+
+				},
```
CyrusJavan: Is this necessary? If `n.(string) == ip && o.(string) == ip` that implies that `n.(string) == o.(string)` so there would be no diff anyway

aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -722,6 +753,18 @@ func resourceAviatrixTransitExternalDeviceConnRead(d *schema.ResourceData, meta
...
 
+	site2cloud := &goaviatrix.Site2Cloud{
+		TunnelName: externalDeviceConn.ConnectionName,
+		VpcID:      externalDeviceConn.VpcID,
+	}
+	s2c, err := client.GetSite2CloudConnDetail(site2cloud)
+	if err != nil {
+		return fmt.Errorf("couldn't find Aviatrix Site2Cloud: %s, %#v", err, s2c)
+	}
+	if s2c != nil {
```
CyrusJavan: Why do we check if s2c != nil ? We already checked if the err != nil above

CyrusJavan (changes requested):

aviatrix/resource_aviatrix_transit_external_device_conn.go
shuoliugit: It looks unnecessary, but I had to use it when I added this feature to s2c. Following are the test cases I used before. If I don't use this DiffSuppressFunc, there'll be diff for some test cases.

```
use remote_gateway_ip = 1.2.3.4
1.
Create s2c without using phase1_remote_identifier
Change phase1_remote_identifier to 1.2.3.4 -> no diff
Change phase1_remote_identifier to 3.3.3.3 -> diff -> set to 3.3.3.3 after apply
2.
Create s2c with phase1_remote_identifier = 1.2.3.4
Delete phase1_remote_identifier -> no diff
Change phase1_remote_identifier to 1.2.3.4 -> no diff
Change phase1_remote_identifier to 3.3.3.3 -> diff -> set to 3.3.3.3 after apply
3.
Create s2c with phase1_remote_identifier = 3.3.3.3
Delete phase1_remote_identifier -> diff -> set to 1.2.3.4 after apply

Destroy
```

CyrusJavan (approved):

mlin-aviatrix (approved):

zhiqiangjin (approved):

PR #953, opened 2021-05-11
22315: Refactor goaviatrix client functions to use GetAPI and PostAPI

CyrusJavan:


goaviatrix/utils.go
```diff
@@ -250,3 +251,19 @@ func ValidateAttachedVPCsForCustomizedRoutes(a, b [][]string) ([][]string, [][]s
...
+	tag := "map"
+	typ := v.Type()
+	for i := 0; i < v.NumField(); i++ {
+		fi := typ.Field(i)
+		if tagv := fi.Tag.Get(tag); tagv != "" && tagv != "-" {
+			out[tagv] = v.Field(i).String()
+		}
+	}
+	return out
+}
```
shuoliugit: I'm still not sure if we should add a new "map" tag in the struct. For me, it'll be easier to maintain if in all the functions we use the same format like:
```
form := map[string]string{
    "a": a,
    "b": b,
}

PostAPI or GetAPI
```

I would like to keep only "json" tag in the struct. What do you guys think? @zhiqiangjin @CyrusJavan @mlin-aviatrix

CyrusJavan: I think it depends, sometimes the "form" tag is nice for APIs that take a lot of fields. Always using map[string]string for GetAPI makes sense I think though because GET APIs usually do not take many parameters, usually just a name or ID.

CyrusJavan: I agree the "map" tag is probably unnecessary. I think this was an experiment I tried in the past to see if it would be cleaner 😄 . I will update this PR to remove use of "map" tag and toMap function.

goaviatrix/aws_tgw.go
```diff
@@ -700,34 +513,26 @@ func (c *Client) GetTransitGwFromVpcID(awsTgw *AWSTgw, gateway *Gateway) (*Gatew
...
-		if strings.Contains(data.Reason, "does not exist") {
-			return nil, ErrNotFound
+	form := map[string]string{
+		"CID":      c.CID,
+		"action":   "list_tgw_details",
+		"tgw_name": awsTgw.Name,
+	}
+	check := func(action, reason string, ret bool) error {
+		if !ret {
+			if strings.Contains(data.Reason, "does not exist") {
```
shuoliugit: should use `reason` instead of `data.Reason`

goaviatrix/aws_tgw_peering.go
```diff
@@ -26,51 +23,28 @@ type AwsTgwPeeringAPIResp struct {
...
+		"CID":      c.CID,
+		"action":   "list_peered_tgw_names",
+		"tgw_name": awsTgwPeering.TgwName1,
 	}
-	if !data.Return {
-		if strings.Contains(data.Reason, "does not exist") {
-			return ErrNotFound
+	check := func(action, reason string, ret bool) error {
+		if !ret {
+			if strings.Contains(data.Reason, "does not exist") {
```
shuoliugit: should use `reason` instead of `data.Reason`

goaviatrix/aws_tgw_peering_domain_conn.go
```diff
@@ -30,63 +27,36 @@ type ConnectedRouteDomainDetail struct {
...
+		"action":            "list_connected_route_domains",
+		"tgw_name":          domainConn.TgwName1,
+		"route_domain_name": domainConn.DomainName1,
 	}
-	if !data.Return {
-		if strings.Contains(data.Reason, "does not exist") {
-			return ErrNotFound
+	check := func(action, reason string, ret bool) error {
+		if !ret {
+			if strings.Contains(data.Reason, "does not exist") {
```
shuoliugit: should use `reason` instead of `data.Reason`

shuoliugit (approved):

zhiqiangjin (approved):

PR #956, opened 2021-05-14
22358: Fix phase1_remote_identifier crash

CyrusJavan:


CyrusJavan:
Duplicate Mantis, Shuo will make a PR to fix under the mantis 22306

PR #987, opened 2021-06-07
22354: add Availability Domain and Fault Domain to OCI gateway and firewall

shuoliugit:
also fixed a bug that HA can't be created during update in gateway

### DIFF
--- a/aviatrix/data_source_aviatrix_gateway.go
+++ b/aviatrix/data_source_aviatrix_gateway.go
@@ -300,6 +300,16 @@ func dataSourceAviatrixGateway() *schema.Resource {
 				Computed:    true,
 				Description: "The IPSec tunnel down detection time for the gateway.",
 			},
+			"availability_domain": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Availability domain for OCI.",
+			},
+			"fault_domain": {
+				Type:        schema.TypeString,
+				Computed:    true,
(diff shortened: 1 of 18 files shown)


CyrusJavan (commented):
Looks good. My question is whether this will impact existing OCI users? Looks like when user upgrades and they have OCI gateways, their next apply will be: 
```
Terraform wants to update:
resource gw {
   availability_domain = "domain" -> "" forces replacement
   fault_domain = "domain" -> "" forces replacement
}
```
So to avoid destroying their gateways, the user will need to edit their config to input the correct domains. Maybe to avoid that we could use Optional/Computed so that existing users are not impacted? What do you think?

shuoliugit:
> Looks good. My question is whether this will impact existing OCI users? Looks like when user upgrades and they have OCI gateways, their next apply will be:
> 
> ```
> Terraform wants to update:
> resource gw {
>    availability_domain = "domain" -> "" forces replacement
>    fault_domain = "domain" -> "" forces replacement
> }
> ```
> 
> So to avoid destroying their gateways, the user will need to edit their config to input the correct domains. Maybe to avoid that we could use Optional/Computed so that existing users are not impacted? What do you think?

AD/FD are optional/computed in gateway/transit/spoke/firewall now.

CyrusJavan (approved):

zhiqiangjin (approved):

PR #993, opened 2021-06-08
typo in tag APIs

CyrusJavan:


goaviatrix/tags.go
```diff
@@ -82,7 +82,7 @@ func (c *Client) GetTagsMap(tags *Tags) (map[string]string, error) {
 	}
 
 	var resp TagAPIResp
-	err := c.GetAPI(&resp, data["Action"], data, BasicCheck)
+	err := c.GetAPI(&resp, data["action"], data, BasicCheck)
```
mlin-aviatrix: GetTagsMap is not being used anymore because tags are gotten from list_vpcs_summary

shuoliugit (dismissed):

goaviatrix/tags.go
CyrusJavan: Removed it.

You would think the `deadcode` linter would have noticed that 🤔

zhiqiangjin (approved):

PR #1009, opened 2021-06-15
22864: add lists of OCI AD/FD to aviatrix_vpc resource and data source

shuoliugit:


aviatrix/data_source_aviatrix_vpc.go
```diff
@@ -163,6 +163,22 @@ func dataSourceAviatrixVpc() *schema.Resource {
 				Computed:    true,
 				Description: "Azure vnet resource ID.",
 			},
+			"availability_domains": {
+				Type:        schema.TypeSet,
```
CyrusJavan: Do you think we should use TypeList instead? That way users can use indexes to access the values
```hcl
availability_domain = data.aviatrix_vpc.A.availability_domains[0]
```

docs/data-sources/aviatrix_vpc.md
```diff
@@ -57,3 +57,5 @@ In addition to all arguments above, the following attributes are exported:
   * `cidr` - Private subnet CIDR.
   * `name` - Private subnet name.
   * `subnet_id` - Private subnet ID.
+* `availability_domains` - List of OCI availability domains.
```
CyrusJavan, line 60: Should we update the OCI examples in the gateway documentation to show the use of these values?

CyrusJavan (approved):

zhiqiangjin (approved):

PR #1012, opened 2021-06-20
22844: use full vpc_id for GCP in transit_external_device_conn + fixes

shuoliugit:
For `aviatrix_transit_external_device_conn`, backend API requires a long vpc_id (vpc name + project name) for GCP transit gateways. However, vpc_id for GCP in `aviatrix_vpc` only contains vpc name. If `aviatrix_vpc` is modified, many other resources must be modified, too. So only `aviatrix_transit_external_device_conn` was modified in this PR.

Here's how to use it (**only for GCP**):

- When creating the resource, user can either reference vpc_id in `aviatrix_vpc` resource or use a "short" vpc_id (vpc name only). But the resource id will be "vpc name~long vpc_id (vpc name + project name)", this is for the purpose of import.
- When importing the resource, user has to use the long vpc_id.

**Please let me know what you think about this change.**

Other fixes:

- Moved the checks for AD/FD for OCI to before the HA checks.
- Fixed the bug where ADs/FDs cause diff for non OCI vpc.

### DIFF
--- a/aviatrix/resource_aviatrix_transit_external_device_conn.go
+++ b/aviatrix/resource_aviatrix_transit_external_device_conn.go
@@ -349,7 +349,6 @@ func resourceAviatrixTransitExternalDeviceConnCreate(d *schema.ResourceData, met
 	client := meta.(*goaviatrix.Client)
 
 	externalDeviceConn := &goaviatrix.ExternalDeviceConn{
-		VpcID:                  d.Get("vpc_id").(string),
 		ConnectionName:         d.Get("connection_name").(string),
 		GwName:                 d.Get("gw_name").(string),
 		ConnectionType:         d.Get("connection_type").(string),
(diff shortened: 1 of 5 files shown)


aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -678,7 +700,11 @@ func resourceAviatrixTransitExternalDeviceConnRead(d *schema.ResourceData, meta
 	}
 
 	if conn != nil {
-		d.Set("vpc_id", conn.VpcID)
+		if strings.Contains(conn.VpcID, "~-~") {
+			d.Set("vpc_id", strings.Split(conn.VpcID, "~-~")[0])
+		} else {
+			d.Set("vpc_id", conn.VpcID)
+		}
```
CyrusJavan, line 707: Won't this cause a diff for current GCP transit_external_device_conn resources?

I'm not sure if this is the best solution, maybe we should consult with backend and see why does transit_external_device_conn vpc_id require the GCP project_name and see if we can make it more uniform across resources.

PR #1055, opened 2021-07-13
22246: add support of key file for Check Point

shuoliugit:


goaviatrix/vendor_integration.go
```diff
@@ -140,3 +141,29 @@ func (c *Client) SyncFireNetFirewallManagerVendorConfig(ctx context.Context, fir
...
+
+	var files []File
+
+	key := File{
+		ParamName:      "private_key_file",
+		UseFileContent: true,
+		FileName:       "key.pem", // fake name for key
+		FileContent:    vendorInfo.PrivateKeyFile,
+	}
+	files = append(files, key)
```
CyrusJavan: Why do you set the 'private_key_file' parameter twice? Once as text and again as file

shuoliugit: good catch

zhiqiangjin (approved):

PR #1058, opened 2021-07-16
AVX11294: Terraform support for spot instances

zhiqiangjin:
To minimize the impact on customers, two optional attributes are added. They are required with each other. No default values.

### DIFF
--- a/aviatrix/resource_aviatrix_gateway.go
+++ b/aviatrix/resource_aviatrix_gateway.go
@@ -412,6 +412,18 @@ func resourceAviatrixGateway() *schema.Resource {
 				ForceNew:    true,
 				Description: "Name of storage account with gateway images. Only valid for Azure China (2048)",
 			},
+			"enable_spot_instance": {
+				Type:         schema.TypeBool,
+				Optional:     true,
+				Description:  "Enable spot instance. NOT supported for production deployment.",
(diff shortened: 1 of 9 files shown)


aviatrix/resource_aviatrix_gateway.go
```diff
@@ -412,6 +412,18 @@ func resourceAviatrixGateway() *schema.Resource {
...
+				Optional:     true,
+				Description:  "Enable spot instance. NOT supported for production deployment.",
+				RequiredWith: []string{"spot_price"},
+			},
+			"spot_price": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "Price for spot instance. NOT supported for production deployment.",
+				RequiredWith: []string{"enable_spot_instance"},
+			},
```
shuoliugit, line 426: If I set `enable_spot_instance` to false without setting `spot_price`, will this trigger an error?

zhiqiangjin: Yes. It is as expected. If customer is not using it, they should not set it to any value (whether true or false), because it is optional only and does not have a default value (it will not show in state file).

Customers should not be impacted at all if they are not using it. And they are not able to use it in production deployment.

shuoliugit (approved):

PR #1059, opened 2021-07-16
AVX-9874: Support preallocated EIP for gateway, spoke gateway and tra…

mlin-aviatrix:
…nsit gateway in Azure and OCI


aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1911,6 +1948,16 @@ func resourceAviatrixGatewayUpdate(d *schema.ResourceData, meta interface{}) err
...
+		// peering_ha_eip can not be changed to empty string because it is computed. ALso check peering_ha_gw_size to detect when Peering HA gateway is being deleted.
+		if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.AzureArmRelatedCloudTypes) && gw.Eip != "" && gw.GwSize != "" {
+			// AVX-9874 Azure EIP has a different format e.g. 'test_ip::104.45.186.20'
+			haAzureEipName, ok := d.GetOk("peering_ha_azure_eip_name")
+			if !ok {
+				return fmt.Errorf("failed to create HA Gateway: 'peering_ha_azure_eip_name' must be set when a custom EIP is provided and cloud_type is Azure (8), AzureGov (32) or AzureChina (2048)")
+			}
+			gw.Eip = fmt.Sprintf("%s::%s", haAzureEipName.(string), gw.Eip)
+		}
+
```
shuoliugit, line 2019: If I only remove `peering_ha_azure_eip_name`, seems no error will be returned, since this check will not be reached.

mlin-aviatrix: This is for custom EIP for HA gateway. EIP (the IP address) and azure EIP name (name of the azure public IP address resource) are both required for custom EIP in Azure.  gwSize is required for HA gateway. If EIP is left empty, azure EIP name does not matter, but I can add negative check for that. If EIP is not empty, then `peering_ha_azure_eip_name` must not be empty and this check will catch that.

shuoliugit (dismissed):

zhiqiangjin (approved):

PR #1069, opened 2021-07-27
AVX-13664: Async controller upgrade

CyrusJavan:
To avoid controller upgrade APIs failing to get a response from the
controller, we will use an asynchronous method of calling the upgrade
API. This has 2 steps:
 1) Call upgrade API with caller=ui parameter. This will cause the API
 to immediately return a request ID.

 2) Use the request ID from step 1 to call the check_upgrade_status API
 to check if the upgrade has completed yet. We will poll every 30 seconds
 for up to 60 iterations (around 30 minutes).

### DIFF
--- a/goaviatrix/version.go
+++ b/goaviatrix/version.go
@@ -135,6 +135,7 @@ func (c *Client) AsyncUpgrade(version *Version, upgradeGateways bool) error {
 	if i == maxPoll {
 		return fmt.Errorf("waited %s but upgrade never finished. Please manually verify the upgrade status", maxPoll*sleepDuration)
 	}
+	c.Login()
 	return nil
 }
 
--- a/goaviatrix/version.go
+++ b/goaviatrix/version.go
@@ -56,57 +56,84 @@ func (av *AviatrixVersion) String(includeBuild bool) string {
 		return fmt.Sprintf("%d.%d.%d", av.Major, av.Minor, av.Build)
 	}
 	return fmt.Sprintf("%d.%d", av.Major, av.Minor)
-
 }
 
-func (c *Client) Upgrade(version *Version, upgradeGateways bool) error {
-	if version.Version == "" {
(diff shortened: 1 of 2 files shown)


goaviatrix/version.go
```diff
@@ -56,58 +56,85 @@ func (av *AviatrixVersion) String(includeBuild bool) string {
...
+		if !data.Done {
+			// Not done yet
+			time.Sleep(sleepDuration)
+			continue
+		}
+		// Upgrade is done, check for error
+		if strings.HasPrefix(data.Result, "Error") {
+			return fmt.Errorf("post check_upgrade_status failed: %s", data.Result)
 		}
-		c.Login()
```
arvindsree, line 108: If CID is invalidated due to some reason, we might need to login again

CyrusJavan: Good point. Added it back.

goaviatrix/version.go
```diff
@@ -56,58 +56,86 @@ func (av *AviatrixVersion) String(includeBuild bool) string {
...
+	// Use the requestID to poll until upgrade is finished
+	form = map[string]string{
+		"action": "check_upgrade_status",
+		"CID":    c.CID,
+		"id":     strconv.Itoa(requestID),
+		"pos":    "0",
+	}
+	backendURL := fmt.Sprintf("https://%s/v1/backend1", c.ControllerIP)
+	const maxPoll = 60
+	sleepDuration := time.Second * 30
```
arvindsree: Maybe sleep duration of 10 seconds and maxpoll of 180 would be slightly more responsive

arvindsree (dismissed):

goaviatrix/version.go
CyrusJavan: Good idea, updated.

zhiqiangjin (approved):

PR #1072, opened 2021-07-28
AVX-13823: Parallel Primary/HA pair upgrade

CyrusJavan:
In the scenario where a user is trying to upgrade
both primary and ha software_version, we can do the
upgrade in parallel.

If user is trying to upgrade image_version as well,
then we cannot safely do parallel upgrade. So, we fall
back to doing it serially.

### DIFF
--- a/aviatrix/resource_aviatrix_gateway.go
+++ b/aviatrix/resource_aviatrix_gateway.go
@@ -2533,7 +2533,7 @@ func resourceAviatrixGatewayUpdate(d *schema.ResourceData, meta interface{}) err
 			}()
 			wg.Wait()
 			if primaryErr != nil && haErr != nil {
-				return fmt.Errorf("could not upgrade either primary or HA gateway "+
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_gateway.go
```diff
@@ -2501,31 +2502,71 @@ func resourceAviatrixGatewayUpdate(d *schema.ResourceData, meta interface{}) err
...
+				primaryErr = client.UpgradeGateway(gw)
+				wg.Done()
+			}()
+			go func() {
+				haErr = client.UpgradeGateway(hagw)
+				wg.Done()
+			}()
+			wg.Wait()
+			if primaryErr != nil && haErr != nil {
+				return fmt.Errorf("could not upgrade either primary or HA gateway "+
```
arvindsree: and* ?

CyrusJavan: Changed to 'could not upgrade primary and HA..."

arvindsree (approved):

PR #1077, opened 2021-07-30
AVX-9873: Add DiffSuppressFunc for old Azure vpc_id in gateway, spoke gateway and transit gateway

mlin-aviatrix:


aviatrix/utils.go
```diff
@@ -197,3 +197,21 @@ func validateAzureEipNameResourceGroup(i interface{}, k string) (warnings []stri
...
+	// If gateway vpc_id is in the new format (3 tuple) e.g. name:rg_name:guid
+	// and the vpc_id provided in the terraform file is in the old format (2 tuple)
+	// e.g. name:rg_name only compare the first two parts and ignore the guid
+	oldValue := strings.Split(old, ":")
+	newValue := strings.Split(new, ":")
+	if len(oldValue) == 3 && len(newValue) == 2 {
+		return oldValue[0] == newValue[0] && oldValue[1] == newValue[1]
+	}
+
+	return false
```
zhiqiangjin, line 216: If DiffSuppressFunc is non-nil, the provided function will be used during diff of this field. So, inside this DiffSuppressFuncGatewayVpcId, not only need to check the special case, but also need to check the normal case, which is that both are 3 tuple, right?

mlin-aviatrix: If they are both 3 tuple, DiffSuppressFunc will return false and Terraform will do its normal difference comparison. It will work as normal.

zhiqiangjin (approved):

PR #1088, opened 2021-08-10
AVX-14208: Ignore 'configuration not changed' error from edit_firenet API

CyrusJavan:
Another API call that got messed up in the GetAPI/PostAPI refactor.

### DIFF
--- a/goaviatrix/firenet.go
+++ b/goaviatrix/firenet.go
@@ -212,9 +212,6 @@ func (c *Client) EditFireNetInspection(fireNet *FireNet) error {
 
 	checkFunc := func(act, method, reason string, ret bool) error {
 		if !ret {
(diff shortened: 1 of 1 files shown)


goaviatrix/firenet.go
```diff
@@ -215,6 +215,9 @@ func (c *Client) EditFireNetInspection(fireNet *FireNet) error {
 			if strings.Contains(reason, "configuration not found") {
 				return nil
 			}
```
shuoliugit: I think we can remove this part.

```
if strings.Contains(reason, "configuration not found") {
	return nil
}
```

CyrusJavan: Done

zhiqiangjin (approved):

PR #1097, opened 2021-08-20
AVX-14149: Handle ErrNotFound properly

CyrusJavan:
To make sure the equality check `err == ErrNotFound` works correctly, we need to make sure we return just ErrNotFound, not ErrNotFound wrapped in another error.


zhiqiangjin:
Let's check how many resources are facing the same issue first.

CyrusJavan:
@zhiqiangjin I've went through all the resources and data sources, I found a few other places that were not handling ErrNotFound correctly.

zhiqiangjin (approved):

PR #1099, opened 2021-08-25
AVX-14601: Update documentation for gateway, spoke gateway and transi…

mlin-aviatrix:
…t gateway vpc_id error


shuoliugit (dismissed):

inc1t3Ful (changes requested):
Technically, this warning isn't accurate, right? Because gateway itself will not throw an error in the apply. It only throws an error if there are dependent resources after that gateway. 

The error gets thrown because the downline resources that depend on the `vpc_id` of the gateway sees that the values have changed after the first apply.

Also, I think we should keep it consistent and be clear with naming conventions. For Azure, its a VNet.

inc1t3Ful (approved):

zhiqiangjin (approved):

PR #1125, opened 2021-10-13
AVX-15904: Move controller security group management config out of aviatrix_controller_config

zhiqiangjin:
1. If customer uses aviatrix_controller_config but did not config security group management in it, after upgrading to provider 2.20.1, they do not need to do anything (a terraform refresh will remove "security_group_management" from state file). They will need to use aviatrix_controller_security_group_management_config to enable/disable security group management; 
2. If customer uses aviatrix_controller_config and did config security group management in it, after upgrading to provider 2.20.1, they need to remove "sg_management_account_name" and "security_group_management" from their configuration file (a terraform refresh will remove "sg_management_account_name" and "security_group_management" from state file). They will need to use aviatrix_controller_security_group_management_config to enable/disable security group management;


shuoliugit (dismissed):

mlin-aviatrix (dismissed):

docs/resources/aviatrix_controller_security_group_management_config.md
```diff
@@ -0,0 +1,44 @@
...
+# aviatrix_controller_security_group_management_config
+
+The **aviatrix_controller_security_group_management_config** resource allows management of an Aviatrix Controller's security group management configurations.
+
+## Example Usage
+
+```hcl
+# Create an Aviatrix Controller Security Group Management Config to Enable Security Group Management
+resource "aviatrix_controller_security_group_management_config" "test_sqm_config" {
+  account_name                     = true
```
CyrusJavan: Should be string not bool

docs/resources/aviatrix_controller_security_group_management_config.md
```diff
@@ -0,0 +1,44 @@
...
+subcategory: "Settings"
+layout: "aviatrix"
+page_title: "Aviatrix: aviatrix_controller_security_group_management_config"
+description: |-
+  Creates and manages an Aviatrix controller security group management config resource
+---
+
+# aviatrix_controller_security_group_management_config
+
+The **aviatrix_controller_security_group_management_config** resource allows management of an Aviatrix Controller's security group management configurations.
```
CyrusJavan: Could add a sentence "This resource is available as of v2.20.1"

docs/resources/aviatrix_controller_config.md
```diff
@@ -113,6 +105,13 @@ In addition to all arguments above, the following attributes are exported:
 * `previous_version` - Previous version of the controller including the build number. Example: "6.5.123". Available as of provider version R2.20.0+.
 * `current_version` - Current version of the controller including the build number. Example: "6.5.123". Available as of provider version R2.20.0+.
 
+The following attributes are deprecated and removed. Please use **aviatrix_controller_security_group_management_config** resource to manage controller's security group management settings.
+
+
+* `sg_management_account_name` - (Optional) Select the [primary access account](https://docs.aviatrix.com/HowTos/aviatrix_account.html#setup-primary-access-account-for-aws-cloud).
+* `security_group_management` - (Optional) Enable to allow Controller to automatically manage inbound rules from gateways. Valid values: true, false. Default value: false.
```
CyrusJavan: I'm not sure if we should still list the attributes here, since we removed them they really should not be listed like other attributes. Maybe instead just put it all in a yellow box like this 
```
~> **NOTE:** The following attributes are deprecated and removed. Please use **aviatrix_controller_security_group_management_config** resource to manage controller's security group management settings.
`sg_management_account_name` - (Optional) Select the [primary access account](https://docs.aviatrix.com/HowTos/aviatrix_account.html#setup-primary-access-account-for-aws-cloud).
`security_group_management` - (Optional) Enable to allow Controller to automatically manage inbound rules from gateways. Valid values: true, false. Default value: false.
```

You can use the doc preview tool to make sure it will look correct https://registry.terraform.io/tools/doc-preview

aviatrix/resource_aviatrix_controller_security_group_management_config.go
```diff
@@ -0,0 +1,135 @@
...
+	sgm, err := client.GetSecurityGroupManagementStatus()
+	if err != nil {
+		return fmt.Errorf("could not read Aviatrix Controller Security Group Management Status: %s", err)
+	}
+	if sgm != nil {
+		if sgm.State == "Enabled" {
+			d.Set("enable_security_group_management", true)
+		} else {
+			d.Set("enable_security_group_management", false)
+		}
```
CyrusJavan: Simplify to `d.Set("enable_security_group_management", sgm.State == "Enabled")`

aviatrix/resource_aviatrix_controller_security_group_management_config.go
```diff
@@ -0,0 +1,135 @@
...
+	d.SetId(strings.Replace(client.ControllerIP, ".", "-", -1))
+	return nil
+}
+
+func resourceAviatrixControllerSecurityGroupManagementConfigUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*goaviatrix.Client)
+
+	account := d.Get("account_name").(string)
+
+	if d.HasChange("enable_security_group_management") {
```
CyrusJavan: What about the case where user changes account_name but not enable_security_group_management? Shouldn't we disable and reenable with the new account_name?

CyrusJavan (changes requested):

CyrusJavan (approved):

PR #1127, opened 2021-10-14
AVX-15939: Fix issue where it returns error if previous version is empty in aviatrix_controller_config

zhiqiangjin:


goaviatrix/version.go
```diff
@@ -258,7 +258,8 @@ func (c *Client) GetLatestVersion() (string, error) {
 func ParseVersion(version string) (string, *AviatrixVersion, error) {
 	version = strings.TrimPrefix(version, "UserConnect-")
 	if version == "" {
-		return "", nil, errors.New("unable to parse version information since it is empty")
+		log.Infof("version is an empty string")
+		return "", nil, nil
```
CyrusJavan: As it is now this will panic when the caller tries to use the AviatrixVersion object. I think if you return an empty object instead of nil, then this will be okay.
```suggestion
		return "", &AviatrixVersion{}, nil
```

CyrusJavan (changes requested):

CyrusJavan (approved):

PR #1130, opened 2021-10-15
Updated documentation to reflect default controller values

t-dever:
Updated aviatrix_remote_syslog and aviatrix_netflow_agent to reflect default values for the Controller.


shuoliugit:
@t-dever The default values are only updated in doc. Do you also want to update the "real" default values of the parameters?

t-dever:
@shuoliugit Yes let me work on updating the real default values.

PR #1134, opened 2021-11-02
AVX-16085: Add support of scale s2c netmap cidrs

zhiqiangjin:
1. Update API add_site2cloud from Get to Post;
2. Keep TypeList for all eight attributes;
3. Add DiffSuppressFunc to ignore the order difference of the elements for above attributes.


goaviatrix/site2cloud.go
```diff
@@ -583,3 +559,74 @@ func S2CPh1RemoteIdDiffSuppressFunc(k, old, new string, d *schema.ResourceData)
 
 	return false
 }
+
+func DiffSuppressFuncRemoteSourceRealCIDRs(k, old, new string, d *schema.ResourceData) bool {
```
mlin-aviatrix, line 563: We should only need 1 DiffSuppressFunc because they are all doing the same thing. `k` is the name of the key used in d.GetChange, which can look like `o, n := d.GetChange(k)`.

mlin-aviatrix (changes requested):

goaviatrix/site2cloud.go
zhiqiangjin: I think it will panic for following reason:
panic: interface conversion: interface {} is string, not []interface {}

k is a string, however, the value of k is a list of string instead of string.

shuoliugit: If k is a string, then d.GetChange(k) should return what o, n := d.GetChange("remote_source_real_cidrs") returns, right?

@mlin-aviatrix I've never used this "k" before, did you use it before?

zhiqiangjin: DiffSuppressFunc has its fixed format, where k, old and new are all string. k is the key, old and new are values before and after change.

mlin-aviatrix: After some testing, it looks like using DiffSuppressFunc on a list type causes it to run the DiffSuppressFunc for every element on the list, which won't work for this implementation because order is irrelevant. There is an open issue on Hashicorp for this: https://github.com/hashicorp/terraform-plugin-sdk/issues/477

mlin-aviatrix (approved):

shuoliugit (approved):

PR #1135, opened 2021-11-04
AVX-16378: check active-active for transit ha enabled unmapped policy based conn

shuoliugit:


aviatrix/resource_aviatrix_site2cloud.go
```diff
@@ -439,16 +439,32 @@ func resourceAviatrixSite2CloudCreate(d *schema.ResourceData, meta interface{})
...
 
+	if s2c.ConnType != "mapped" && s2c.ConnType != "unmapped" {
+		return fmt.Errorf("'connection_type' should be 'mapped' or 'unmapped'")
+	}
+
+	gateway := &goaviatrix.Gateway{
+		GwName: s2c.GwName,
+	}
+
+	gw, err := client.GetGateway(gateway)
```
zhiqiangjin, line 450: We should avoid calling more APIs here.

aviatrix/resource_aviatrix_site2cloud.go
```diff
@@ -439,16 +439,32 @@ func resourceAviatrixSite2CloudCreate(d *schema.ResourceData, meta interface{})
...
+	if err != nil {
+		return fmt.Errorf("couldn't find Aviatrix Gateway %s: %v", s2c.GwName, err)
+	}
+
+	if gw.IsTransit == "yes" {
```
zhiqiangjin: If IsTransit has to be added, it should be a boolean instead of a string.

zhiqiangjin (changes requested):

aviatrix/resource_aviatrix_site2cloud.go
shuoliugit: this is a special case only for transit gateway, any suggestions about how to get the type of a gateway?

aviatrix/resource_aviatrix_site2cloud.go
shuoliugit: changed IsTransit to TransitVpc

aviatrix/resource_aviatrix_site2cloud.go
shuoliugit: have to call an API here to get the gateway type

aviatrix/resource_aviatrix_site2cloud.go
```diff
@@ -439,16 +439,32 @@ func resourceAviatrixSite2CloudCreate(d *schema.ResourceData, meta interface{})
...
+		return fmt.Errorf("'connection_type' should be 'mapped' or 'unmapped'")
+	}
+
+	gateway := &goaviatrix.Gateway{
+		GwName: s2c.GwName,
+	}
+
+	gw, err := client.GetGateway(gateway)
+	if err != nil {
+		return fmt.Errorf("couldn't find Aviatrix Gateway %s: %v", s2c.GwName, err)
```
zhiqiangjin: This error message will be no good if err == ErrNotFound

zhiqiangjin (changes requested):

mlin-aviatrix (approved):

zhiqiangjin (approved):

PR #1136, opened 2021-11-04
AVX-15157: Remove support of non-ActiveMesh feature

zhiqiangjin:


mlin-aviatrix (dismissed):

shuoliugit (dismissed):

inc1t3Ful:
Instead of removing all mention of ActiveMesh in the spoke and transit doc, we should just move the attribute (and description) to the bottom of the doc where the 'deprecated' attributes are.

We can have another PR in release notes to be sure to be clear that we are officially removing ActiveMesh due to \< reason we stated in previous release \>

inc1t3Ful (approved):

PR #1140, opened 2021-11-10
AVX-16551: Add enable_learned_cidrs_approval to aviatrix_cloudn_transit_gateway_attachment

mlin-aviatrix:


aviatrix/resource_aviatrix_cloudn_transit_gateway_attachment.go
```diff
@@ -172,16 +215,27 @@ func resourceAviatrixCloudnTransitGatewayAttachmentRead(ctx context.Context, d *
...
 	client := meta.(*goaviatrix.Client)
+	d.Partial(true)
 
 	attachment := marshalCloudnTransitGatewayAttachmentInput(d)
 
+	approvedCidrs := getStringSet(d, "approved_cidrs")
+	enableLearnedCIDRApproval := d.Get("enable_learned_cidrs_approval").(bool)
+	// `approved_cidrs` is Optional/Computed so Terraform will not detect a change even if `approved_cidrs` is removed from Terraform configuration.
+	// Only prevent from setting `enable_learned_cidrs_approval = false` when Terraform detects a change to `approved_cidrs`.
+	if !enableLearnedCIDRApproval && len(approvedCidrs) > 0 && d.HasChange("approved_cidrs") {
```
mlin-aviatrix, line 239: Checking for `d.HasChange("approved_cidrs")` will allow users to set `enable_learned_cidrs_approval = false` when `approved_cidrs` is not empty on the Terraform configuration file. On any subsequent applies, Terraform will give an error because `approved_cidrs` is not empty. I think this is better than leaving it out, because it always prevents the user from setting `enable_learned_cidrs_approval = false`

aviatrix/resource_aviatrix_cloudn_transit_gateway_attachment.go
```diff
@@ -80,6 +80,18 @@ func resourceAviatrixCloudnTransitGatewayAttachment() *schema.Resource {
 				Default:     true,
 				Description: "Enable dead peer detection.",
 			},
+			"enable_learned_cidrs_approval": {
```
zhiqiangjin, line 83: There should be default value for a bool attribute.

aviatrix/resource_aviatrix_cloudn_transit_gateway_attachment.go
```diff
@@ -172,16 +215,27 @@ func resourceAviatrixCloudnTransitGatewayAttachmentRead(ctx context.Context, d *
 	d.Set("enable_over_private_network", attachment.EnableOverPrivateNetwork)
 	d.Set("enable_jumbo_frame", attachment.EnableJumboFrame)
 	d.Set("enable_dead_peer_detection", attachment.EnableDeadPeerDetection)
+	d.Set("enable_learned_cidrs_approval", attachment.EnableLearnedCidrsApproval)
+	d.Set("approved_cidrs", attachment.ApprovedCidrs)
```
zhiqiangjin: approved_cidrs is a type of set, which is not a primitive type. Add a check for setting it.

zhiqiangjin (changes requested):

aviatrix/resource_aviatrix_cloudn_transit_gateway_attachment.go
mlin-aviatrix: Do we need default value for bool when it is false? I thought it is implied that default is false for optional values.

zhiqiangjin (approved):

PR #1142, opened 2021-11-11
AVX-16797: Update async call for update due to API change

zhiqiangjin:


goaviatrix/version.go
```diff
@@ -83,21 +83,24 @@ func (c *Client) AsyncUpgrade(version *Version, upgradeGateways bool) error {
...
-	requestID, err := strconv.Atoi(buf.String())
-	if err != nil {
+	bodyString := buf.String()
+	bodyIoCopy := strings.NewReader(bodyString)
+	if err = json.NewDecoder(bodyIoCopy).Decode(&data); err != nil {
+		return errors.New("Json Decode " + form["action"] + " failed: " + err.Error() + "\n Body: " + bodyString)
+	}
+
+	requestID := data.Result
+	if requestID == 0 {
```
CyrusJavan: Should we check the Return field in the response?

zhiqiangjin: Yes, we should.

CyrusJavan (approved):
looks good

PR #1174, opened 2021-12-14
AVX-16002: Add support of GCP BGP over LAN transit external device connection

zhiqiangjin:


aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -674,11 +674,16 @@ func resourceAviatrixTransitExternalDeviceConnRead(d *schema.ResourceData, meta
 		id := d.Id()
 		log.Printf("[DEBUG] Looks like an import, no 'connection_name' or 'vpc_id' received. Import Id is %s", id)
 		parts := strings.Split(id, "~")
```
mlin-aviatrix: Maybe we can use [strings.SplitN ](https://pkg.go.dev/strings#SplitN) here so that the rest of the code does not need to be changed. We can SplitN id into 2 parts so that the "~" in vpc_id will not be split and no changes need to be made to the rest of the code.
`parts := strings.SplitN(id, "~", 2)`

mlin-aviatrix (dismissed):

mlin-aviatrix (approved):

PR #1176, opened 2021-12-16
AVX-16077: add support for SSL certificate verification

shuoliugit:


aviatrix/config.go
```diff
@@ -22,11 +27,24 @@ type Config struct {
 // Returns:
 //    the aviatrix client (from goaviatrix)
 //    error (if any)
-func (c *Config) Client() (*goaviatrix.Client, error) {
+func (c *Config) Client(config *Config) (*goaviatrix.Client, error) {
```
mlin-aviatrix: We don't need the extra `config *Config` parameter. You can reference the `*Config*` using `c.VerifyCert` and `c.PathToCaCert` because `Client` is a method of `c *Config`

mlin-aviatrix (changes requested):

mlin-aviatrix (approved):

zhiqiangjin (approved):

PR #1177, opened 2021-12-21
AVX-17663: add workaround for EOF error when deleting HA gateway

shuoliugit:


aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -3096,9 +3096,33 @@ func resourceAviatrixTransitGatewayDelete(d *schema.ResourceData, meta interface
...
-		err := client.DeleteGateway(gateway)
-		if err != nil {
-			return fmt.Errorf("failed to delete Aviatrix Transit Gateway HA gateway: %s", err)
+		try, EOFTimes, maxTries, backoff := 0, 0, 2, 500*time.Millisecond
+
+		for {
+			try++
+			err := client.DeleteGateway(gateway)
+			if err != nil {
+				if EOFTimes != 0 && strings.Contains(err.Error(), "not found") {
```
zhiqiangjin: EOFTimes is useless here, right?
The HA gateway could be destroyed from controller, then EOFTimes == 0 and strings.Contains(err.Error(), "not found"). If so, you should just break.

I think you can remove EOFTimes. Make it as simple as possible. Customers do not need to know how many times you try.

zhiqiangjin (changes requested):

aviatrix/resource_aviatrix_transit_gateway.go
shuoliugit: The question is whether we need to distinguish two cases:
1. HA has been destroyed from controller manually -> we'll get "not found" error, and EOFTimes == 0
  - in this case, I think we should return the error to let user know that HA is not found
2. in the 1st try EOF is returned, but backend starts to destroy HA -> in 2nd try we'll get "not found" error, and EOFTimes != 0
  - in this case, we should break the loop, because HA has been destroyed

zhiqiangjin (approved):

PR #1178, opened 2021-12-22
AVX-17738: Add support of BGP LAN ActiveMesh mode for GCP Transit External Device Conn

zhiqiangjin:


aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -564,6 +571,16 @@ func resourceAviatrixTransitExternalDeviceConnCreate(d *schema.ResourceData, met
 		}
 	}
 
+	if val, ok := d.GetOk("enable_bgp_lan_activemesh"); ok && val.(bool) {
```
mlin-aviatrix: `enable_bgp_lan_active_mesh` is TypeBool with default false so we shouldn't need to check with GetOK. We can just check Get: `d.Get("enable_bgp_lan_activemesh").(bool)`

mlin-aviatrix (changes requested):

mlin-aviatrix (approved):

shuoliugit (approved):

PR #1198, opened 2022-01-20
AVX-18579: Revert changes to remote_vpc_name documentation in transit…

mlin-aviatrix:
…_external_device_conn


mlin-aviatrix:
remote_vpc_name has a different format that the vpc_id. It uses subscription id in the third part instead of guid.

zhiqiangjin (approved):

PR #1249, opened 2022-04-07
AVX-21060 Sort lists in aviatrix_firewall_instance_images data source

MingweiZhu666:
Sort lists in aviatrix_firewall_instance_images data source
fix:
sort the firewall_image_version list form latest to old
can sort version format example:
 firewall_image_version = [
                "10.2.0",
                "10.0.9",
                "10.0.8-h8",
                "10.0.7",
            ]
 firewall_image_version = [
                "9.1.2",
                "9.1.0-h3",
                "9.0.9-h1.xfr",
                "9.0.9.xfr",
            ]
 firewall_image_version = [
                "(7.0.5)",
                "(7.0.3)",
                "(6.4.8)",
                "(6.4.7)",
            ]
  firewall_image_version = [
                "R81.10-335.986",
                "R81-392.983",
                "R80.40-294.983",
            ]
  firewall_image_version = [
                "R81.10_rev1.1",
                "R81.10_rev1.0",
                "R81_rev1.2",
            ]
  firewall_image_version = [
                "PA-VM-10.2.0",
                "PA-VM-10.1.0",
                "PA-VM-10.0.9",
            ]
sort the firewall_size list

### DIFF
--- a/aviatrix/data_source_aviatrix_firewall_instance_images.go
+++ b/aviatrix/data_source_aviatrix_firewall_instance_images.go
@@ -66,7 +66,7 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 		sort.Slice(versionList, func(i, j int) bool {
 			return sortVersion(versionList, i, j)
 		})
-		fI["firewall_image_version"] = reverseArray(versionList)
+		fI["firewall_image_version"] = versionList
 		sizeList := image.Size
 		sort.Slice(sizeList, func(i, j int) bool {
 			return sortSize(sizeList, i, j)
--- a/aviatrix/data_source_aviatrix_firewall_instance_images.go
+++ b/aviatrix/data_source_aviatrix_firewall_instance_images.go
@@ -2,12 +2,13 @@ package aviatrix
 
 import (
 	"fmt"
-	"github.com/hashicorp/go-version"
 	"log"
 	"regexp"
 	"sort"
 	"strconv"
 	"strings"
+
+	"github.com/hashicorp/go-version"
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v2/goaviatrix"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
--- a/aviatrix/data_source_aviatrix_firewall_instance_images.go
+++ b/aviatrix/data_source_aviatrix_firewall_instance_images.go
@@ -2,6 +2,9 @@ package aviatrix
 
 import (
 	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v2/goaviatrix"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -61,8 +64,12 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 	for _, image := range *firewallInstanceImages {
 		fI := make(map[string]interface{})
 		fI["firewall_image"] = image.Image
-		fI["firewall_image_version"] = image.Version
-		fI["firewall_size"] = image.Size
+		listVersion := image.Version
+		sortImageVersion(listVersion)
+		fI["firewall_image_version"] = listVersion
+		listSize := image.Size
+		sortImageSize(listSize)
+		fI["firewall_size"] = listSize
 		images = append(images, fI)
 	}
 
@@ -72,4 +79,141 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 
 	d.SetId(vpcId)
 	return nil
+}
+
+func sortImageVersion(listVersion []string) []string {
+	var n = len(listVersion)
+	for i := 0; i <= n-1; i++ {
+		for j := i; j <= n-1; j++ {
+			var splitFlag string
+			if checkFirstCharacter(listVersion[i]) == "R" {
+				if strings.Contains(listVersion[i], "_") {
+					//format: Rab_abcxx.x
+					splitFlag = "_"
+					if compareVersion2(listVersion[i], listVersion[j], splitFlag) == 2 {
+						listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
+					}
+				} else {
+					//format: Rab-xxx.xxx
+					splitFlag = "-"
+					if compareVersion2(listVersion[i], listVersion[j], splitFlag) == 2 {
+						listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
+					}
+				}
+			} else if checkFirstCharacter(listVersion[i]) == "P" {
+				//format: Pa-bc-xx.xx.xx
+				splitFlag = "-"
+				if compareVersion3(listVersion[i], listVersion[j], splitFlag) == 2 {
+					listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
+				}
+			} else {
+				//format: xx.xxx.xxx
+				splitFlag = "."
+				if compareVersion(listVersion[i], listVersion[j], splitFlag) == 2 {
+					listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
+				}
+			}
+		}
+	}
+	return listVersion
+}
+
+func checkFirstCharacter(input string) string {
+	firstCharacter := input[0:1]
+	return firstCharacter
+}
+
+//format: xx.xx.xx
+func compareVersion(version1, version2, splitFlag string) (res int) {
+	imageVersionArray1 := strings.Split(version1, splitFlag)
+	imageVersionArray2 := strings.Split(version2, splitFlag)
+	for index := range imageVersionArray1 {
+		reg, _ := regexp.Compile("[^0-9]+")
+		v1SliceString := reg.ReplaceAllString(imageVersionArray1[index], ".")
+		v2SliceString := reg.ReplaceAllString(imageVersionArray2[index], ".")
+		int1, _ := strconv.ParseFloat(v1SliceString, 32)
+		int2, _ := strconv.ParseFloat(v2SliceString, 32)
+		if int1 > int2 {
+			return 1
+		}
+		if int1 < int2 {
+			return 2
+		}
+	}
+	return 1
+}
+
+//format: Rab-xxx.xxx and Rab_abcxx.x
+func compareVersion2(version1, version2, flag string) (res int) {
+	imageVersionArray1 := strings.Split(version1, flag)
+	imageVersionArray2 := strings.Split(version2, flag)
+	for index := range imageVersionArray1 {
+		reg, _ := regexp.Compile("[^0-9.]+")
+		v1SliceString := reg.ReplaceAllString(imageVersionArray1[index], "")
+		v2SliceString := reg.ReplaceAllString(imageVersionArray2[index], "")
+		int1, _ := strconv.ParseFloat(v1SliceString, 32)
+		int2, _ := strconv.ParseFloat(v2SliceString, 32)
+		if int1 > int2 {
+			return 1
+		}
+		if int1 < int2 {
+			return 2
+		}
+	}
+	return 1
+}
+
+//format: Pa-bc-xx.xx.xx
+func compareVersion3(version1, version2, flag string) (res int) {
+	imageVersionArray1 := strings.Split(version1, flag)
+	imageVersionArray2 := strings.Split(version2, flag)
+	return compareVersion(imageVersionArray1[2], imageVersionArray2[2], ".")
+}
+
+func sortImageSize(sizeList []string) []string {
+	var n = len(sizeList)
+	for i := 0; i <= n-1; i++ {
+		for j := i; j <= n-1; j++ {
+			if strings.Contains(sizeList[i], "-") {
+				if compareImageSize(sizeList[i], sizeList[j], "-", 2) == 1 {
+					sizeList[i], sizeList[j] = sizeList[j], sizeList[i]
+				}
+			} else if strings.Contains(sizeList[i], ".") {
+				if compareImageSize(sizeList[i], sizeList[j], ".", 1) == 1 {
+					sizeList[i], sizeList[j] = sizeList[j], sizeList[i]
+				}
+			} else if strings.Contains(sizeList[i], "_") {
+				if compareImageSize(sizeList[i], sizeList[j], "_", 1) == 1 {
+					sizeList[i], sizeList[j] = sizeList[j], sizeList[i]
+				}
+			}
+		}
+	}
+	return sizeList
+}
+func compareImageSize(imageSize1, imageSize2, flag string, indexFlag int) int {
+	imageSizeArray1 := strings.Split(imageSize1, flag)
+	imageSizeArray2 := strings.Split(imageSize2, flag)
+	for index := range imageSizeArray1 {
+		if index >= indexFlag {
+			reg, _ := regexp.Compile("[^0-9]+")
+			imageSizeIndex1 := reg.ReplaceAllString(imageSizeArray1[index], "")
+			imageSizeIndex2 := reg.ReplaceAllString(imageSizeArray2[index], "")
+			int1, _ := strconv.Atoi(imageSizeIndex1)
+			int2, _ := strconv.Atoi(imageSizeIndex2)
+			if int1 > int2 {
+				return 1
+			}
+			if int1 < int2 {
+				return 2
+			}
+		}
+		if imageSizeArray1[index] > imageSizeArray2[index] {
+			return 1
+		}
+		if imageSizeArray1[index] < imageSizeArray2[index] {
+			return 2
+		}
+	}
+	return 1
 }
--- a/aviatrix/data_source_aviatrix_firewall_instance_images.go
+++ b/aviatrix/data_source_aviatrix_firewall_instance_images.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -66,10 +67,10 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 		fI := make(map[string]interface{})
 		fI["firewall_image"] = image.Image
 		listVersion := image.Version
-		sortImageVersion(listVersion)
+		sort.Sort(byVersion(listVersion))
 		fI["firewall_image_version"] = listVersion
 		listSize := image.Size
-		sortImageSize(listSize)
+		sort.Sort(sort.Reverse(bySize(listSize)))
 		fI["firewall_size"] = listSize
 		images = append(images, fI)
 	}
@@ -82,52 +83,33 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 	return nil
 }
 
-func sortImageVersion(listVersion []string) []string {
-	var n = len(listVersion)
-	for i := 0; i <= n-1; i++ {
-		for j := i; j <= n-1; j++ {
-			var splitFlag string
-			if checkFirstCharacter(listVersion[i]) == "R" {
-				if strings.Contains(listVersion[i], "_") {
-					//format: Rab_abcxx.x
-					splitFlag = "_"
-					if compareVersion2(listVersion[i], listVersion[j], splitFlag) == 2 {
-						listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
-					}
-				} else if strings.Contains(listVersion[i], "-") {
-					//format: Rab-xxx.xxx
-					splitFlag = "-"
-					if compareVersion2(listVersion[i], listVersion[j], splitFlag) == 2 {
-						listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
-					}
-				} else {
-					log.Printf("need to add a new method sort this version format")
-				}
-			} else if checkFirstCharacter(listVersion[i]) == "P" {
-				//format: Pa-bc-xx.xx.xx
-				splitFlag = "-"
-				if compareVersion3(listVersion[i], listVersion[j], splitFlag) == 2 {
-					listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
-				}
-			} else {
-				//format: xx.xxx.xxx
-				splitFlag = "."
-				if compareVersion(listVersion[i], listVersion[j], splitFlag) == 2 {
-					listVersion[i], listVersion[j] = listVersion[j], listVersion[i]
-				}
-			}
-		}
-	}
-	return listVersion
-}
+type byVersion []string
 
-func checkFirstCharacter(input string) string {
-	firstCharacter := input[0:1]
-	return firstCharacter
+func (p byVersion) Len() int {
+	return len(p)
+}
+func (p byVersion) Swap(i, j int) {
+	p[i], p[j] = p[j], p[i]
+}
+func (p byVersion) Less(i, j int) bool {
+	if checkFirstCharacter(p[i]) == "R" {
+		if strings.Contains(p[i], "_") {
+			return compareVersion2(p[i], p[j], "_")
+		} else if strings.Contains(p[i], "-") {
+			return compareVersion2(p[i], p[j], "-")
+		} else {
+			log.Printf("need to add a new method sort this version format")
+		}
+	} else if checkFirstCharacter(p[i]) == "P" {
+		return compareVersion3(p[i], p[j], "-")
+	} else {
+		return compareVersion(p[i], p[j], ".")
+	}
+	return false
 }
 
 //format: xx.xx.xx
-func compareVersion(version1, version2, splitFlag string) (res int) {
+func compareVersion(version1, version2, splitFlag string) bool {
 	imageVersionArray1 := strings.Split(version1, splitFlag)
 	imageVersionArray2 := strings.Split(version2, splitFlag)
 	for index := range imageVersionArray1 {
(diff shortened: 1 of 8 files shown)


aviatrix/data_source_aviatrix_firewall_instance_images.go
```diff
@@ -73,3 +80,140 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 	d.SetId(vpcId)
 	return nil
 }
+
+func sortImageVersion(listVersion []string) []string {
+	var n = len(listVersion)
+	for i := 0; i <= n-1; i++ {
```
mlin-aviatrix: I think will be better to use the builtin sort.Slice function instead of iterating over the entire list. It will definitely be faster than n^2 nested for loops (not that it really matters as there will always be a small number of elements in the given slice), but I think it is more clear.

You can learn more about sort.Slice at https://pkg.go.dev/sort#Slice

aviatrix/data_source_aviatrix_firewall_instance_images.go
```diff
@@ -73,3 +80,140 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
...
+
+//format: Rab-xxx.xxx and Rab_abcxx.x
+func compareVersion2(version1, version2, flag string) (res int) {
+	imageVersionArray1 := strings.Split(version1, flag)
+	imageVersionArray2 := strings.Split(version2, flag)
+	for index := range imageVersionArray1 {
+		reg, _ := regexp.Compile("[^0-9.]+")
+		v1SliceString := reg.ReplaceAllString(imageVersionArray1[index], "")
+		v2SliceString := reg.ReplaceAllString(imageVersionArray2[index], "")
+		int1, _ := strconv.ParseFloat(v1SliceString, 32)
```
mlin-aviatrix: Is it a good idea to ignore the error here? Can we assume that this will always succeed?

mlin-aviatrix (changes requested):

aviatrix/data_source_aviatrix_firewall_instance_images.go
```diff
@@ -73,3 +80,140 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 	d.SetId(vpcId)
 	return nil
 }
+
+func sortImageVersion(listVersion []string) []string {
+	var n = len(listVersion)
+	for i := 0; i <= n-1; i++ {
+		for j := i; j <= n-1; j++ {
+			var splitFlag string
+			if checkFirstCharacter(listVersion[i]) == "R" {
```
mlin-aviatrix: Instead of deciding how to sort by using the first character in the version name, you should probably use the description. That way if we add anther type of versioning that starts with R, but has a different format it won't break .

mlin-aviatrix (changes requested):

aviatrix/data_source_aviatrix_firewall_instance_images.go
MingweiZhu666: thank you! i will check and fix it.

shuoliugit:
There is a nice solution from Hashicorp - https://github.com/hashicorp/go-version
https://stackoverflow.com/questions/18409373/how-to-compare-two-version-number-strings-in-golang

Need discussion to decide whether to use this or not.

MingweiZhu666:
> There is a nice solution from Hashicorp - https://github.com/hashicorp/go-version https://stackoverflow.com/questions/18409373/how-to-compare-two-version-number-strings-in-golang
> 
> Need discussion to decide whether to use this or not.

i checked the gp-version library, this library only support the Semantic Versioning format, like: 
firewall_image_version = [
                "10.2.0",
                "10.1.4",
                "10.1.3",
                "10.1.1",
                "10.1.0",
                "10.0.8-h8",
                "10.0.9",
                "10.0.7",
                "10.0.6",
                "10.0.5",
            ]
 which works for the sort, but in our case, we have other version formats like:
 firewall_image_version = [
                "R81-392.957",
                "R81-392.710",
                "R81-386.706",
                "R81-373.150",
                "R81-351.146",
                "R81-344.139",
                "R81.10-335.883",
                "R80.40-294.759",
                "R80.40-294.687",
                "R80.40-294.127",
            ]
 and
 firewall_image_version = [
                "R81_rev1.2",
                "R81_rev1.1",
                "R81_rev1.0",
                "R81.10_rev1.1",
                "R81.10_rev1.0",
                "R80.40_rev1.0-042420202345",
                "R80.40_rev1.3",
                "R80.40_rev1.2",
                "R80.40_rev1.1",
            ]
and 
firewall_image_version = [
                "PA-VM-10.1.0",
                "PA-VM-10.0.3-012120210118",
                "PA-VM-10.0.9",
                "PA-VM-10.0.6",
            ]
 these formats i test in my local, the go version is not support to sort

aviatrix/data_source_aviatrix_firewall_instance_images.go
MingweiZhu666: i will add the error log here, thank you

aviatrix/data_source_aviatrix_firewall_instance_images.go
MingweiZhu666: i try to use sort.Slice(versionList,
		func(i, j int) bool {
			return versionList[i] > versionList[j]
		})
which in the Slice method use the _quickSort_func()_:
func Slice(x any, less func(i, j int) bool) {
	rv := reflectValueOf(x)
	swap := reflectSwapper(x)
	length := rv.Len()
	quickSort_func(lessSwap{less, swap}, 0, length, maxDepth(length))
}
it likes a lexicographic sort, so if it sort the version like:
                "10.0.3",
                "10.0.0",
                "9.1.6",
output:
[9.1.6
 10.0.3
 10.0.0]
which is no support sort list.

MingweiZhu666: got your point, have fixed base on the builtin sort.Slice function and custom the less() compare function, thank you!

aviatrix/data_source_aviatrix_firewall_instance_images.go
```diff
@@ -61,8 +68,12 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 	for _, image := range *firewallInstanceImages {
 		fI := make(map[string]interface{})
 		fI["firewall_image"] = image.Image
-		fI["firewall_image_version"] = image.Version
-		fI["firewall_size"] = image.Size
+		listVersion := image.Version
+		sort.Sort(sort.Reverse(byVersion(listVersion)))
```
mlin-aviatrix: You should be able to use sort.Slice instead of sort.Sort. That way, you only need to define 2 new Less functions and you don't have to define new types.

aviatrix/data_source_aviatrix_firewall_instance_images.go
```diff
@@ -73,3 +84,115 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
...
+	} else {
+		return compareVersion(p[i], p[j])
+	}
+	return false
+}
+
+//Rxx.xx-xxx.xxx
+//Rxx.xx_revx.x
+func versionFormat1(version1, version2, flag string) bool {
+	//[Rxx.xx, xxx.xxx]
```
mlin-aviatrix: Not sure what this comment means. Can you be more explicit about what this is trying to say?

aviatrix/data_source_aviatrix_firewall_instance_images.go
```diff
@@ -73,3 +84,115 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
 	d.SetId(vpcId)
 	return nil
 }
+
```
mlin-aviatrix: Everything under here can probably me moved to `aviatrix/utils.go`

mlin-aviatrix (changes requested):

aviatrix/data_source_aviatrix_firewall_instance_images.go
MingweiZhu666: fix:
sort.Slice(versionList, func(i, j int) bool {
 			return sortVersion(versionList, i, j)
 		})
 fI["firewall_image_version"] = reverseArray(versionList)
 sizeList := image.Size
 sort.Slice(sizeList, func(i, j int) bool {
 			return sortSize(sizeList, i, j)
 		})

aviatrix/data_source_aviatrix_firewall_instance_images.go
MingweiZhu666: remove this comment and add new comment to descripe the function

aviatrix/data_source_aviatrix_firewall_instance_images.go
MingweiZhu666: yes, move all sort methods into utils.go file, thank you

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
 
 	return false
 }
+
+func reverseArray(arr []string) []string {
```
mlin-aviatrix: Small nit: Technically should be reverseSlice

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+	for i, j := 0, len(arr)-1; i < j; i, j = i+1, j-1 {
+		arr[i], arr[j] = arr[j], arr[i]
+	}
+	return arr
+}
+
+// sort the firewall_image_version list
```
mlin-aviatrix: Comments about the function should start with the function name. So in this case it might be:
`sortVersion sorts the firewall_image_version list`

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+		}
+	} else if checkFirstCharacter(versionList[i]) == "P" {
+		return versionFormat2(versionList[i], versionList[j], "-")
+	} else {
+		return compareVersion(versionList[i], versionList[j])
+	}
+	return false
+}
+
+// sort the firewall_size list
```
mlin-aviatrix: Same as above.

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+		return compareImageSize(sizeList[i], sizeList[j], "-", 2)
+	} else if strings.Contains(sizeList[i], ".") {
+		return compareImageSize(sizeList[i], sizeList[j], ".", 1)
+	} else if strings.Contains(sizeList[i], "_") {
+		return compareImageSize(sizeList[i], sizeList[j], "_", 1)
+	}
+	return false
+}
+
+// this function can sort the firewall_image_version format like: R81.10-335.883 && R81.10_rev1.0
```
mlin-aviatrix: Same as above.

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+	versionArray1 := strings.Split(version1, flag)
+	versionArray2 := strings.Split(version2, flag)
+	reg, _ := regexp.Compile("[^0-9.-]+")
+	if reg.ReplaceAllString(versionArray1[0], "") == reg.ReplaceAllString(versionArray2[0], "") {
+		return compareVersion(reg.ReplaceAllString(versionArray1[1], ""), reg.ReplaceAllString(versionArray2[1], ""))
+	}
+	return compareVersion(reg.ReplaceAllString(versionArray1[0], ""), reg.ReplaceAllString(versionArray2[0], ""))
+}
+
+// this function can sort the firewall_image_version format like: PA-VM-10.1.0
```
mlin-aviatrix: Same as above.

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+func versionFormat2(version1, version2, flag string) bool {
+	versionArray1 := strings.Split(version1, flag)
+	versionArray2 := strings.Split(version2, flag)
+	return compareVersion(versionArray1[2], versionArray2[2])
+}
+
+// this function can sort the Semantic Version
```
mlin-aviatrix: Same as above.

aviatrix/data_source_aviatrix_firewall_instance_images.go
```diff
@@ -68,12 +62,16 @@ func dataSourceAviatrixFirewallInstanceImagesRead(d *schema.ResourceData, meta i
...
-		sort.Sort(sort.Reverse(byVersion(listVersion)))
-		fI["firewall_image_version"] = listVersion
-		listSize := image.Size
-		sort.Sort(sort.Reverse(bySize(listSize)))
-		fI["firewall_size"] = listSize
+		versionList := image.Version
+		sort.Slice(versionList, func(i, j int) bool {
+			return sortVersion(versionList, i, j)
+		})
+		fI["firewall_image_version"] = reverseArray(versionList)
```
mlin-aviatrix: Is this reverse really needed? I think you can change how the sort works so that we don't need to reverse the list after. Probably swapping true and false in `compareVersion` should work but you will need to test that.

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+
+// sort the firewall_image_version list
+func sortVersion(versionList []string, i, j int) bool {
+	if checkFirstCharacter(versionList[i]) == "R" {
+		if strings.Contains(versionList[i], "_") {
+			return versionFormat1(versionList[i], versionList[j], "_")
+		} else if strings.Contains(versionList[i], "-") {
+			return versionFormat1(versionList[i], versionList[j], "-")
+		} else {
+			log.Printf("need to add a new method sort this version format")
```
mlin-aviatrix: This should also include an example of the version.

aviatrix/utils.go
```diff
@@ -196,3 +200,97 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+func reverseArray(arr []string) []string {
+	for i, j := 0, len(arr)-1; i < j; i, j = i+1, j-1 {
+		arr[i], arr[j] = arr[j], arr[i]
+	}
+	return arr
+}
+
+// sort the firewall_image_version list
+func sortVersion(versionList []string, i, j int) bool {
+	if checkFirstCharacter(versionList[i]) == "R" {
```
mlin-aviatrix: We probably should check the name of the firewall provider instead of jus the first character. That way we won't run into issues if a new format is added.

mlin-aviatrix (changes requested):

aviatrix/utils.go
MingweiZhu666: i check the name of firewall_image, but find this format version: R81.10-335.986 firewall_image name is too same as Semantic Version firewall_image name, it is hard to find a way to distinguish them.

Semantic Version：
Check Point CloudGuard IaaS Single Gateway R80.30 - Bring Your Own License
Check Point CloudGuard IaaS Single Gateway R80.30 - Pay As You Go (NGTP)
Check Point CloudGuard IaaS Single Gateway R80.30 - Pay As You Go (NGTX)
Check Point CloudGuard IaaS Single Gateway R80.40 - Bring Your Own License
Check Point CloudGuard IaaS Single Gateway R80.40 - Pay As You Go (NGTP)
Check Point CloudGuard IaaS Single Gateway R80.40 - Pay As You Go (NGTX)
Check Point CloudGuard IaaS Standalone (gateway + management) R80.40 - Bring Your Own License

R81.10-335.986：
Check Point CloudGuard IaaS Next-Gen Firewall with Threat Prevention
Check Point CloudGuard IaaS All-In-One
Check Point CloudGuard IaaS Firewall & Threat Prevention (Gateway only)
Check Point CloudGuard IaaS Firewall & Threat Prevention (Gateway only) (BYOL)
Check Point CloudGuard IaaS Firewall & Threat Prevention (Standalone)
Check Point CloudGuard IaaS Firewall & Threat Prevention (Standalone) (BYOL)

aviatrix/utils.go
MingweiZhu666: have removed the reverseArray function

aviatrix/utils.go
MingweiZhu666: have fixed, thank you

aviatrix/data_source_aviatrix_firewall_instance_images.go
MingweiZhu666: remove the reverseArray function and set the compare somatic version function from LessThan to GreaterThan

aviatrix/utils.go
```diff
@@ -249,18 +242,18 @@ func versionFormat1(version1, version2, flag string) bool {
...
-// this function can sort the firewall_image_version format like: PA-VM-10.1.0
+// versionFormat2 sorts the firewall_image_version format like: PA-VM-10.1.0
 func versionFormat2(version1, version2, flag string) bool {
 	versionArray1 := strings.Split(version1, flag)
 	versionArray2 := strings.Split(version2, flag)
 	return compareVersion(versionArray1[2], versionArray2[2])
 }
 
-// this function can sort the Semantic Version
+// compareVersion sorts the Semantic Version
```
mlin-aviatrix: Small Nit: Should be `compareVersion compares two Semantic Versions`

aviatrix/utils.go
```diff
@@ -238,7 +231,7 @@ func sortSize(sizeList []string, i, j int) bool {
 	return false
 }
 
-// this function can sort the firewall_image_version format like: R81.10-335.883 && R81.10_rev1.0
+// versionFormat1 sorts the firewall_image_version format like: R81.10-335.883 && R81.10_rev1.0
```
mlin-aviatrix: Should be `versionFormat1 compares...`

aviatrix/utils.go
```diff
@@ -249,18 +242,18 @@ func versionFormat1(version1, version2, flag string) bool {
 	return compareVersion(reg.ReplaceAllString(versionArray1[0], ""), reg.ReplaceAllString(versionArray2[0], ""))
 }
 
-// this function can sort the firewall_image_version format like: PA-VM-10.1.0
+// versionFormat2 sorts the firewall_image_version format like: PA-VM-10.1.0
```
mlin-aviatrix: Should be `versionFormat2 compares...`

aviatrix/utils.go
mlin-aviatrix: Also the function name could be more descriptive. Maybe something like `compareCheckpointVersion`

aviatrix/utils.go
mlin-aviatrix: This one could also be more descriptive. Something like `comparePAVersion`

mlin-aviatrix (changes requested):

aviatrix/utils.go
MingweiZhu666: have fixed base on the image name to choose the sort function

aviatrix/utils.go
MingweiZhu666: have fixed: compareVersion compares two Semantic Versions

aviatrix/utils.go
MingweiZhu666: have fixed: compareCheckpointVersion compares firewall_image_version format like: R81.10-335.883 && R81.10_rev1.0

aviatrix/utils.go
MingweiZhu666: thank you, have fixed: comparePAVersion compares firewall_image_version format like: PA-VM-10.1.0

aviatrix/utils.go
```diff
@@ -198,6 +201,107 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+	version = reg.ReplaceAllString(version, "")
+	pointNumber := strings.Count(version, ".")
+	if pointNumber > 2 {
+		version = removePoint(version)
+	}
+	return version
+}
+
+// removePoint removes all points after the second point in a version
+func removePoint(version string) string {
```
mlin-aviatrix: It doesn't make sense to remove the dot because some comparisons will be ordered incorrectly. For example, `1.0.1.10` > `1.0.10.1` because when the dot is removed it will compare `1.0.110 > 1.0.101`.  The best solution might be to ignore everything after the third dot. We should talk about this.

aviatrix/utils.go
```diff
@@ -198,6 +201,107 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+}
+
+// compareVersion compares two Semantic Versions
+func compareVersion(version1, version2 string) bool {
+	v1, _ := version.NewVersion(version1)
+	v2, _ := version.NewVersion(version2)
+	return v1.GreaterThan(v2)
+}
+
+// checkVersionFormat removes special characters, only keep character, number, point and minus in a version
```
mlin-aviatrix: The Semantic Versioning documentation refers to "." as a dot and "-" as a hyphen so we should probably do the same. Also "character, number" can be replaced with alphanumerics

mlin-aviatrix (changes requested):

aviatrix/utils.go
MingweiZhu666: yes, you are correct, my mistake, i ignore this scenario, before i only tested this example:
                "9.0.9-h1.xfr",
                "9.0.9.xfr",
                "8.1.15",
  
let me think a new way to solve this issue

MingweiZhu666: add a new function: removeAfterThirdDotValue, which removes everything after the third dot in a version

aviatrix/utils.go
MingweiZhu666: fixed as: checkVersionFormat removes special characters, only keep dot, hyphen, alphanumerics in a version

aviatrix/utils.go
```diff
@@ -198,6 +201,115 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+		return compareImageSize(sizeList[i], sizeList[j], "_", 1)
+	}
+	return false
+}
+
+// compareCheckpointVersion compares firewall_image_version format like: R81.10-335.883 && R81.10_rev1.0
+func compareCheckpointVersion(version1, version2, flag string) bool {
+	versionArray1 := strings.Split(version1, flag)
+	versionArray2 := strings.Split(version2, flag)
+	reg, _ := regexp.Compile("[^0-9.-]+")
```
mlin-aviatrix: We can probably define const regexp for each one so that we don't need to compile a new regexp every time. Looks at regexp.MustCompile.

aviatrix/utils.go
```diff
@@ -198,6 +201,115 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+// compareVersion compares two Semantic Versions
+func compareVersion(version1, version2 string) bool {
+	v1, _ := version.NewVersion(version1)
+	v2, _ := version.NewVersion(version2)
+	return v1.GreaterThan(v2)
+}
+
+// checkVersionFormat removes special characters, only keep dot, hyphen, alphanumerics in a version
+func checkVersionFormat(version string) string {
+	reg, _ := regexp.Compile("[^a-zA-Z0-9.-]+")
```
mlin-aviatrix: Same as above.

aviatrix/utils.go
```diff
@@ -198,6 +201,115 @@ func DiffSuppressFuncGatewayVpcId(k, old, new string, d *schema.ResourceData) bo
...
+	}
+	return result
+}
+
+func compareImageSize(imageSize1, imageSize2, flag string, indexFlag int) bool {
+	imageSizeArray1 := strings.Split(imageSize1, flag)
+	imageSizeArray2 := strings.Split(imageSize2, flag)
+	for index := range imageSizeArray1 {
+		if index >= indexFlag {
+			reg, _ := regexp.Compile("[^0-9]+")
```
mlin-aviatrix: Same as above.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/go-multierror v1.1.0 // indirect
 	github.com/hashicorp/go-uuid v1.0.2 // indirect
+	github.com/hashicorp/go-version v1.4.0 // indirect
```
mlin-aviatrix: I think you need to run `go mod tidy`

mlin-aviatrix (changes requested):

aviatrix/utils.go
MingweiZhu666: fix as:
        const DotHyphenNumberRegexp = "[^0-9.-]+"
	reg := regexp.MustCompile(DotHyphenNumberRegexp)

aviatrix/utils.go
MingweiZhu666: fix as:
        const DotHyphenAlphanumericsRegexp = "[^a-zA-Z0-9.-]+"
	reg := regexp.MustCompile(DotHyphenAlphanumericsRegexp)

aviatrix/utils.go
MingweiZhu666: fix as:
        const NumberRegexp = "[^0-9]+"
	reg := regexp.MustCompile(NumberRegexp)
	for index := range imageSizeArray1 {
		if index >= indexFlag {

go.mod
MingweiZhu666: done

mlin-aviatrix (approved):

zhiqiangjin (approved):

PR #1268, opened 2022-04-26
AVX-22266 [FireNet Data source] Causes crash in provider when running…

MingweiZhu666:
… OCI FireNet


shuoliugit:
Will it crash if there're new image names in the future?

MingweiZhu666:
> Will it crash if there're new image names in the future?

if image name not belong to "CloudGuard Next-Gen Firewall", ("Check Point CloudGuard IaaS" &&( "Next-Gen Firewall with Threat Prevention" || "All-In-One" || "Firewall & Threat Prevention")), "Palo Alto Networks VM-Series Bundle", "Palo Alto Networks VM-Series Next Generation Firewall"
it will default use the checkVersionFormat function to remove other special character only keep the "a-zA-Z0-9.-" and then go-version library to compare the version, so if the new image version not the Semantic Version, I think it will happen again.

zhiqiangjin (approved):

PR #1274, opened 2022-04-26
AVX-22266 [FireNet Data source] Causes crash in provider when running…

MingweiZhu666:
… OCI FireNet

### DIFF
--- a/aviatrix/utils.go
+++ b/aviatrix/utils.go
@@ -3,6 +3,7 @@ package aviatrix
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"reflect"
 	"regexp"
 	"strconv"
@@ -298,6 +299,11 @@ func checkPAVMVersionFormat(version string) string {
 
 // compareVersion compares two Semantic Versions
 func compareVersion(version1, version2 string) bool {
+	defer func() {
+		if err := recover(); err != nil {
+			log.Println("not support compare this version format: ", version1, version2, err)
+		}
+	}()
 	v1, _ := version.NewVersion(version1)
 	v2, _ := version.NewVersion(version2)
 	return v1.GreaterThan(v2)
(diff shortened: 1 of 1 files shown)


aviatrix/utils.go
```diff
@@ -298,6 +299,11 @@ func checkPAVMVersionFormat(version string) string {
 
 // compareVersion compares two Semantic Versions
 func compareVersion(version1, version2 string) bool {
+	defer func() {
```
mlin-aviatrix: You probably don't need to defer here. `version.NewVersion` returns an error as its second return value. Right now you have not assigned it to anything. You can do something like this:
```
v1, err := version.NewVersion(version1)
if err != nil {
  // Log if needed
  return false
}
```

mlin-aviatrix (changes requested):

aviatrix/utils.go
MingweiZhu666: that cash error comes from the v1.GreaterThan(v2) function and I check that function doesn't have the error as its second return value.

MingweiZhu666: have fixed

aviatrix/utils.go
```diff
@@ -298,8 +299,16 @@ func checkPAVMVersionFormat(version string) string {
 
 // compareVersion compares two Semantic Versions
 func compareVersion(version1, version2 string) bool {
-	v1, _ := version.NewVersion(version1)
-	v2, _ := version.NewVersion(version2)
+	v1, err := version.NewVersion(version1)
+	if err != nil {
+		log.Println("not support sort this version format: ", version1, err)
```
mlin-aviatrix: This error message should be reworded and should probably use `log.Printf`. Something like `log.Printf("unsupported firewall image version format: %s\n", version1)`

mlin-aviatrix (changes requested):

aviatrix/utils.go
MingweiZhu666: thank you, have fixed

mlin-aviatrix (approved):

zhiqiangjin (approved):

PR #1294, opened 2022-05-10
AVX-22130 Controller Access Account RBAC Group Attachment Fails When …

MingweiZhu666:
…Using Aviatrix Provider for Terraform with 'non admin group' user

### DIFF
--- a/aviatrix/resource_aviatrix_account.go
+++ b/aviatrix/resource_aviatrix_account.go
@@ -388,11 +388,15 @@ func resourceAviatrixAccount() *schema.Resource {
 				Computed:    true,
 				Description: "AWS Secret Region CAP Certificate Key file path on the controller.",
 			},
-			"group_name": {
-				Type:        schema.TypeList,
-				Elem:        &schema.Schema{Type: schema.TypeString},
-				Optional:    true,
-				Description: "RBAC permission group name.",
+			"group_names": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Optional: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					groupNameOld, _ := d.GetChange("group_names")
+					return len(groupNameOld.([]interface{})) != 0
+				},
+				Description: "List of RBAC permission group names.",
 			},
 		},
 	}
@@ -465,8 +469,8 @@ func resourceAviatrixAccountCreate(ctx context.Context, d *schema.ResourceData, 
 		AwsSCapCertKeyPath:                    d.Get("awss_cap_cert_key_path").(string),
 	}
 
-	if _, ok := d.GetOk("group_name"); ok {
-		account.GroupName = strings.Join(goaviatrix.ExpandStringList(d.Get("group_name").([]interface{})), ",")
+	if _, ok := d.GetOk("group_names"); ok {
+		account.GroupName = strings.Join(goaviatrix.ExpandStringList(d.Get("group_names").([]interface{})), ",")
 	}
 
 	awsIam := d.Get("aws_iam").(bool)
@@ -864,9 +868,8 @@ func resourceAviatrixAccountRead(ctx context.Context, d *schema.ResourceData, me
 			d.Set("awss_cap_cert_key_path", acc.AwsSCapCertKeyPath)
 			d.Set("aws_ca_cert_path", acc.AwsCaCertPath)
 		}
-		if acc.GroupName != "" {
-			d.Set("group_name", acc.GroupName)
-		}
+
+		d.Set("group_names", acc.GroupNameReturn)
 		d.SetId(acc.AccountName)
 	}
 
@@ -953,10 +956,6 @@ func resourceAviatrixAccountUpdate(ctx context.Context, d *schema.ResourceData, 
 		AwsSCapCertKeyPath:                    d.Get("awss_cap_cert_key_path").(string),
 	}
 
-	if _, ok := d.GetOk("group_name"); ok {
-		account.GroupName = strings.Join(goaviatrix.ExpandStringList(d.Get("group_name").([]interface{})), ",")
-	}
-
 	awsIam := d.Get("aws_iam").(bool)
 	account.AwsIam = strconv.FormatBool(awsIam)
 
@@ -976,10 +975,6 @@ func resourceAviatrixAccountUpdate(ctx context.Context, d *schema.ResourceData, 
 
 	if d.HasChange("account_name") {
 		return diag.Errorf("update account name is not allowed")
-	}
-
-	if d.HasChanges("group_name") {
-		return diag.Errorf("update RBAC permission group is not allowed")
 	}
 
 	if d.HasChanges("aws_gateway_role_app", "aws_gateway_role_ec2") {
--- a/aviatrix/resource_aviatrix_account.go
+++ b/aviatrix/resource_aviatrix_account.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"strings"
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v2/goaviatrix"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -387,6 +388,12 @@ func resourceAviatrixAccount() *schema.Resource {
 				Computed:    true,
 				Description: "AWS Secret Region CAP Certificate Key file path on the controller.",
 			},
+			"group_name": {
+				Type:        schema.TypeList,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "RBAC permission group name.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_account.go
```diff
@@ -387,6 +388,12 @@ func resourceAviatrixAccount() *schema.Resource {
 				Computed:    true,
 				Description: "AWS Secret Region CAP Certificate Key file path on the controller.",
 			},
+			"group_name": {
```
mlin-aviatrix: Since the variable is a list of group names, it should be plural: `group_names`

aviatrix/resource_aviatrix_account.go
```diff
@@ -387,6 +388,12 @@ func resourceAviatrixAccount() *schema.Resource {
...
+				Type:        schema.TypeList,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Optional:    true,
+				Description: "RBAC permission group name.",
```
mlin-aviatrix: Should be: "List of RBAC permission group names."

aviatrix/resource_aviatrix_account.go
```diff
@@ -853,6 +864,9 @@ func resourceAviatrixAccountRead(ctx context.Context, d *schema.ResourceData, me
 			d.Set("awss_cap_cert_key_path", acc.AwsSCapCertKeyPath)
 			d.Set("aws_ca_cert_path", acc.AwsCaCertPath)
 		}
+		if acc.GroupName != "" {
```
mlin-aviatrix: 1. It is possible to change the RBAC groups attached to the account using the other RBAC group resources, which will cause this attribute to show a diff. Since this is a create only attribute, we need a way to prevent this diff. It should either always have a diff suppress func or should be marked computed some how.
2. Since the attribute is a list of strings, it should be converted to a slice of strings to make sure it's ok. 
3. This if check will have some errors. Lets say that after the account is created with associated RBAC groups, the API returns the empty string (this could be because the API didn't attach the groups on create but returned successfully or because a user removed the access group on the UI). The attribute will not be set and there will be no diff even though there was a change. We can just remove the if condition.

docs/resources/aviatrix_account.md
```diff
@@ -182,28 +195,33 @@ The following arguments are supported:
 * `aws_role_ec2` - (Optional) AWS EC2 role ARN, this option is for UserConnect. Required when `aws_iam` is "true" and when creating an account for AWS.
 * `aws_gateway_role_app` - (Optional) A separate AWS App role ARN to assign to gateways created by the controller. Required when `aws_gateway_role_ec2` is set. Only allowed when `aws_iam`, `awsgov_iam`, or `awschina_iam` is "true" when creating an account for AWS, AWSGov or AWSChina, respectively. Available as of provider version R2.19+.
 * `aws_gateway_role_ec2` - (Optional) A separate AWS EC2 role ARN to assign to gateways created by the controller. Required when `aws_gateway_role_app` is set. Only allowed when `aws_iam`, `awsgov_iam`, or `awschina_iam` is "true" when creating an account for AWS, AWSGov or AWSChina, respectively. Available as of provider version R2.19+.
+* `group_name` - (Optional) A list of RBAC groups name.
```
mlin-aviatrix: This can be moved to the "Misc" section so that it only shows up once.

mlin-aviatrix (changes requested):

aviatrix/resource_aviatrix_account.go
MingweiZhu666: add DiffSuppressFunc to prevent each update will no change.

aviatrix/resource_aviatrix_account.go
MingweiZhu666: fixed

docs/resources/aviatrix_account.md
MingweiZhu666: fixed

aviatrix/resource_aviatrix_account.go
MingweiZhu666: have fixed

goaviatrix/account.go
```diff
@@ -78,8 +78,10 @@ type Account struct {
 	AwsSCapCert                           string
 	AwsSCapCertKey                        string
 	AwsSCaChainCert                       string
-	AwsSCapCertPath                       string `form:"aws_red_cap_cert_path,omitempty" json:"aws_red_cap_cert_path,omitempty"`
-	AwsSCapCertKeyPath                    string `json:"aws_red_cap_key_path,omitempty"`
+	AwsSCapCertPath                       string   `form:"aws_red_cap_cert_path,omitempty" json:"aws_red_cap_cert_path,omitempty"`
+	AwsSCapCertKeyPath                    string   `json:"aws_red_cap_key_path,omitempty"`
+	GroupName                             string   `form:"groups,omitempty"`
```
mlin-aviatrix: This should also be changed to `GroupNames`

goaviatrix/account.go
```diff
@@ -78,8 +78,10 @@ type Account struct {
...
+	GroupNameReturn                       []string `json:"rbac_groups,omitempty"`
```
mlin-aviatrix: I think we normally use attribute name + Read for read only attributes. So in this case it would be `GroupNamesRead`

aviatrix/resource_aviatrix_account.go
```diff
@@ -387,6 +388,16 @@ func resourceAviatrixAccount() *schema.Resource {
 				Computed:    true,
 				Description: "AWS Secret Region CAP Certificate Key file path on the controller.",
 			},
+			"group_names": {
+				Type:     schema.TypeList,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				Optional: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					groupNameOld, _ := d.GetChange("group_names")
+					return len(groupNameOld.([]interface{})) != 0
```
mlin-aviatrix, line 397: The `old` and `new` parameters passed into the DiffSuppressFunc should already be the values that `d.GetChange` returns. You shouldn't need to call that function.

Also, can you list which cases you tested so that we can be sure the DIffSuppressFunc works as intended. This list can just be added as a comment to the JIRA ticket.

mlin-aviatrix (changes requested):

aviatrix/resource_aviatrix_account.go
mlin-aviatrix: Also did you test what happens when the RBAC groups are changed outside this resource? Does the return value from the API actually change?

MingweiZhu666: yes, I have tested:
first try to create an Access account: "access-account1", give this account with two RBAC group: ["group-a", "group-b"], then, when we delete one group from UI, and terraform apply again, it will show "No changes."
second test is when we use the resource: "aviatrix_rbac_group_access_account_attachment" to add a RBAC group: "group-c" to the Access account: "access-account1", and then terraform apply again, it shows "No changes".

I will add this test case in the JIRA ticket:
# Create an Aviatrix AWS Account with IAM roles
resource "aviatrix_account" "temp_acc_aws" {
  account_name       = "access-account1"
  cloud_type         = 1
  aws_account_number = "722837227438"
  aws_iam            = true
  aws_role_app       = "arn:aws:iam::722837227438:role/aviatrix-role-app"
  aws_role_ec2       = "arn:aws:iam::722837227438:role/aviatrix-role-ec2"
  group_names        = ["group-a", "group-b"]
}

# Create an Aviatrix RBAC Group Access Account Attachment
resource "aviatrix_rbac_group_access_account_attachment" "test_attachment1" {
  group_name          = "group-c"
  access_account_name = "access-account1"
}

docs/resources/aviatrix_account.md
```diff
@@ -255,6 +268,7 @@ The following arguments are supported:
 ### Misc.
 ~> **NOTE:** On Terraform versions 0.12.x, 0.13.x, and 0.14.x, Terraform will not detect any changes to the account when the account audit fail warning is given. In order to apply changes or set `audit_account = false`, please run `terraform apply -refresh=false`. 
 * `audit_account` - (Optional) Specify whether to enable the audit account feature. If this feature is enabled, terraform will give a warning if there is an issue with the account credentials. Changing `audit_account` to "false" will not prevent the Controller from performing account audits. It will only prevent Terraform from displaying a warning. Valid values: true, false. Default: false. Available as of provider version 2.19+. **Note: The warning may still appear for a few hours after fixing the underlying issue.**
+* `group_names` - (Optional) A list of RBAC groups name. This attribute not allow update in aviatrix_account resource, if update will show no change. Only allow to add the RBAC group which under the current user account
```
mlin-aviatrix: The grammar in this should be better. Something like:
`A list of existing RBAC group names.  This attribute should only be used when creating an account. Updating this attribute will have no effect.`

mlin-aviatrix (changes requested):

mlin-aviatrix (approved):

aviatrix/resource_aviatrix_account.go
shuoliugit: > The `old` and `new` parameters passed into the DiffSuppressFunc should already be the values that `d.GetChange` returns. You shouldn't need to call that function.
> 
> Also, can you list which cases you tested so that we can be sure the DIffSuppressFunc works as intended. This list can just be added as a comment to the JIRA ticket.

The `old` and `new` may generate undefined behaviors here because this attribute is a list.

zhiqiangjin (approved):

PR #1295, opened 2022-05-18
AVX-21557 Create data source that lists all transit gateways and thei…

MingweiZhu666:
…r attributes


aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -0,0 +1,870 @@
...
+	"github.com/AviatrixSystems/terraform-provider-aviatrix/v2/goaviatrix"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAviatrixAllTransitGatewayRead,
+
+		Schema: map[string]*schema.Schema{
+			"all_transit_gateway": {
```
mlin-aviatrix: I don't see why this attribute is needed. All it does is block the data source from reading, but I feel like the customer can just remove the data source to have the same effect.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -0,0 +1,870 @@
...
+	if err != nil {
+		return err
+	}
+	var result []map[string]interface{}
+	for i := range TransitGatewayNameList {
+		transitGatewayName := TransitGatewayNameList[i]
+		gateway := &goaviatrix.Gateway{
+			GwName: transitGatewayName,
+		}
+		gw, err := client.GetGateway(gateway)
```
mlin-aviatrix: This isn't right, you shouldn't be calling an API for every single gateway as this resource will take too long to read for a large number of gateways. If you call `list_vpcs_summary` without the `gateway_name` attribute, it will return information for all gateways, which should be better than calling `list_vpcs_summary` each time for every gateway. It looks like you are already doing that when you call `GetTransitGatewayNameList`. Just use that information instead of filtering only the gateway names out.

Also, when calling `list_vpcs_summary` without `gateway_name`, some attributes will not be returned, which is ok. You should just remove those from the data source.

aviatrix/data_source_aviatrix_all_transit_gateways_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func TestAccDataSourceAviatrixAllTransitGateways_basic(t *testing.T) {
+	rName := acctest.RandString(5)
+	resourceName := "data.aviatrix_all_transit_gateways.foo"
+
+	skipAcc := os.Getenv("SKIP_DATA_ALL_TRANSIT_GATEWAY")
+	if skipAcc == "yes" {
+		t.Skip("Skipping Data Source All Transit Gateway tests as SKIP_DATA_ALL_TRANSIT_GATEWAY is set")
+	}
+	if skipAcc != "yes" {
```
mlin-aviatrix: Since it is not possible for `skipAcc == "yes" && skipAcc != "yes"` to be false, this second if statement should not be required and can be removed.

aviatrix/data_source_aviatrix_all_transit_gateways_test.go
```diff
@@ -0,0 +1,126 @@
...
+			},
+		})
+	} else {
+		t.Log("Skipping Date Source All Transit Gateway test in AWS as SKIP_DATA_ALL_TRANSIT_GATEWAY_AWS is set")
+	}
+}
+
+func testAccDataSourceAviatrixAllTransitGatewaysConfigPreBasic(rName string) string {
+	gcpGwSize := os.Getenv("GCP_GW_SIZE")
+	if gcpGwSize == "" {
```
mlin-aviatrix: If you are using a new environment variable for this test, in this case `GCP_GW_SIZE`, you should add it to test-infra/README_accep_test.md. You shouldn't need to hard code it here.

docs/data-sources/aviatrix_all_transit_gateways.md
```diff
@@ -0,0 +1,87 @@
...
+
+The following arguments are supported:
+
+* `all_transit_gateway` - (Required) True or False.
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `transit_gateway_list` - A list include all transit gateways
```
mlin-aviatrix: This should be reworded. Maybe `The list of all transit gateways`

Also, everything underneath should be indented to indicate that it is part of the block.

mlin-aviatrix (changes requested):

aviatrix/data_source_aviatrix_all_transit_gateways.go
MingweiZhu666: yes, I have remove this attribute, currently the data source is like:
data "aviatrix_all_transit_gateways" "foo" {

}


which doesn't need any attribute

aviatrix/data_source_aviatrix_all_transit_gateways.go
MingweiZhu666: have fixed, currently will use the GetTransitGatewayList() func, which returns a list including the all transit gateways, by this way only call on time.

aviatrix/data_source_aviatrix_all_transit_gateways_test.go
MingweiZhu666: have fixed, remove this part

aviatrix/data_source_aviatrix_all_transit_gateways_test.go
MingweiZhu666: add the request variable in test-infra/README_accep_test.md

docs/data-sources/aviatrix_all_transit_gateways.md
MingweiZhu666: have fixed

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
+		gw := TransitGatewayList[i]
+		transitGateway := make(map[string]interface{})
+		transitGateway["cloud_type"] = gw.CloudType
+		transitGateway["account_name"] = gw.AccountName
+		transitGateway["gw_name"] = gw.GwName
+		transitGateway["subnet"] = gw.VpcNet
+		if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.AWSRelatedCloudTypes) {
+			transitGateway["vpc_id"] = strings.Split(gw.VpcID, "~~")[0]
+			transitGateway["vpc_reg"] = gw.VpcRegion
+			if gw.AllocateNewEipRead {
```
mlin-aviatrix: This if statement is not needed. You can just set `transitGateway["allocate_new_eip"] = gw.AllocateNewEipRead`

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
 			} else {
-				transitGateway["single_ip_snat"] = false
+				transitGateway["allocate_new_eip"] = false
 			}
-			if gw.SingleAZ == "yes" {
-				transitGateway["single_az_ha"] = true
+		} else if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.GCPRelatedCloudTypes) {
+			transitGateway["vpc_id"] = gw.VpcID
+			transitGateway["vpc_reg"] = gw.GatewayZone
+			if gw.AllocateNewEipRead {
```
mlin-aviatrix: Same as above

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
-				return fmt.Errorf("couldn't get Aviatrix Transit Gateway: %s", err)
+				log.Printf("[WARN] Failed to get tags for transit gateway %s: %v", tags.ResourceName, err)
 			}
 
-			transitGateway["enable_firenet"] = gwDetail.EnableFireNet
-			transitGateway["enable_transit_firenet"] = gwDetail.EnableTransitFireNet
-			transitGateway["enable_egress_transit_firenet"] = gwDetail.EnableEgressTransitFireNet
-			transitGateway["customized_transit_vpc_routes"] = gwDetail.CustomizedTransitVpcRoutes
+			if len(tagList) > 0 {
+				transitGateway["tag_list"] = tagList
```
mlin-aviatrix: `tag_list` is deprecated, you should be using the `tags` attribute from Transit Gateway. On a separate note, I feel like we should not be calling any other APIs in this data source. Since tags information is not given by default, it should not be part of this resource.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
 
-			if len(gw.FilteredSpokeVpcRoutes) != 0 {
-				transitGateway["filtered_spoke_vpc_routes"] = strings.Join(gw.FilteredSpokeVpcRoutes, ",")
+		if len(gw.ExcludeCidrList) != 0 {
+			transitGateway["excluded_advertised_spoke_routes"] = strings.Join(gw.ExcludeCidrList, ",")
+		} else {
+			transitGateway["excluded_advertised_spoke_routes"] = ""
+		}
+
+		gwDetail, err := client.GetGatewayDetail(&gw)
```
mlin-aviatrix: Same comment as below, we should not be calling any other APIs.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
+		if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.OCIRelatedCloudTypes) {
+			transitGateway["availability_domain"] = gw.GatewayZone
+			transitGateway["fault_domain"] = gw.FaultDomain
+		}
+
+		haGateway := &goaviatrix.Gateway{
+			AccountName: gw.AccountName,
+			GwName:      gw.GwName + "-hagw",
+		}
+		haGw, _ := client.GetGateway(haGateway)
```
mlin-aviatrix: Same as above.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
-
-				tagList, err := client.GetTags(tags)
-				if err != nil {
-					log.Printf("[WARN] Failed to get tags for transit gateway %s: %v", tags.ResourceName, err)
-				}
 
-				if len(tagList) > 0 {
-					transitGateway["tag_list"] = tagList
-				}
+			lanCidr, err := client.GetTransitGatewayLanCidr(gw.HaGw.GwName)
```
mlin-aviatrix: Same as above.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
+		if gw.EnableTransitFirenet && goaviatrix.IsCloudType(gw.CloudType, goaviatrix.GCPRelatedCloudTypes) {
+			transitGateway["lan_vpc_id"] = gw.BundleVpcInfo.LAN.VpcID
+			transitGateway["lan_private_subnet"] = strings.Split(gw.BundleVpcInfo.LAN.Subnet, "~~")[0]
+		}
 
-			if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.OCIRelatedCloudTypes) {
-				transitGateway["availability_domain"] = gw.GatewayZone
-				transitGateway["fault_domain"] = gw.FaultDomain
+		if gw.EnableLearnedCidrsApproval {
+			transitAdvancedConfig, err := client.GetTransitGatewayAdvancedConfig(&goaviatrix.TransitVpc{GwName: gw.GwName})
```
mlin-aviatrix: Same as above.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
-				transitGateway["approved_learned_cidrs"] = nil
+				transitGateway["ha_bgp_lan_interfaces"] = haInterfaces
 			}
 
-			var prependAsPath []string
-			for _, p := range strings.Split(gw.PrependASPath, " ") {
-				if p != "" {
-					prependAsPath = append(prependAsPath, p)
-				}
+			bgpLanIpInfo, err := client.GetBgpLanIPList(&goaviatrix.TransitVpc{GwName: gw.GwName})
```
mlin-aviatrix: Same as above.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -516,349 +511,331 @@ func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
...
+			if len(azureEip) == 3 {
+				transitGateway["azure_eip_name_resource_group"] = fmt.Sprintf("%s:%s", azureEip[0], azureEip[1])
+			} else {
+				log.Printf("[WARN] could not get Azure EIP name and resource group for the Transit Gateway %s", gw.GwName)
 			}
-			transitGateway["lan_interface_cidr"] = lanCidr
+		}
 
-			result = append(result, transitGateway)
+		lanCidr, err := client.GetTransitGatewayLanCidr(gw.GwName)
```
mlin-aviatrix: Same as above.

mlin-aviatrix (changes requested):

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -0,0 +1,558 @@
+package aviatrix
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	//"github.com/AviatrixSystems/terraform-provider-aviatrix/v2/goaviatrix"
```
mlin-aviatrix: This can be removed.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -0,0 +1,558 @@
...
+
+func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAviatrixAllTransitGatewayRead,
+
+		Schema: map[string]*schema.Schema{
+			"transit_gateway_list": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "list of all transit gateways and attributes",
```
mlin-aviatrix: This can just be: `List of all Transit Gateways`

mlin-aviatrix (changes requested):

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -0,0 +1,557 @@
...
+
+func dataSourceAviatrixListAllTransitGateway() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAviatrixAllTransitGatewayRead,
+
+		Schema: map[string]*schema.Schema{
+			"transit_gateway_list": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "List of all Transit Gateways",
```
mlin-aviatrix: Should have a period at the end of the description.

aviatrix/data_source_aviatrix_all_transit_gateways.go
```diff
@@ -0,0 +1,557 @@
...
+			}
+		}
+
+		result = append(result, transitGateway)
+	}
+
+	if err = d.Set("transit_gateway_list", result); err != nil {
+		return fmt.Errorf("couldn't set transit_gateway_list: %s", err)
+	}
+	d.SetId("transit-gateway-list-id")
```
mlin-aviatrix: The ID for this data source should probably be the Controller IP. We want the data source to be unique per Controller, but not necessarily unique per configuration. You can refer to `aviatrix_controller_bgp_max_as_limit_config` as an example of how to set ID.

mlin-aviatrix (changes requested):

mlin-aviatrix (dismissed):

docs/data-sources/aviatrix_all_transit_gateways.md
```diff
@@ -0,0 +1,60 @@
...
+
+# aviatrix_all_transit_gateways
+
+The **aviatrix_all_transit_gateways** data source provides details about all transit gateways created by the Aviatrix Controller.
+
+## Example Usage
+
+```hcl
+# Aviatrix All Transit Gateways Data Source
+data "aviatrix_all_transit_gateways" "foo" {
```
mlin-aviatrix: Since this data source has no input attributes, we probably don't need the extra newlines e.g. 
`data "aviatrix_all_transit_gateways" "foo" {}`

MingweiZhu666: have fixed

mlin-aviatrix (dismissed):

aviatrix/data_source_aviatrix_transit_gateways.go
```diff
@@ -0,0 +1,557 @@
...
+	"github.com/AviatrixSystems/terraform-provider-aviatrix/v2/goaviatrix"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceAviatrixTransitGateways() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceAviatrixTransitGatewaysRead,
+
+		Schema: map[string]*schema.Schema{
+			"transit_gateway_list": {
```
zhiqiangjin: Maybe change it to "gateway_list"

zhiqiangjin (changes requested):

aviatrix/data_source_aviatrix_transit_gateways.go
MingweiZhu666: have fixed

zhiqiangjin (approved):

PR #1305, opened 2022-05-25
AVX-22130 Controller Access Account RBAC Group Attachment Fails When …

MingweiZhu666:
…Using Aviatrix Provider for Terraform with 'non admin group' user(fix attribute name)

### DIFF
--- a/docs/resources/aviatrix_account.md
+++ b/docs/resources/aviatrix_account.md
@@ -23,7 +23,6 @@ resource "aviatrix_account" "temp_acc_aws" {
   aws_iam            = true
   aws_role_app       = "arn:aws:iam::123456789012:role/aviatrix-role-app"
   aws_role_ec2       = "arn:aws:iam::123456789012:role/aviatrix-role-ec2"
-  rbac_groups        = ["write_only"]
 }
 ```
 ```hcl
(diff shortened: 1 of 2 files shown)


shuoliugit (dismissed):

docs/resources/aviatrix_account.md
```diff
@@ -23,7 +23,7 @@ resource "aviatrix_account" "temp_acc_aws" {
   aws_iam            = true
   aws_role_app       = "arn:aws:iam::123456789012:role/aviatrix-role-app"
   aws_role_ec2       = "arn:aws:iam::123456789012:role/aviatrix-role-ec2"
-  group_names        = ["write_only"]
+  rbac_groups        = ["write_only"]
```
mlin-aviatrix: By adding the `rbac_groups` attribute to every example in the documentation, we are effectively saying that by default the `write_only` RBAC group should be used. Is this the intended default group or should we only add a separate example section for setting RBAC groups? What do you guys think @shuoliugit @zhiqiangjin @inc1t3Ful ?

MingweiZhu666: the write_only RBAC group is not the default group, user must to use aviatrix_rbac_group resource to create their own rbac group, just like:

resource "aviatrix_rbac_group" "test_group" {
  group_name = "write_only"
}

mlin-aviatrix (approved):

zhiqiangjin (approved):

PR #1317, opened 2022-06-06
AVX-24069: add support for conn as path prepend for bgpspoke transit atch

shuoliugit:
This is a feature for 6.8.

The `route_tables` option is not working for BGPSpoke now, but backend suggested leaving it there. Once backend change is in, it will work.

### DIFF
--- a/aviatrix/resource_aviatrix_spoke_transit_attachment.go
+++ b/aviatrix/resource_aviatrix_spoke_transit_attachment.go
@@ -14,6 +14,7 @@ func resourceAviatrixSpokeTransitAttachment() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceAviatrixSpokeTransitAttachmentCreate,
 		Read:   resourceAviatrixSpokeTransitAttachmentRead,
+		Update: resourceAviatrixSpokeTransitAttachmentUpdate,
 		Delete: resourceAviatrixSpokeTransitAttachmentDelete,
 		Importer: &schema.ResourceImporter{
 			State: schema.ImportStatePassthrough,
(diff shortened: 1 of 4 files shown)


aviatrix/resource_aviatrix_spoke_transit_attachment.go
```diff
@@ -130,10 +190,88 @@ func resourceAviatrixSpokeTransitAttachmentRead(d *schema.ResourceData, meta int
...
+			for _, str := range strings.Split(transitGatewayPeering.PrependAsPath2, " ") {
+				prependAsPath = append(prependAsPath, strings.TrimSpace(str))
+			}
+
+			err = d.Set("transit_prepend_as_path", prependAsPath)
+			if err != nil {
+				return fmt.Errorf("could not set transit_prepend_as_path: %v", err)
+			}
+		}
+	}
```
mlin-aviatrix, line 231: We should explicitly set `spoke_prepend_as_path` and `transit_prepend_as_path` to nil if `SpokeBgpEnabled = false`

aviatrix/resource_aviatrix_spoke_transit_attachment.go
```diff
@@ -50,6 +76,15 @@ func resourceAviatrixSpokeTransitAttachmentCreate(d *schema.ResourceData, meta i
 
 	attachment := marshalSpokeTransitAttachmentInput(d)
 
+	spoke, err := client.GetGateway(&goaviatrix.Gateway{GwName: attachment.SpokeGwName})
```
mlin-aviatrix, line 79: Is this API call needed? Should we be validating this on Terraform side or should we be leaving it to the API/backend to validate? 

I have something similar in Private Mode where an attribute can only be used on Transit/Spoke Gateways if the Controller is in Private Mode, but it seems expensive to make every Spoke/Transit Gateway have that API call in create. Especially for non-Private Mode setups. In theory, create is only called once per lifetime for a Gateway, so the cost might not be that big.

What do you think @zhiqiangjin?

aviatrix/resource_aviatrix_spoke_transit_attachment.go
```diff
@@ -130,10 +190,88 @@ func resourceAviatrixSpokeTransitAttachmentRead(d *schema.ResourceData, meta int
...
 
+	if attachment.SpokeBgpEnabled {
+		transitGatewayPeering := &goaviatrix.TransitGatewayPeering{
+			TransitGatewayName1: spokeGwName,
+			TransitGatewayName2: transitGwName,
+		}
+
+		transitGatewayPeering, err = client.GetTransitGatewayPeeringDetails(transitGatewayPeering)
+		if err == goaviatrix.ErrNotFound {
+			d.SetId("")
```
mlin-aviatrix, line 201: Should the resource be marked as gone here? If yes, can this information be returned in `client.GetSpokeTransitAttachment`?

mlin-aviatrix (changes requested):

aviatrix/resource_aviatrix_spoke_transit_attachment.go
shuoliugit: I think it should be marked as gone. bgpspoke-transit conn is just like transit peering, if it can't be found, it's gone.

bgpspoke-transit conn and spoke-transit conn are different, backend only returns info about bgpspoke-transit conn in API `get_inter_transit_gateway_peering_details`.

aviatrix/resource_aviatrix_spoke_transit_attachment.go
shuoliugit: Talked with Tim yesterday afternoon, let's keep it for now.

mlin-aviatrix (approved):

zhiqiangjin (approved):

PR #1322, opened 2022-06-09
AVX-15352: Add support for Private Mode feature

mlin-aviatrix:


docs/resources/aviatrix_controller_private_mode_config.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+### Required
+* `enable_private_mode` - (Required) Whether to enable Private Mode on an Aviatrix Controller.
+* `copilot_instance_id` - (Optional) Instance ID of a copilot instance to associate with an Aviatrix Controller in Private Mode. The copilot instance must be in the same VPC as the Aviatrix Controller.
+* `proxies` - (Optional) List of Controller proxies for Private Mode.
+
```
shuoliugit, line 34: 1. copilot_instance_id and proxies are optional, should be in the optional category
2. proxies is a set, not list.

aviatrix/resource_aviatrix_controller_private_mode_config.go
```diff
@@ -0,0 +1,191 @@
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Copilot instance ID to associate with the Controller for Private Mode.",
+			},
+			"proxies": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "List of proxies.",
+			},
```
shuoliugit, line 38: set of proxies

docs/resources/aviatrix_private_mode_multicloud_endpoint.md
```diff
@@ -0,0 +1,44 @@
...
+The **aviatrix_private_mode_multicloud_endpoint** resource allows management of a Private Mode multicloud endpoint. This resource is available as of provider version R2.23+.
+
+## Example Usage
+
+```hcl
+# Create an Aviatrix Controller Private Mode config
+resource "aviatrix_private_mode_multicloud_endpoint" "test" {
+  enable_private_mode = true
+}
+```
```
shuoliugit, line 23: required attributes missing

docs/resources/aviatrix_spoke_gateway.md
```diff
@@ -335,6 +349,9 @@ The following arguments are supported:
...
 * `enable_bgp` - (Optional) Enable BGP for this spoke gateway. Only available for AWS and Azure. Valid values: true, false. Default value: false. Available in provider R2.21.0+.
 * `rx_queue_size` - (Optional) Gateway ethernet interface RX queue size. Once set, can't be deleted or disabled. Available for AWS as of provider version R2.22+.
+* `private_mode_lb_vpc_id` - (Optional) VPC ID of Private Mode load balancer. Required when Private Mode is enabled on the Controller. Available in provider version R2.23+.
+* `private_mode_subnet_zone` - (Optional) Availability Zone of the subnet. Required when Private Mode is enabled on the Controller and `cloud_type` is AWS or AWSGov(). Available in Provider version R2.23+.
+
```
shuoliugit, line 354: remove ()

docs/resources/aviatrix_transit_gateway.md
```diff
@@ -410,6 +424,8 @@ The following arguments are supported:
...
 * `tunnel_detection_time` - (Optional) The IPsec tunnel down detection time for the Transit Gateway in seconds. Must be a number in the range [20-600]. The default value is set by the controller (60 seconds if nothing has been changed). **NOTE: The controller UI has an option to set the tunnel detection time for all gateways. To achieve the same functionality in Terraform, use the same TF_VAR to manage the tunnel detection time for all gateways.** Available in provider R2.19+.
 * `rx_queue_size` - (Optional) Gateway ethernet interface RX queue size. Once set, can't be deleted or disabled. Available for AWS as of provider version R2.22+.
+* `private_mode_lb_vpc_id` - (Optional) VPC ID of Private Mode load balancer. Required when Private Mode is enabled on the Controller. Available in Provider version R2.23+.
+* `private_mode_subnet_zone` - (Optional) Availability Zone of the subnet. Required when Private Mode is enabled on the Controller and `cloud_type` is AWS or AWSGov(). Available in Provider version R2.23+.
```
shuoliugit, line 429: remove ()

test-infra/cmdExportOutput.sh
```diff
@@ -131,6 +131,7 @@ SetEnv SKIP_CONTROLLER_CERT_DOMAIN_CONFIG "no"
 SetEnv SKIP_CONTROLLER_CONFIG "no"
 SetEnv SKIP_CONTROLLER_EMAIL_EXCEPTION_NOTIFICATION_CONFIG "no"
 SetEnv SKIP_CONTROLLER_GATEWAY_KEEPALIVE_CONFIG "no"
+SetEnv SIP_CONTROLLER_PRIVATE_MODE_CONFIG "no"
 SetEnv SKIP_CONTROLLER_PRIVATE_OOB "no"
```
shuoliugit, line 135: SKIP

aviatrix/resource_aviatrix_controller_private_mode_config.go
```diff
@@ -0,0 +1,191 @@
...
+
+func resourceAviatrixControllerPrivateModeConfigCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*goaviatrix.Client)
+
+	enablePrivateMode := d.Get("enable_private_mode").(bool)
+	if !enablePrivateMode {
+		for _, key := range []string{"copilot_instance_id", "proxies"} {
+			if _, ok := d.GetOk(key); ok {
+				return diag.Errorf("failed to create Controller Private Mode config: %q must be empty when %q is false", key, "enable_private_mode")
+			}
```
shuoliugit, line 51: Maybe in the error message we can list all the attributes that should be empty. Otherwise, if the user set both, it will error out twice if the user just correct one attribute each time.

aviatrix/resource_aviatrix_controller_private_mode_config.go
```diff
@@ -0,0 +1,191 @@
...
+				return diag.Errorf("failed to set Controller proxies: %s", err)
+			}
+		}
+	} else {
+		err := client.DisablePrivateMode(ctx)
+		if err != nil {
+			return diag.Errorf("failed to disable Private Mode: %s", err)
+		}
+	}
+
```
shuoliugit, line 90: Do we need to return the error if private mode is already disabled? Should we just log it like controller_security_group_management?
@zhiqiangjin

aviatrix/resource_aviatrix_private_mode_lb_test.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		vpcId := rs.Primary.ID
+		_, err := client.GetPrivateModeLoadBalancer(context.Background(), vpcId)
+		if err != nil {
+			if err == goaviatrix.ErrNotFound {
+				return nil
+			}
+			return fmt.Errorf("failed to destroy Private Mode load balancer: %s", err)
+		}
```
shuoliugit, line 123: this error message is not very accurate, getter failed.

aviatrix/resource_aviatrix_private_mode_multicloud_endpoint.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		return diag.Errorf("failed to get Private Mode multicloud endpoint: %s", err)
+	}
+
+	d.Set("account_name", privateModeMulticloudEndpoint.AccountName)
+	d.Set("vpc_id", privateModeMulticloudEndpoint.VpcId)
+	d.Set("region", privateModeMulticloudEndpoint.Region)
+	d.Set("controller_lb_vpc_id", privateModeMulticloudEndpoint.ControllerLbVpcId)
+	d.Set("dns_entry", privateModeMulticloudEndpoint.DnsEntry)
+
```
shuoliugit, line 110: id is not set

aviatrix/resource_aviatrix_private_mode_multicloud_endpoint_test.go
```diff
@@ -0,0 +1,149 @@
...
+			return fmt.Errorf("failed to destroy Private Mode load balancer")
+		} else if rs.Type == "aviatrix_private_mode_multicloud_load_balancer" {
+			vpcId := rs.Primary.ID
+			_, err := client.GetPrivateModeMulticloudEndpoint(context.Background(), vpcId)
+
+			if err != nil {
+				if err == goaviatrix.ErrNotFound {
+					continue
+				}
+				return fmt.Errorf("failed to destroy Private Mode multicloud endpoint: %s", err)
```
shuoliugit: same

aviatrix/resource_aviatrix_controller_private_mode_config.go
mlin-aviatrix: That is an interesting point. One of the interesting features of using the Context methods and returning diag.Diagnostics is that we can return multiple errors at the same time. This could be something to consider here.

Another feature is that we can specify the exact attribute that caused the error, which should make fixing the error easier. I was testing something like this on microseg policies because if multiple policies have errors, it's hard to quickly fix it.

aviatrix/resource_aviatrix_controller_private_mode_config.go
zhiqiangjin: We should be consistent with this type of error.

mlin-aviatrix: If Private Mode is already disabled, the `DisablePrivateMode` method will not return an error. I don't think this needs a change.

aviatrix/resource_aviatrix_private_mode_lb.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	if d.HasChange("proxies") {
+		err := client.UpdatePrivateModeMulticloudProxies(ctx, privateModeLb)
+		if err != nil {
+			return diag.Errorf("failed to set Multicloud proxies during Private Mode Controller load balancer"+
+				"update: %s", err)
+		}
+	}
+
```
shuoliugit, line 219: also need to take care of other attributes, either forcenew or disallow update.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -489,6 +490,17 @@ func resourceAviatrixTransitGateway() *schema.Resource {
...
+				Optional:      true,
+				Description:   "Private Mode Controller load balancer VPC ID. Required when private mode is enabled for the Controller.",
+				ConflictsWith: []string{"allocate_new_eip"},
+			},
+			"private_mode_subnet_zone": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "Private Mode subnet availability zone.",
+			},
 			"availability_domain": {
```
shuoliugit, line 506: For the new attributes in spoke, transit and vpc, can they be updated?

mlin-aviatrix: Good catch, they should be force new.

shuoliugit (approved):

zhiqiangjin (approved):

PR #1323, opened 2022-06-09
AVX-22358: Add ability to ignore tags across all resources

mlin-aviatrix:


aviatrix/provider.go
```diff
@@ -42,6 +44,30 @@ func Provider() *schema.Provider {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"ignore_tags": {
```
zhiqiangjin, line 47: Is it a good name?

zhiqiangjin (changes requested):

aviatrix/provider.go
mlin-aviatrix: This is what aws uses on their provider.

zhiqiangjin (approved):

PR #1329, opened 2022-06-16
AVX-24635: EaaS BGP over LAN ext dev conn

shuoliugit:


test-infra/variables.tf
```diff
@@ -67,3 +67,5 @@ variable "type" {default="metered"}
 variable "customer_id" {default = ""}
 
 variable "datadog_api_key" {}
+
+variable "edge_spoke_site_id" {}
```
mlin-aviatrix, line 72: What is this variable used for?

aviatrix/resource_aviatrix_edge_spoke_external_device_conn_test.go
```diff
@@ -0,0 +1,116 @@
...
+resource "aviatrix_edge_spoke_external_device_conn" "test" {
+	site_id           = "%s"
+	connection_name   = "%s"
+	gw_name           = "%s"
+	bgp_local_as_num  = "65001"
+	bgp_remote_as_num = "65002"
+	local_lan_ip      = "1.2.3.4"
+	remote_lan_ip     = "5.6.7.8"
+}
+	`, os.Getenv("EDGE_SPOKE_SITE_ID"), rName, os.Getenv("EDGE_SPOKE_NAME"))
```
mlin-aviatrix, line 74: I think we should check that these environment variables are set in the PreCheck above.

mlin-aviatrix (changes requested):

test-infra/variables.tf
shuoliugit: That's the "VPC ID" for EaaS. In order to pass the test, EaaS has to be up. Some manual work is required to make an EaaS up, so we have to prepare a running EaaS and provide these required variables.

mlin-aviatrix (dismissed):

zhiqiangjin (approved):

PR #1345, opened 2022-07-07
AVX-23092: add new attribute rx_queue_size to EaaS

shuoliugit:


aviatrix/resource_aviatrix_edge_spoke.go
```diff
@@ -510,6 +525,7 @@ func resourceAviatrixEdgeSpokeRead(ctx context.Context, d *schema.ResourceData,
 		d.Set("longitude", "")
 	}
 	d.Set("wan_public_ip", edgeSpoke.WanPublicIp)
+	d.Set("rx_queue_size", edgeSpoke.RxQueueSize)
```
zhiqiangjin, line 528: Here may be an issue. If edgeSpoke.RxQueueSize is empty, you still set rx_queue_size.

shuoliugit: It's ok to set without if statement in this case, so the if statement is removed.

zhiqiangjin (approved):

PR #1533, opened 2023-01-11
AVX-34278: Fix aws peering force replace issue

zhiqiangjin:
1. Ignore the order of route tables in rtb_list1 and rtb_list2 since the order does not matter;
2. Remove support of value ["all"] for rtb_list1 and rtb_list2.

### DIFF
--- a/docs/resources/aviatrix_aws_peer.md
+++ b/docs/resources/aviatrix_aws_peer.md
@@ -44,9 +44,10 @@ The following arguments are supported:
 * `rtb_list1` - (Optional) List of Route table IDs of VPC1. Example: ["rtb-abcd1234", "rtb-wxyz5678"].
 * `rtb_list2` - (Optional) List of Route table IDs of VPC2. Example: ["rtb-abcd1234", "rtb-wxyz5678"].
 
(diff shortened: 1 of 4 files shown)


shuoliugit (dismissed):

docs/resources/aviatrix_aws_peer.md
```diff
@@ -44,6 +44,10 @@ The following arguments are supported:
 * `rtb_list1` - (Optional) List of Route table IDs of VPC1. Example: ["rtb-abcd1234", "rtb-wxyz5678"].
 * `rtb_list2` - (Optional) List of Route table IDs of VPC2. Example: ["rtb-abcd1234", "rtb-wxyz5678"].
 
+~> **NOTE:** For attributes `rtb_list1` and `rtb_list2`, only valid route table ID with prefix "rtb-" is supported and
```
inc1t3Ful: ... , only valid route IDs with prefix "rtb-" are supported. Therefore, "all" will no longer supported as a valid input as of 3.0.1 onward. If an **aviatrix_aws_peer** resource was created with provider 3.0.0- and any `rtb_list1` or `rtb_list2` was set as ["all"], it will need to be updated to a list of valid route table IDs.

inc1t3Ful (changes requested):

inc1t3Ful (approved):

PR #1603, opened 2023-02-16
AVX-36157: support disabling auto LAN CIDR propagation in edge csp

shuoliugit:


aviatrix/resource_aviatrix_edge_csp.go
```diff
@@ -407,6 +407,12 @@ func resourceAviatrixEdgeCSP() *schema.Resource {
 				Default:     false,
 				Description: "Enable Single IP SNAT.",
 			},
+			"auto_advertise_lan_cidrs": {
```
zhiqiangjin: Can we use "enable_auto_advertise_lan_cidrs" instead? It seems a type of list not boolean.

zhiqiangjin (changes requested):

zhiqiangjin (approved):

PR #1703, opened 2023-03-31
AVX-27582: Add support of disabling GRO/GSO on gateways (#1701)

zhiqiangjin:


shuoliugit (approved):

zhiqiangjin:
AVX-27582 would be only fixed in 7.1/3.1.0. Closed this PR.

PR #1752, opened 2023-04-29
AVX-40376: remove subnet validation in spoke_ha_gateway

shuoliugit:


arvindsree:
We should only change at API layer

zhiqiangjin:
Have been fixed in backend:
https://aviatrix.atlassian.net/browse/AVX-40376

PR #1753, opened 2023-04-29
AVX-40482: remove resources not for 7.1

shuoliugit:


ajain621 (commented):
is there no easy way to disable it than removing the whole code?

shuoliugit:
> is there no easy way to disable it than removing the whole code?

If we only disable the resources, the code will still be exposed to the public. It's better not to allow users to try the unsupported features.

zhiqiangjin (approved):

PR #1774, opened 2023-05-16
AVX-41013: Add a check for reading transit gateway peerings

zhiqiangjin:


shuoliugit (approved):

pwp333:
Need to do the same check for all code base

PR #1849, opened 2023-08-11
AVX-44451: add support of spot instance for azure

shuoliugit:


avtx-akonnur (commented):
@shuoliugit  Does this add the optional argument of `--delete_spot`? only for Azure spot GW creation?

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -432,6 +432,13 @@ func resourceAviatrixGateway() *schema.Resource {
 				Description:  "Price for spot instance. NOT supported for production deployment.",
 				RequiredWith: []string{"enable_spot_instance"},
 			},
+			"delete_spot": {
```
zhiqiangjin, line 435: "spot instance" support is only for internal test use. Should NOT be exposed in terraform state if not used.

zhiqiangjin (changes requested):

avtx-akonnur (commented):
Thanks for adding the delete spot argument, I will let @zhiqiangjin approve the PR.

zhiqiangjin (approved):

PR #1919, opened 2023-11-15
Correct GCP regions in example aviatrix_vpc resource

arvindsree:
Correct GCP regions in example aviatrix_vpc resource


arvindsree:
ignore

PR #1968, opened 2024-03-06
AVX-51277: Do not set bucket_name for config backup in Azure.

dnardo:
On the controller we have a bound validator that ensures that the `bucket_name` parameter is non-empty.  When we send a request to backup configuration for Azure, we use a set of different parameters and we do not set `bucket_name`. This causes validation to fail for Azure.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v3
       with:
-        go-version: ^1.21.7
+        go-version: ^1.21
       id: go
 
     - name: Check out code into the Go module directory
@@ -40,7 +40,7 @@ jobs:
       - name: Set up Go 1.x
         uses: actions/setup-go@v3
         with:
-          go-version: ^1.21.7
+          go-version: ^1.21
         id: go
 
       - name: Check out code into the Go module directory
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v3
       with:
-        go-version: ^1.14
+        go-version: ^1.21.7
       id: go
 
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v3
       with:
-        go-version: ^1.14
+        go-version: ^1.21.7
```
avx-rchung: should we pin to ^1.21 so we pick up patch releases automatically?

goaviatrix/controller.go
```diff
@@ -155,15 +155,22 @@ func (c *Client) GetSecurityGroupManagementStatus() (*SecurityGroupInfo, error)
...
-		"container_name": cloudnBackupConfiguration.BackupContainerName,
-		"region":         cloudnBackupConfiguration.BackupRegion,
+		"CID":        c.CID,
+		"action":     "enable_cloudn_backup_config",
+		"cloud_type": strconv.Itoa(cloudnBackupConfiguration.BackupCloudType),
+		"acct_name":  cloudnBackupConfiguration.BackupAccountName,
+		"region":     cloudnBackupConfiguration.BackupRegion,
+	}
+	// Azure has a set of different parameters that must be set.
+	if cloudnBackupConfiguration.BackupCloudType == Azure ||
```
avx-rchung: i think you can do:

`if goaviatrix.IsCloudType(cloudnBackupConfiguration.BackupCloudType, goaviatrix.AzureArmRelatedCloudTypes)`

.github/workflows/go.yml
dnardo: Ah thanks, I didn't realize that is what that did :-)

goaviatrix/controller.go
dnardo: Ah nice, I probably should have looked for something like that.  Thanks. Done.

avx-rchung (approved):

PR #1970, opened 2024-03-07
AVX-50899: fqdn_lan_interface is not needed for firenet fqdn association

shanxu2017:
### Issue
After create Firenet FQDN association, reapply same config, will cause destroy and reapply due to fqdn_lan_interface change
```
aviatrix_firewall_instance_association.arm_fqdn_gw_association[0] must be replaced
-/+ resource "aviatrix_firewall_instance_association" "arm_fqdn_gw_association" {
      ~ id              = "arm-transit-firenet-vnet:rg-av-arm-transit-firenet-vnet-817240:228a1824-5210-4398-97bb-f7bf5900da65~~arm-transit-firenet-gateway~~arm-fqdn-transit-firenet-gw" -> (known after apply)
      ~ lan_interface   = "av-nic-arm-fqdn-transit-firenet-gw_eth-fqdnlan0" -> "av-nic-arm-fqdn-transit-firenet-gw_eth1" # forces replacement
        # (5 unchanged attributes hidden)
    }
```

### RCA
`av-nic-arm-fqdn-transit-firenet-gw_eth1` is from avixtrix_gateway resource, however, this value is hard-coded in terraform code, not from controller API response.
`av-nic-arm-fqdn-transit-firenet-gw_eth-fqdnlan0` is from Firenet resouce, which is returned from controller API response.
These 2 values mismatch.

### Fix
For Firenet FQDN gateway, this `fqdn_lan_interface` is not used at all. Controller/copilot UI doesn't use it as well. Only controller backend use it internally.
Therefore, set it to empty value and remove from terraform logic.

### DIFF
--- a/aviatrix/data_source_aviatrix_gateway.go
+++ b/aviatrix/data_source_aviatrix_gateway.go
@@ -789,16 +789,13 @@ func dataSourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) err
 			d.Set("renegotiation_interval", -1)
 		}
 
-		fqdnGatewayLanInterface := fmt.Sprintf("av-nic-%s_eth1", gw.GwName)
 		fqdnLanCidr, ok := gw.ArmFqdnLanCidr[gw.GwName]
 		if ok && goaviatrix.IsCloudType(gw.CloudType, goaviatrix.AzureArmRelatedCloudTypes) {
-			d.Set("fqdn_lan_interface", fqdnGatewayLanInterface)
 			d.Set("fqdn_lan_cidr", fqdnLanCidr)
 		} else if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.GCPRelatedCloudTypes) {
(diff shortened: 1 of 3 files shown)


inc1t3Ful:
By removing the `fqdn_lan_interface` from outputting in the gateway resource/data source, there is no way to pass a `lan_interface` into the **aviatrix_firewall_instance_association** resource. Is that intended to be unnecessary now as an input for attaching the FQDN gateway to Firenet?

shanxu2017:
> By removing the `fqdn_lan_interface` from outputting in the gateway resource/data source, there is no way to pass a `lan_interface` into the **aviatrix_firewall_instance_association** resource. Is that intended to be unnecessary now as an input for attaching the FQDN gateway to Firenet?

You can pass `lan_interface` into the **aviatrix_firewall_instance_association** resource. I didn't remove it. I tested, pass in no problem. The code just ignore whatever passed in.

inc1t3Ful (approved):
LGTM

PR #1981, opened 2024-05-29
AVX-51501: remove the http_access configuration

jpeach:
The controller "http_access" API hasn't worked correctly since at least 7.0, so remove the implementation and mark it as deprecated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 3.1.5 (TBD, 2024)
+###
+- Supported Controller version: **UserConnect-TBD**
+- Supported Terraform version: **v1.x**
+
+### Deprecations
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 3.1.5 (TBD, 2024)
+###
+- Supported Controller version: **UserConnect-TBD**
```
dnardo, line 3: If this is 3.15 then this would be 7.1.d.

dnardo (approved):

CHANGELOG.md
jpeach: IIUC 7.1.d isn't a release, it's a branch?

jpeach: Assuming that we will update this with a concrete version number at release time :)

PR #1985, opened 2024-06-13
AVX-52626: Fix S2C so that it correctly updates remote_subnet_cidrs

dnardo:
There are a few things going on here:

1) At some point it looks like we made changes so that `remote_subnet_cidr` and
   `local_subnet_cidr` were copied into `cloud_subnet_cidr` when editing an sc2
   connection. We may have, at some point, been using the network type field
   to differentiate the two. Network type isn't used as a differentiator
   so this breaks when making changes to remote or local subnet cidr.

2) Additionally it seems that for edit, we use `remote_cidr`, not
   `remote_subnet_cidr` as we do for create.

This fix ensure that only `local_subnet_cidr` uses `cloud_subnet_cidr` and `remote_subnet_cidr` uses `remote_cidr` when updating.

Ideally we fix the API in the controller as this is a bit confusing on mutiple levels.  We should normalize the API for create and edit and most likely deprecate cloud_subnet_cidr or do away with it completely.

### DIFF
--- a/aviatrix/resource_aviatrix_site2cloud.go
+++ b/aviatrix/resource_aviatrix_site2cloud.go
@@ -1187,7 +1187,6 @@ func resourceAviatrixSite2CloudUpdate(d *schema.ResourceData, meta interface{}) 
 			GwName:                        d.Get("primary_cloud_gateway_name").(string),
 			VpcID:                         d.Get("vpc_id").(string),
 			ConnName:                      d.Get("connection_name").(string),
-			NetworkType:                   "3",
 			RemoteSourceRealCIDRs:         getCSVFromStringList(d, "remote_source_real_cidrs"),
 			RemoteSourceVirtualCIDRs:      getCSVFromStringList(d, "remote_source_virtual_cidrs"),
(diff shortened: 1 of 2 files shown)


aviatrix/resource_aviatrix_site2cloud.go
```diff
@@ -1051,7 +1051,6 @@ func resourceAviatrixSite2CloudUpdate(d *schema.ResourceData, meta interface{})
 		}
 		editSite2cloud.CloudSubnetCidr = d.Get("local_subnet_cidr").(string)
 		editSite2cloud.CloudSubnetVirtual = d.Get("local_subnet_virtual").(string)
-		editSite2cloud.NetworkType = "1"
```
jpeach, line 1054: We should also remove this field from the API structs. AFAICT, there is no such field in the site2conn API.

jpeach (changes requested):
Looks good, just one small change.

FWIW I didn't see the `network_type` parameter anywhere in the git history :shrug:

aviatrix/resource_aviatrix_site2cloud.go
dnardo: Done.

jpeach (approved):
LGTM

PR #1986, opened 2024-06-13
AVX-51456: DNAT should also include "eth0" when updating s2c.

dnardo:
This is a work-around for now, ideally we shouldn't need to send any interface information when using s2c connection in the SNAT/DNAT config.

Longer team goal with be to remove this workaround by adjusting the controller.

### DIFF
--- a/aviatrix/utils.go
+++ b/aviatrix/utils.go
@@ -221,12 +221,13 @@ func DiffSuppressFuncNatInterface(k, old, new string, d *schema.ResourceData) bo
 	connectionKey := strings.Replace(k, "interface", "connection", 1)
 	connection := d.Get(connectionKey).(string)
 
-	// Check if the number of snat or dnat policies has not changed so that
-	// interface can be set when a policy is added. Without this check, the
-	// value for interface will be suppressed and interface = "" will
-	// be passed to the API even if interface = "eth0" in the configuration.
-	// TODO(AVX-54006): This should be fixed on the controller side inthe future
-	// so that this check is no longer necessary.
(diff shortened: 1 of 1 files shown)


aviatrix/utils.go
```diff
@@ -224,7 +224,7 @@ func DiffSuppressFuncNatInterface(k, old, new string, d *schema.ResourceData) bo
 	// Check if the number of snat policies has not changed so that interface can be set when a policy is added.
 	// Without this check, the value for interface will be suppressed and interface = "" will
 	// be passed to the API even if interface = "eth0" in the configuration.
-	if !d.HasChange("snat_policy.#") && !(connection == "" || connection == "None") {
+	if !d.HasChange("snat_policy.#") && !d.HasChange("dnat_policy.#") && !(connection == "" || connection == "None") {
```
jpeach, line 231: Probably need to update the comment too since it seems specific to snat.

Would be nice to link to terraform docs about what `.#` means in a key. I couldn't find it and I assume it is special in some way?

jpeach (dismissed):

aviatrix/utils.go
dnardo: Good point. I updated the comment. Also reference the bug that was filed to remove this check entirely since we shouldn't need it.  As for the `.#` notation. I couldn't find a good reference for it in documentation but I see it used in the code: https://github.com/hashicorp/terraform/blob/9cf5350590ab25e8f2b51a69f9136f11d97369a0/helper/schema/resource_diff.go#L86

I can dig a little deeper though if you'd like.

jpeach (approved):

PR #1997, opened 2024-07-09
Fix import of edge device

CyrusJavan:
Will update with a Jira ticket when created.


rpanzer-aviatrix:
Maybe you can use AVX-54454?
This PR seems to be a superset of #1992

dnardo (changes requested):
Already had a fix for this, just needed a review.
https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/1992

dnardo:
> Already had a fix for this, just needed a review. #1992

Also neo is deprecated resource we don't need to fix that one. In fact it should be removed for 3.2

CyrusJavan:
@dnardo @rpanzer-aviatrix Thanks for the info, will close this in favor of the existing PR.

PR #2007, opened 2024-07-17
AVX-54880: Fix edge device response to match the correct field name.

dnardo:
This resulted in always incorrectly unmarshaling the `Network` field which made Terraform always think there was a diff.

### DIFF
--- a/goaviatrix/edge_neo_device_onbarding.go
+++ b/goaviatrix/edge_neo_device_onbarding.go
@@ -33,7 +33,7 @@ type EdgeNEODeviceResp struct {
 	DeviceId         string                  `json:"deviceId"`
 	SerialNumber     string                  `json:"serial"`
 	HardwareModel    string                  `json:"hardwareId"`
-	Network          []*EdgeNEODeviceNetwork `json:"interfaces"`
+	Network          []*EdgeNEODeviceNetwork `json:"network"`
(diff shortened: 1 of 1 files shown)


rpanzer-aviatrix (approved):
LGTM and the change makes sense.

I am just trying to associate this change with the corresponding API in cloudn, and I have to admit that I am not really able to find the corresponding API.
The closest I got was this:
https://github.com/AviatrixDev/cloudn/blob/da887fd5984eb6fb9c2b62685dcea053ecc50606/cloudx-local/edge_csp/api.py#L129
But that must be wrong, since the `hardware` property is also off from this definition.

dnardo:
> LGTM and the change makes sense.
> 
> I am just trying to associate this change with the corresponding API in cloudn, and I have to admit that I am not really able to find the corresponding API. The closest I got was this: [AviatrixDev/cloudn@`da887fd`/cloudx-local/edge_csp/api.py#L129](https://github.com/AviatrixDev/cloudn/blob/da887fd5984eb6fb9c2b62685dcea053ecc50606/cloudx-local/edge_csp/api.py#L129) But that must be wrong, since the `hardware` property is also off from this definition.

Because it actually comes from our SaaS controller, its separate repo.  controller is acting as a relay here.

PR #2008, opened 2024-07-17
AVX-54785: Don't use schema.TypeSet for DNS servers since order matters.

dnardo:
schema.TypeSet is unordered, but we need to preserve the DNS entries so that the first one in the list is primary and the second one in the list is secondary.


rpanzer-aviatrix (approved):
LGTM

PR #2016, opened 2024-08-12
AVX-49421: Cache the results of list accounts instead of making n calls

dnardo:
Currently we call list_accounts for every account, iterate through those results to find a specific account and return that reuslt. This can result in hundreds of calls if the customer has hundreds of accounts.

Instead, cache the results of the list_accounts call and use that to populate all accounts.

Tested with ~600 accounts and this brings us from ~35 minutes to about 30 to 15 seconds.

### DIFF
--- a/goaviatrix/account.go
+++ b/goaviatrix/account.go
@@ -133,8 +133,6 @@ func (c *Client) CreateOCIAccount(account *Account) error {
 	return c.PostFileAPI(params, files, DuplicateBasicCheck)
 }
 
-
-
 func (c *Client) ListAccounts() ([]Account, error) {
 	// If cached accounts are recent enough, return them
 	c.cacheMutex.Lock()
--- a/goaviatrix/account.go
+++ b/goaviatrix/account.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 )
@@ -133,18 +134,40 @@ func (c *Client) CreateOCIAccount(account *Account) error {
 	return c.PostFileAPI(params, files, DuplicateBasicCheck)
 }
 
-func (c *Client) GetAccount(account *Account) (*Account, error) {
+
+
+func (c *Client) ListAccounts() ([]Account, error) {
+	// If cached accounts are recent enough, return them
+	c.cacheMutex.Lock()
+	defer c.cacheMutex.Unlock()
+	if c.cachedAccounts != nil {
+		return c.cachedAccounts, nil
+	}
+
+	// Otherwise, fetch from the backend
+
 	form := map[string]string{
 		"CID":    c.CID,
 		"action": "list_accounts",
 	}
-
 	var resp AccountListResp
 	err := c.GetAPI(&resp, form["action"], form, BasicCheck)
 	if err != nil {
 		return nil, err
 	}
-	accList := resp.Results.AccountList
+	accounts := resp.Results.AccountList
+
+	// Cache the result
+	c.cachedAccounts = accounts
+
+	return accounts, nil
+}
+
+func (c *Client) GetAccount(account *Account) (*Account, error) {
+	accList, err := c.ListAccounts()
+	if err != nil {
+		return nil, err
+	}
 	for i := range accList {
 		if accList[i].AccountName == account.AccountName {
 			log.Infof("Found Aviatrix Account %s", account.AccountName)
--- a/goaviatrix/account.go
+++ b/goaviatrix/account.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-	"time"
 
 	log "github.com/sirupsen/logrus"
 )
--- a/goaviatrix/client.go
+++ b/goaviatrix/client.go
@@ -61,7 +61,6 @@ type Client struct {
 	baseURL          string
 	IgnoreTagsConfig *IgnoreTagsConfig
 	cachedAccounts   []Account
-	cacheTimestamp   time.Time
 	cacheMutex       sync.Mutex
 }
 
--- a/goaviatrix/client.go
+++ b/goaviatrix/client.go
@@ -16,6 +16,7 @@ import (
 	"os"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ajg/form"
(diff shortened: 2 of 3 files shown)


goaviatrix/client.go
```diff
@@ -59,6 +60,9 @@ type Client struct {
 	ControllerIP     string
 	baseURL          string
 	IgnoreTagsConfig *IgnoreTagsConfig
+	cachedAccounts   []Account
```
rpanzer-aviatrix, line 63: There might be extreme cases where a single terraform configuration has 2 configurations of the aviatrix provider.
Is this scenario completely unthinkable?
And if not, would this change have any impact on this?

rpanzer-aviatrix: Nvm, it's per client, so it should of course have 2 caches for 2 controllers.

goaviatrix/account.go
```diff
@@ -218,18 +219,38 @@ func (c *Client) CreateAWSSAccount(account *Account) error {
 	return c.PostFileAPI(params, files, DuplicateBasicCheck)
 }
 
-func (c *Client) GetAccount(account *Account) (*Account, error) {
+func (c *Client) ListAccounts() ([]Account, error) {
+	// If cached accounts are recent enough, return them
+	c.cacheMutex.Lock()
+	defer c.cacheMutex.Unlock()
+	if c.cachedAccounts != nil && time.Since(c.cacheTimestamp) < 10*time.Minute {
```
avx-rchung: if the cache is for one run of terraform it seems like we don't need to time-bound it -- it seems reasonable to say that each run of terraform should have a consistent view of accounts?

goaviatrix/account.go
```diff
@@ -218,18 +219,38 @@ func (c *Client) CreateAWSSAccount(account *Account) error {
 	return c.PostFileAPI(params, files, DuplicateBasicCheck)
 }
 
-func (c *Client) GetAccount(account *Account) (*Account, error) {
+func (c *Client) ListAccounts() ([]Account, error) {
+	// If cached accounts are recent enough, return them
+	c.cacheMutex.Lock()
```
rpanzer-aviatrix, line 138: Do we need to invalidate the cache when an account is added, updated or delete?

goaviatrix/account.go
dnardo: Correct.  We can make it non-expiring, I just threw something in there to prove it out.
Each command should be stateless so the cache will be rebuilt for each command, (plan, apply etc...)  I only see this is way to force one call to happen per command.

goaviatrix/account.go
dnardo: Each command is stateless, so for instance, if I add a new account to my TF config, it should only see that diff.  This function should only affect read back.  I assume in this case you mean account in the terraform config right?

Lets say I have a bunch of accounts already defined and I add a new one, I should only see this:

```
aviatrix_account.aws_account[410]: Refreshing state... [id=aws-411]
aviatrix_account.aws_account[4]: Refreshing state... [id=aws-5]
aviatrix_account.aws_account[324]: Refreshing state... [id=aws-325]
aviatrix_account.aws_account[46]: Refreshing state... [id=aws-47]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

  # aviatrix_account.azure_account-2 will be created
  + resource "aviatrix_account" "azure_account-2" {
      + account_name            = "azure-new-account"
      + arm_application_id      = (sensitive value)
      + arm_application_key     = (sensitive value)
      + arm_directory_id        = (sensitive value)
      + arm_subscription_id     = (sensitive value)
      + audit_account           = false
      + aws_ca_cert_path        = (known after apply)
      + aws_role_app            = (known after apply)
      + aws_role_ec2            = (known after apply)
      + awschina_role_app       = (known after apply)
      + awschina_role_ec2       = (known after apply)
      + awsgov_role_app         = (known after apply)
      + awsgov_role_ec2         = (known after apply)
      + awss_cap_cert_key_path  = (known after apply)
      + awss_cap_cert_path      = (known after apply)
      + awsts_cap_cert_key_path = (known after apply)
      + awsts_cap_cert_path     = (known after apply)
      + cloud_type              = 8
      + id                      = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.
````

Same for if I delete one.

Does that answer the question?

Comment (rpanzer-aviatrix) [reply goaviatrix/account.go:138]: I have to admit I didn't look at the Terraform's component model yet.
I was thinking in terms of 
1. Terraform gets the AWS account to reconcile resource A
2. Terraform creates updates the AWS account
3. Terraform stores the AWS account in the output and finds the original account in the cache.

But yeah, I am not really familiar with the component model, so I am probably wrong.

File: goaviatrix/account.go (hunk 2 above)
Comment (dnardo) [reply goaviatrix/account.go]: Removed expiry

**Review by avx-rchung — COMMENTED** [review]
is there a reasonable UT we can add for this, maybe in aviatrix/resource_aviatrix_account_user_test.go ?

**Comment by dnardo** [comment]
> is there a reasonable UT we can add for this, maybe in aviatrix/resource_aviatrix_account_user_test.go ?
If we want a unit test for this I don't think that is the place to do it.  Those are acceptance tests, and I'm not even clear on when they run or what triggers them.  From what I can see there isn't much actual unit tests. I'm happy to add something, but we should at least agree on what that test should test in this particular case.

**Comment by avx-rchung** [comment]
oh, i was thinking about something simple, like two calls to GetAccount() do not generate multiple API calls, but both get the "right" results

**Comment by dnardo** [comment]
> oh, i was thinking about something simple, like two calls to GetAccount() do not generate multiple API calls, but both get the "right" results

I'm not proud of it, but I've added a test that should verify this.  I had to mock RoundTripper unfortunately due to the way the client and client interfaces are currently defined.  Something else we can probably make a bit better. 

```
❯ go test -run TestListAccountsCallCount -v
=== RUN   TestListAccountsCallCount
INFO[0000] Found Aviatrix Account test-account
INFO[0000] Found Aviatrix Account test-account
--- PASS: TestListAccountsCallCount (0.00s)
PASS
ok      github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix    0.003s
```

goaviatrix/account.go
dnardo: Its a good point, let me convince myself that can't happen.

avx-rchung (dismissed):

avx-rchung (approved):

PR #2021, opened 2024-08-15
AVX-47161 - Add Edge Proxy Profile Support

avx-vjain:


goaviatrix/edge_neo_device_onbarding.go
```diff
@@ -19,12 +19,12 @@ type EdgeNEODevice struct {
 }
 
 type EdgeNEODeviceNetwork struct {
```
dnardo, line 21: I note its not your PR, but we can fix the name of this file.  First off its say `onbarding` second we don't really want to use the "neo" term.  Can we call this `edge_platform_device_onboarding.go` ? or something like that?

goaviatrix/edge_neo_proxy_profile.go
```diff
@@ -0,0 +1,88 @@
+package goaviatrix
```
dnardo: Since this a new file, lets not use the term `neo` lets call this edge_platform_proxy_profile (or something like that)

dnardo (commented):
A couple of initial comments, still needs another pass.  But we need to update the docs also in this PR.  Can you take a look at the existing documentation and add this resource and update the related docs?

aviatrix/resource_aviatrix_edge_proxy_profile_config.go
```diff
@@ -0,0 +1,120 @@
...
+
+func resourceAviatrixEdgeProxyProfileConfigCreate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*goaviatrix.Client)
+
+	proxy := marshalEdgeProxyProfileConfigInput(d)
+	if err := client.CreateEdgeProxyProfile(context.Background(), proxy); err != nil {
+		return fmt.Errorf("could not config proxy: %v", err)
+	}
+
+	d.SetId(strings.Replace(client.ControllerIP, ".", "-", -1))
```
dnardo: Is the controller ip unique enough for proxy profiles?  I would have expected this to be something like proxy_profile_id or a combo of account_name and profile_name (if that is unique).  Either way this doesn't seem unique enough to me.

aviatrix/resource_aviatrix_edge_proxy_profile_config.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		return fmt.Errorf("couldn't get proxy configuration: %s", err)
+	}
+
+	d.Set("ip_address", proxy.IPAddress)
+	d.Set("port", proxy.Port)
+	d.Set("proxy_profile_id", proxy.ProxyID)
+	d.Set("name", proxy.Name)
+	if proxy.CaCert != nil && *proxy.CaCert != "" {
+		d.Set("caCert", proxy.CaCert)
```
dnardo: Should this be
d.set("ca_certificate", proxy.CaCert)

aviatrix/resource_aviatrix_edge_proxy_profile_config.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	d.SetId(strings.Replace(client.ControllerIP, ".", "-", -1))
```
dnardo: Again I don't think this is the right thing to use here.

aviatrix/resource_aviatrix_edge_neo_device_onboarding.go
```diff
@@ -123,11 +123,11 @@ func marshalEdgeNEODeviceOnboardingInput(d *schema.ResourceData) *goaviatrix.Edg
 		network1 := network0.(map[string]interface{})
 
 		network2 := &goaviatrix.EdgeNEODeviceNetwork{
-			InterfaceName: network1["interface_name"].(string),
-			EnableDhcp:    network1["enable_dhcp"].(bool),
-			GatewayIp:     network1["gateway_ip"].(string),
-			Ipv4Cidr:      network1["ipv4_cidr"].(string),
-			ProxyServerIp: network1["proxy_server_ip"].(string),
+			InterfaceName:  network1["interface_name"].(string),
+			EnableDhcp:     network1["enable_dhcp"].(bool),
```
dnardo, line 127: We still need to update the markdown docs for this resource. Please take a look at the Docs folder.

dnardo (changes requested):

aviatrix/resource_aviatrix_edge_neo_device_onboarding.go
avx-vjain: Sure. Will update the docs in the next PR

aviatrix/resource_aviatrix_edge_proxy_profile_config.go
avx-vjain: Using proxyID

aviatrix/resource_aviatrix_edge_neo_device_onboarding.go
avx-vjain: Will fix the docs in the next PR

dnardo (approved):

PR #2031, opened 2024-09-02
AVX-56220: Fix various issues with the account creation logic.

dnardo:
- Make sure we lock the cache when iterating through the accounts.
- Return a copy of the account not a pointer.                                                                                                                                                                                       
- Don't defer read in the creation of an account before we perform a  `SetId`.
- Perform the invalidate explicitly after success.                                                                                                                                                                                                                                                                                                  
- Always perform a read back on success.                                                                                                                                                                                                                                                                                                            
- Fix other related files as a result of returning a copy of an account.                                                                                                                                                                                                                                                                            
- Removes now unused `resourceAviatrixAccountReadIfRequired` function.

### DIFF
--- a/aviatrix/data_source_aviatrix_vpc.go
+++ b/aviatrix/data_source_aviatrix_vpc.go
@@ -240,10 +240,7 @@ func dataSourceAviatrixVpcRead(d *schema.ResourceData, meta interface{}) error {
 			}
 		}
 
-		var subscriptionId string
-		if acc != nil {
-			subscriptionId = acc.ArmSubscriptionId
-		}
+		subscriptionId := acc.ArmSubscriptionId
 
 		var resourceGroup string
 		if vC.ResourceGroup != "" {
--- a/aviatrix/resource_aviatrix_account.go
+++ b/aviatrix/resource_aviatrix_account.go
@@ -435,7 +435,6 @@ func resourceAviatrixAccount() *schema.Resource {
 
 func resourceAviatrixAccountCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*goaviatrix.Client)
-	defer client.InvalidateCache()
 	account := &goaviatrix.Account{
 		AccountName:                           d.Get("account_name").(string),
 		CloudType:                             d.Get("cloud_type").(int),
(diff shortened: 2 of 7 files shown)


aviatrix/data_source_aviatrix_vpc.go
```diff
@@ -240,10 +240,7 @@ func dataSourceAviatrixVpcRead(d *schema.ResourceData, meta interface{}) error {
 			}
 		}
 
-		var subscriptionId string
```
tommy-carver, line 243: nit: For safety sake, can we throw a `if acc == nil` case before all of this and error accordingly? The old code didn't make sense, but I think we should throw some error if the is nil but we didn't get an error

aviatrix/resource_aviatrix_account.go
```diff
@@ -733,13 +733,11 @@ func resourceAviatrixAccountCreate(ctx context.Context, d *schema.ResourceData,
 		return diag.Errorf("failed to create Aviatrix Account: %s", err)
```
tommy-carver: Can we simplify this? Its the same error returned twice. Can we just check if the err is nil before doing casting? Then return the casted error if the error matches? Just confusing logic.

Something like `if err != nil { if _, ok := err.(...) {`

aviatrix/resource_aviatrix_account.go
```diff
@@ -780,76 +778,75 @@ func resourceAviatrixAccountRead(ctx context.Context, d *schema.ResourceData, me
 		return diag.Errorf("aviatrix Account: %s", err)
```
tommy-carver, line 765: Same nit as above regarding acc == nil, but up to you on preference

goaviatrix/client.go
```diff
@@ -46,7 +46,7 @@ type APIRequest struct {
 //go:generate moq -rm -out client_mock.go . ClientInterface
 type ClientInterface interface {
 	DeleteAccount(account *Account) error
-	GetAccount(account *Account) (*Account, error)
```
tommy-carver, line 49: Whats the reason for this? Seems logical to keep pointer for consistency?

goaviatrix/account.go
```diff
@@ -158,27 +157,26 @@ func (c *Client) ListAccounts() ([]Account, error) {
...
 	for i := range accList {
 		if accList[i].AccountName == account.AccountName {
 			log.Infof("Found Aviatrix Account %s", account.AccountName)
-			return &accList[i], nil
+			return accList[i], nil
 		}
 	}
 	log.Errorf("Couldn't find Aviatrix account %s", account.AccountName)
-	return nil, ErrNotFound
+	return Account{}, ErrNotFound
```
tommy-carver, line 184: Nit: I prefer nil because of predictability, since default values change and easier to handle, but up you.

goaviatrix/client.go
dnardo: This is the point of the change.  We return an object now instead of a pointer.  If we pass a pointer back from the cache that could cause issues, so we make a copy instead.

dnardo: Added comment to GetAccount

aviatrix/resource_aviatrix_account.go
dnardo: Fixed, but also realized something else in the process.  I need to test this change a bit more thoroughly.

tommy-carver (approved):
LGTM, thanks!

PR #2014, opened 2024-07-29
AVX-52621 Updating the edge transit resources

sairane10:
- Update the transit resource to support edge API's
- The following api's are modified by edge
`
create_multicloud_primary_gateway
create_multicloud_ha_gateway
attach_spoke_to_transit_gateway
delete_transit_gateway

`

PR #2032, opened 2024-09-03
AVX-52621 Updating the interface ordering and documentation for transit resource

sairane10:
This PR is a follow up of https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2014
- Adding EAT resource document changes
- Updating the interface ordering in transit resource
- Updating the interface names
- Removing device_id from transit resource state

### DIFF
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -752,26 +752,6 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 					},
 				},
 			},
-			"eip_map": {
-				Type: schema.TypeMap,
-				Elem: &schema.Schema{
-					Type: schema.TypeList,
-					Elem: &schema.Resource{
-						Schema: map[string]*schema.Schema{
-							"private_ip": {
-								Type:     schema.TypeString,
-								Required: true,
-							},
-							"public_ip": {
-								Type:     schema.TypeString,
-								Required: true,
-							},
-						},
-					},
-				},
-				Optional:    true,
-				Description: "Map of interfaces containing the private IP to public IP EIP assignment",
-			},
 			"peer_backup_port": {
 				Type:        schema.TypeString,
 				Optional:    true,
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -752,6 +752,26 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 					},
 				},
 			},
+			"eip_map": {
+				Type: schema.TypeMap,
+				Elem: &schema.Schema{
+					Type: schema.TypeList,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"private_ip": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+							"public_ip": {
+								Type:     schema.TypeString,
+								Required: true,
+							},
+						},
+					},
+				},
+				Optional:    true,
+				Description: "Map of interfaces containing the private IP to public IP EIP assignment",
+			},
 			"peer_backup_port": {
 				Type:        schema.TypeString,
 				Optional:    true,
@@ -798,7 +818,7 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 			if val, exists := ifaceInfo["name"]; exists && val != nil {
 				ifaceName, ok = val.(string)
 				if !ok {
-					return fmt.Errorf("ifname is not a string")
+					return fmt.Errorf("name is not a string")
 				}
 			}
 			// Check and set 'type'
@@ -819,7 +839,7 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 			if val, exists := ifaceInfo["ip_address"]; exists && val != nil {
 				ifaceIP, ok = val.(string)
 				if !ok {
-					return fmt.Errorf("ipaddr is not a string")
+					return fmt.Errorf("ip address is not a string")
 				}
 			}
 			// Check and set 'public_ip'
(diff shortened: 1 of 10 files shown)


aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1871,42 +1871,70 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
...
 			}
 			if err = d.Set("interfaces", interfaces); err != nil {
 				return fmt.Errorf("could not set interfaces into state: %v", err)
 			}
 		}
-		// set the interface map
-		d.Set("interface_map", gw.InterfaceMapping)
-		if gw.PeerBackupPort != "" && gw.ConnectionType != "" {
-			d.Set("peer_backup_port", gw.PeerBackupPort)
-			d.Set("connection_type", gw.ConnectionType)
```
inc1t3Ful, line 1908: trying to understand, here you just no longer set the interface map?

goaviatrix/transit_ha_gateway.go
```diff
@@ -12,18 +12,18 @@ type TransitHaGateway struct {
...
+	BgpLanVpcId           string `json:"bgp_lan_vpc,omitempty"`
+	BgpLanSubnet          string `json:"bgp_lan_specify_subnet,omitempty"`
 	Eip                   string `json:"eip,omitempty"`
 	InsaneMode            string `json:"insane_mode"`
-	TagList               string `json:"tag_string"`
-	TagJson               string `json:"tag_json"`
-	AutoGenHaGwName       string `json:"autogen_hagw_name"`
+	TagList               string `json:"tag_string,omitempty"`
+	TagJson               string `json:"tag_json,omitempty"`
+	AutoGenHaGwName       string `json:"autogen_hagw_name,omitempty"`
```
inc1t3Ful: What is the reason we are using `omitempty` here so generously? Was its exclusion causing an issue before?

inc1t3Ful (changes requested):
Please also update the transit_gateway markdown, as you have changed the attribute names (`name` and `ip_address`)

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Interface map is being set statically in the code and so I am not setting the EAT state here.

goaviatrix/transit_ha_gateway.go
sairane10: Since the same transit ha_gw is being used for EAT to set the state, some of these values are not applicable for it. So added the omitempty.

sairane10: Updated the transit_ha_gateway.go

docs/resources/aviatrix_transit_gateway.md
```diff
@@ -346,34 +346,34 @@ resource "aviatrix_transit_gateway" "test-edge-transit-1-hagw" {
...
-        ifname                      = "eth4"
-        ipaddr                     = "192.168.23.12/24"
-        type                        = "WAN"
+        gateway_ip   = "192.168.23.1"
+        name         = "eth4"
+        ip_address   = "192.168.23.12/24"
+        type         = "WAN"
     }
 }
 ```
```
inc1t3Ful, line 379: After looking further into markdown, we are missing the actual attributes in description, that are supposed to be nested under the supposed `interfaces` map description.

We have `interfaces` listed in the Arguments Reference, but not the nested attributes such as `gateway_ip`, `name`, `ip_address`, `type` etc. (Any other ones I'm missing?)

inc1t3Ful (changes requested):

docs/resources/aviatrix_transit_gateway.md
sairane10: Updated the docs to have interface subparameters.

dnardo (changes requested):
Can we split out the documentation changes from the code changes.  They seem completely separate things and the PR title is misleading.

sairane10:
@dnardo I have changed the title to reflect changes getting added in this PR. It contains the feedback changes suggested by Anthony along with interface ordering.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -794,11 +794,11 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
...
 			var ifaceDHCP bool
 			var secondaryCIDRs []string
-			// Check and set 'ifname'
-			if val, exists := ifaceInfo["ifname"]; exists && val != nil {
+			// Check and set 'interface name'
+			if val, exists := ifaceInfo["name"]; exists && val != nil {
 				ifaceName, ok = val.(string)
 				if !ok {
-					return fmt.Errorf("ifname is not a string")
+					return fmt.Errorf("name is not a string")
```
dnardo: Suggest something like "interface name is not a string" since "name" may be unclear.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1871,42 +1871,70 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
...
+					secondaryCIDRs := make([]string, 0)
+					for _, cidr := range intf.SecondaryCIDRs {
+						if cidr != "" {
+							secondaryCIDRs = append(secondaryCIDRs, cidr)
+						}
+					}
+					interfaceDict["secondary_private_cidr_list"] = secondaryCIDRs
 				}
 				interfaces = append(interfaces, interfaceDict)
+				log.Printf("[TRACE] Interface Dictionary %s: %#v", intf.Name, interfaceDict)
```
dnardo: Do you mean to leave this in?

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1871,42 +1871,70 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
 
 	// edge cloud type
 	if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.EdgeRelatedCloudTypes) {
-		d.Set("device_id", gw.DeviceID)
```
dnardo, line 1874: This change isn't in the PR description, can you explain why this is safe to remove?

dnardo (dismissed):

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Updated the interface name

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Removed the trace log. Had this check in to test secondary_cidr.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: the device_id gets populated after the gateway is up which takes about 20 mins. If the user tries to apply the resource again then this field will try to update even when the gateway is in the middle of being deployed.

dnardo: Make sense, can you update the PR description please with that change.

dnardo (approved):

inc1t3Ful (approved):
LGTM

PR #1996, opened 2024-06-28
AVX-36418 Deprecating bandwidth attribute in interface

sairane10:
Fix - [AVX-36148](https://aviatrix.atlassian.net/browse/AVX-36418)

Marking the bandwidth attribute as deprecated in interfaces. This feature was never fully implemented so it will be removed in the future releases.

PR #2040, opened 2024-09-11
AVX-55729 Removing the bandwidth attribute and updating interface config

sairane10:
- Updating the interface config for vrrp_state, virtual_ip and subinterfaces in edge_equinix resource 
- Removing the bandwidth attribute implementation from edge resources.
- Updating the deprecated attribute in the feature-changelist 

https://aviatrix.atlassian.net/browse/AVX-55729

### DIFF
--- a/aviatrix/resource_aviatrix_edge_equinix.go
+++ b/aviatrix/resource_aviatrix_edge_equinix.go
@@ -371,7 +371,6 @@ func marshalEdgeEquinixInput(d *schema.ResourceData) *goaviatrix.EdgeEquinix {
 		interface2 := &goaviatrix.EdgeEquinixInterface{
 			IfName:       interface1["name"].(string),
 			Type:         interface1["type"].(string),
-			Bandwidth:    interface1["bandwidth"].(int),
 			PublicIp:     interface1["wan_public_ip"].(string),
 			Tag:          interface1["tag"].(string),
 			Dhcp:         interface1["enable_dhcp"].(bool),
@@ -675,7 +674,6 @@ func resourceAviatrixEdgeEquinixRead(ctx context.Context, d *schema.ResourceData
 		interface1 := make(map[string]interface{})
 		interface1["name"] = interface0.IfName
 		interface1["type"] = interface0.Type
-		interface1["bandwidth"] = interface0.Bandwidth
 		interface1["wan_public_ip"] = interface0.PublicIp
 		interface1["tag"] = interface0.Tag
 		interface1["enable_dhcp"] = interface0.Dhcp
--- a/aviatrix/resource_aviatrix_edge_equinix.go
+++ b/aviatrix/resource_aviatrix_edge_equinix.go
@@ -211,6 +211,12 @@ func resourceAviatrixEdgeEquinix() *schema.Resource {
 							Description:  "Interface type.",
 							ValidateFunc: validation.StringInSlice([]string{"WAN", "LAN", "MANAGEMENT"}, false),
 						},
+						"bandwidth": {
+							Type:        schema.TypeInt,
+							Optional:    true,
+							Description: "The rate of data can be moved through the interface, requires an integer value. Unit is in Mb/s.",
+							Deprecated:  "Bandwidth will be removed in a future release.",
+						},
 						"enable_dhcp": {
 							Type:        schema.TypeBool,
 							Optional:    true,
@@ -365,6 +371,7 @@ func marshalEdgeEquinixInput(d *schema.ResourceData) *goaviatrix.EdgeEquinix {
 		interface2 := &goaviatrix.EdgeEquinixInterface{
 			IfName:       interface1["name"].(string),
 			Type:         interface1["type"].(string),
+			Bandwidth:    interface1["bandwidth"].(int),
 			PublicIp:     interface1["wan_public_ip"].(string),
 			Tag:          interface1["tag"].(string),
 			Dhcp:         interface1["enable_dhcp"].(bool),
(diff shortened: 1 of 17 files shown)


aviatrix/resource_aviatrix_edge_equinix.go
```diff
@@ -371,16 +365,18 @@ func marshalEdgeEquinixInput(d *schema.ResourceData) *goaviatrix.EdgeEquinix {
...
 			GatewayIp:    interface1["gateway_ip"].(string),
 			DnsPrimary:   interface1["dns_server_ip"].(string),
 			DnsSecondary: interface1["secondary_dns_server_ip"].(string),
-			VrrpState:    interface1["enable_vrrp"].(bool),
-			VirtualIp:    interface1["vrrp_virtual_ip"].(string),
+		}
+
+		if interface1["type"].(string) == "LAN" && interface1["enable_vrrp"].(bool) {
+			interface2.VrrpState = interface1["enable_vrrp"].(bool)
+			interface2.VirtualIp = interface1["vrrp_virtual_ip"].(string)
```
inc1t3Ful, line 386: Could you explain what this change is?

inc1t3Ful (changes requested):
Please also update the `docs/guides/feature-changelist-v3.md` with note of deprecation for this attribute

aviatrix/resource_aviatrix_edge_equinix.go
sairane10: vrrp_state and the virtual_ip is only applicable to LAN interfaces. This check was added to the api a month back but never updated on the equinix tf.

sairane10: Here's the backend change:
https://github.com/AviatrixDev/cloudn/blob/4ec0266ee378f86b7f19ec63ad60641daefc3c9c/cloudx-local/avx/edge/interface.py#L253

jpeach: Seems worth having a code comment to explain the situation to future maintainers.

sairane10: Added the comment.

inc1t3Ful:
Please also update the `docs/guides/feature-changelist-v3.md` with note of deprecation for this attribute. That doc tracks impacting changes to resources/ environment such as deprecations

docs/guides/feature-changelist-v3.md
```diff
@@ -87,6 +87,12 @@ The following logging resources are removed:
 |(deprecated) | controller_config | http_access | **Yes**; This configuration value no longer has any effect. It will be removed from the Aviatrix provider in the 3.2.0 release |
 
 
+## R3.2.0 (UserConnect-7.2)**
+### Attribute Deprecations
+| Diff | Resource | Attribute | Action Required? |
+|:----:|:--------:|:---------:|:----------------:|
+|(deprecated) | edge_equinix, edge_equinix_ha, edge_csp, edge_csp_ha, edge_neo, edge_neo_ha, edge_platform, edge_platform_ha, edge_zededa, edge_zededa_ha | bandwidth | **Yes**; This configuration value no longer has any effect. It will be removed from the Aviatrix provider in the 3.2.0 release |
```
inc1t3Ful, line 94: I see there are multiple resources listed here, but only `edge_equinix` and `edge_equinix_ha` are actually changed in the PR. 

Any particular reasons for these other resources listed here?

inc1t3Ful (changes requested):

docs/guides/feature-changelist-v3.md
sairane10: Updated other resources as well.

inc1t3Ful (changes requested):
Now that all the resources are updated, please make sure to update their respective docs as well to note the deprecation (remove the attribute)

sairane10:
@inc1t3Ful The docs are already updated for each of these resources along with the change log. It was done in my older PR.
https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/1996/files

jpeach (approved):

aviatrix/resource_aviatrix_edge_equinix.go
```diff
@@ -371,16 +371,19 @@ func marshalEdgeEquinixInput(d *schema.ResourceData) *goaviatrix.EdgeEquinix {
...
 			DnsPrimary:   interface1["dns_server_ip"].(string),
 			DnsSecondary: interface1["secondary_dns_server_ip"].(string),
-			VrrpState:    interface1["enable_vrrp"].(bool),
-			VirtualIp:    interface1["vrrp_virtual_ip"].(string),
+		}
+
+		// vrrp_state and virtual_ip are only applicable for LAN interfaces
+		if interface1["type"].(string) == "LAN" && interface1["enable_vrrp"].(bool) {
+			interface2.VrrpState = interface1["enable_vrrp"].(bool)
+			interface2.VirtualIp = interface1["vrrp_virtual_ip"].(string)
```
rpanzer-aviatrix, line 386: Would it make sense to have a validation error if the interface type is not `"LAN"` and `vrrp_virtual_ip` is set?
Or if we want to remove any validation in Terraform, then rather forward it to the controller and have that reject it?
To me, it would be surprising if a property is silently dropped.

rpanzer-aviatrix (approved):
Approving the change, left a minor comment.
Out of scope of this PR, I think it would be great though if the acceptance would work.
Right now they fail for me.

aviatrix/resource_aviatrix_edge_equinix.go
sairane10: Currently we have the check in the controller to throw an error if vrrp_state or virtual_ip is set for interfaces other than LAN so I didn't include it here.

inc1t3Ful (approved):

PR #2052, opened 2024-09-18
AVX-47161: Fix proxy profile update

avx-vjain:
- Removed ForceNew from all the fields that were optional
- Added ProxyProfile update functionality

### DIFF
--- a/aviatrix/resource_aviatrix_edge_proxy_profile_config.go
+++ b/aviatrix/resource_aviatrix_edge_proxy_profile_config.go
@@ -121,8 +121,8 @@ func resourceAviatrixEdgeProxyProfileConfigUpdate(d *schema.ResourceData, meta i
 			AccountName: proxy.AccountName,
 			Name:        proxy.Name,
 			Address:     existingProxy.IPAddress,
-			Port:        int(existingProxy.Port),
-			CACert:      *existingProxy.CaCert,
+			Port:        existingProxy.Port,
+			CACert:      existingProxy.CaCert,
 		},
 		ProxyID: existingProxy.ProxyID,
 	}
@@ -161,9 +161,7 @@ func resourceAviatrixEdgeProxyProfileConfigRead(d *schema.ResourceData, meta int
 	d.Set("port", proxy.Port)
 	d.Set("proxy_profile_id", proxy.ProxyID)
 	d.Set("name", proxy.Name)
-	if proxy.CaCert != nil && *proxy.CaCert != "" {
-		d.Set("ca_certificate", proxy.CaCert)
-	}
+	d.Set("ca_certificate", proxy.CaCert)
 
 	d.SetId(proxy.ProxyID)
 	return nil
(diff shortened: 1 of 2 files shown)


aviatrix/resource_aviatrix_edge_proxy_profile_config.go
```diff
@@ -8,9 +8,20 @@ import (
...
 )
 
+type proxyProfile struct {
+	Action      string  `json:"action"`
+	CID         string  `json:"CID"`
+	AccountName string  `json:"account_name"`
+	Name        string  `json:"proxy_name"`
+	Address     *string `json:"address"`
+	Port        *int    `json:"port"`
+	CACert      *string `json:"ca_cert"`
```
dnardo, line 18: Why do these need to be pointers? Are we trying to determine  Zero Values and Unset Fields ?

aviatrix/resource_aviatrix_edge_proxy_profile_config.go
```diff
@@ -70,6 +103,48 @@ func resourceAviatrixEdgeProxyProfileConfigCreate(d *schema.ResourceData, meta i
...
+		}
+		return fmt.Errorf("couldn't get proxy configuration: %s", err)
+	}
+
+	modifiedProxy := &goaviatrix.EdgePlatformProxyProfileUpdate{
+		EdgePlatformProxyProfile: goaviatrix.EdgePlatformProxyProfile{
+			AccountName: proxy.AccountName,
+			Name:        proxy.Name,
+			Address:     existingProxy.IPAddress,
+			Port:        int(existingProxy.Port),
```
dnardo: Why don't we just avoid the conversion here and fix the struct in edge_platform_proxy_profile.go to be int. We know port won't be above 65535 and we know they will not have more than 65535 proxy profiles.  Seems simpler to me.

aviatrix/resource_aviatrix_edge_proxy_profile_config.go
avx-vjain: Unset fields

goaviatrix/edge_platform_proxy_profile.go
```diff
@@ -15,16 +15,21 @@ type EdgePlatformProxyProfile struct {
...
-	ProxyProfileCount int64      `json:"deviceCount"`
-	CreatedAt         time.Time  `json:"createdAt"`
-	UpdatedAt         time.Time  `json:"updatedAt"`
-	CaCert            *string    `json:"caCert,omitempty"`
-	ExpiredAt         *time.Time `json:"expiredAt,omitempty"`
+	ProxyID           string    `json:"proxyID"`
+	Name              string    `json:"name"`
+	IPAddress         string    `json:"address"`
+	Port              int       `json:"port"`
+	ProxyProfileCount int64     `json:"deviceCount"`
```
dnardo: I'd make this an int also just so there is no conversion down the road.

dnardo (dismissed):

dnardo (approved):

PR #2055, opened 2024-09-20
AVX-56803 Adding Interface Mapping to transit gateways

sairane10:
- Updating the edge resources with interface mapping

https://aviatrix.atlassian.net/browse/AVX-56803

### DIFF
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -57,7 +57,7 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 			},
 			"vpc_id": {
 				Type:             schema.TypeString,
-				Required:         true,
+				Optional:         true,
 				ForceNew:         true,
 				Description:      "VPC-ID/VNet-Name of cloud provider.",
 				DiffSuppressFunc: DiffSuppressFuncGatewayVpcId,
@@ -753,6 +753,20 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 					},
 				},
 			},
+			"interface_mapping": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "Mapping of interface names to types and indices.",
+				Elem: &schema.Schema{
+					Type: schema.TypeList,
+					Elem: &schema.Schema{
+						Type:        schema.TypeString,
+						Required:    true,
+						Description: "Interface type (e.g., 'wan', 'mgmt') and index.",
+					},
+					MaxItems: 2,
+				},
+			},
 			"peer_backup_port": {
 				Type:        schema.TypeString,
 				Optional:    true,
@@ -778,7 +792,6 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 			CloudType:   d.Get("cloud_type").(int),
 			AccountName: d.Get("account_name").(string),
 			GwName:      d.Get("gw_name").(string),
-			VpcID:       d.Get("vpc_id").(string),
 			VpcRegion:   d.Get("vpc_reg").(string),
 			VpcSize:     d.Get("gw_size").(string),
 			DeviceID:    d.Get("device_id").(string),
@@ -871,14 +884,28 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 		gateway.Interfaces = b64.StdEncoding.EncodeToString(interfaceList)
 
 		if goaviatrix.IsCloudType(cloudType, goaviatrix.EDGENEO) {
-			// set the static interface mapping for AEP
-			interfaceMapping := map[string][]string{
-				"eth0": {"wan", "0"},
-				"eth1": {"wan", "1"},
-				"eth2": {"wan", "2"},
-				"eth3": {"mgmt", "0"},
-				"eth4": {"wan", "3"},
+			// interfaceMapping := map[string][]string{
+			// 	"eth0": {"wan", "0"},
+			// 	"eth1": {"wan", "1"},
+			// 	"eth2": {"wan", "2"},
+			// 	"eth3": {"mgmt", "0"},
+			// 	"eth4": {"wan", "3"},
+			// }
+			interfaceMappingInput := d.Get("interface_mapping").(map[string]interface{})
+			interfaceMapping := make(map[string][]string)
+
+			for key, value := range interfaceMappingInput {
+				// Each value is a list of two strings (interface type and index)
+				mappingList := value.([]interface{})
+				if len(mappingList) == 2 {
+					interfaceType := mappingList[0].(string)
+					interfaceIndex := mappingList[1].(string)
+					interfaceMapping[key] = []string{interfaceType, interfaceIndex}
+				} else {
+					return fmt.Errorf("invalid interface mapping for %s", key)
+				}
 			}
+
 			// Convert interfaceMapping to JSON byte slice
 			interfaceMappingJSON, err := json.Marshal(interfaceMapping)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


docs/resources/aviatrix_transit_gateway.md
```diff
@@ -407,6 +432,10 @@ The following arguments are supported:
   * `public_ip` - (Optional) The public IP address associated with this interface.
   * `dhcp` - (Optional) Whether DHCP is enabled on this interface. Set the value to true or false. Applicable to only 'MANAGEMENT' type interface.
   * `secondary_private_cidr_list` - (Optional) A list of secondary private CIDR blocks associated with this interface.
+* `interface_mapping` - (Optional) A list of interface names mapped to interface types and indices. Required and valid only for edge transit gateways (AEP). Each interface has the following attributes:
+  * `name` - (Required) Interface name e.g. eth0, eth1, eh2 etc.
```
jpeach, line 436:
```suggestion
  * `name` - (Required) Interface name e.g. eth0, eth1, eth2 etc.
```

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -753,6 +753,31 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 					},
 				},
 			},
+			"interface_mapping": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "List of interface names mapped to interface types and indices.",
```
jpeach, line 759: Is this sufficient documentation? I feel like I would need a couple of paragraphs of explanation to even understand whether I need to worry about this parameter at all.

Are there public documents that we could link to here?

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -871,13 +896,20 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
...
-				"eth1": {"wan", "1"},
-				"eth2": {"wan", "2"},
-				"eth3": {"mgmt", "0"},
-				"eth4": {"wan", "3"},
+			interfaceMappingInput := d.Get("interface_mapping").([]interface{})
+			interfaceMapping := make(map[string][]string)
+			for _, value := range interfaceMappingInput {
+				mappingMap, ok := value.(map[string]interface{})
+				if !ok {
+					return fmt.Errorf("invalid type for interface mapping, expected a map")
```
jpeach, line 904:
```suggestion
					return fmt.Errorf("invalid type %T for interface mapping, expected a map", value)
```

aviatrix/utils.go
```diff
@@ -391,3 +391,20 @@ func sortInterfacesByCustomOrder(interfaces []goaviatrix.EdgeTransitInterface) [
 	})
 	return interfaces
 }
+
+// Sorting function that uses the interface mapping order
+func sortInterfaceMappingByCustomOrder(interfaceMapping []goaviatrix.InterfaceMapping) []goaviatrix.InterfaceMapping {
```
jpeach, line 396: We should add tests for this. What happens when you have `eth10`? What happens when the interface names are not in the `interfaceOrder` slice?

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -871,13 +896,20 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
...
+				}
+				interfaceName, ok1 := mappingMap["name"].(string)
+				interfaceType, ok2 := mappingMap["type"].(string)
+				interfaceIndex, ok3 := mappingMap["index"].(string)
+				if !ok1 || !ok2 || !ok3 {
+					return fmt.Errorf("invalid interface mapping, 'name', 'type', and 'index' must be strings")
+				}
+				interfaceMapping[interfaceName] = []string{interfaceType, interfaceIndex}
```
jpeach, line 912: What happens if `interfaceName` is not unique? What happens if `interfaceIndex` is not unique?

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -753,6 +753,31 @@ func resourceAviatrixTransitGateway() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "Interface name (e.g., 'eth0', 'eth1').",
+						},
+						"type": {
+							Type:         schema.TypeString,
+							Required:     true,
+							Description:  "Interface type (e.g., 'wan', 'mgmt').",
```
jpeach, line 770: The examples should exactly match what you validate for

docs/resources/aviatrix_transit_gateway.md
```diff
@@ -375,6 +375,31 @@ resource "aviatrix_transit_gateway" "test-edge-transit-1-hagw" {
...
+      index = "1"
+    }
+    interface_mapping{
+      name  = "eth2"
+      type  = "wan"
+      index = "2"
+    }
+    interface_mapping{
+      name  = "eth3"
+      type  = "mgmt"
```
jpeach, line 395: IIUC, only "WAN" and "MANAGEMENT" are accepted by validation?

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -753,6 +753,31 @@ func resourceAviatrixTransitGateway() *schema.Resource {
...
+							ValidateFunc: validation.StringInSlice([]string{"WAN", "MANAGEMENT"}, false),
+						},
+						"index": {
+							Type:        schema.TypeString,
```
jpeach, line 774: Why string? Could you specify TypeInt and validate >= 0?

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1909,6 +1941,21 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
...
+			sortedInterfaceMappings := sortInterfaceMappingByCustomOrder(gw.InterfaceMapping)
+			for _, intf := range sortedInterfaceMappings {
+				interfaceMap := make(map[string]interface{})
+				interfaceMap["name"] = intf.Name
+				interfaceMap["type"] = intf.Type
+				interfaceMap["index"] = intf.Index
+				interfaceMapping = append(interfaceMapping, interfaceMap)
+			}
+			if err = d.Set("interface_mapping", interfaceMapping); err != nil {
+				return fmt.Errorf("could not set interface mapping into state: %v", err)
```
jpeach, line 1956:
```suggestion
				return fmt.Errorf("failed to set \"interface_mapping\" key: %w", err)
```

Nit: use "%w" to stack the error, and phrase it to help the operator.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -1909,6 +1941,21 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
 				return fmt.Errorf("could not set interfaces into state: %v", err)
 			}
 		}
+		// set interface mapping
+		if len(gw.InterfaceMapping) != 0 {
+			var interfaceMapping []map[string]interface{}
+			sortedInterfaceMappings := sortInterfaceMappingByCustomOrder(gw.InterfaceMapping)
```
jpeach, line 1947: Can you add a comment with an explanation of why the special sorting is needed? Future maintainers will thank you :)

sairane10:
Changed the interface mapping config

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: interface_mapping is an optional field. Users don't have to provide this value. The mapping value is expected only for esxi devices which is used in internal lab testing.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Added the change in validation function

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Since the map is declared as a part of the list, its only accepting either all string values or int values. Changing the index type to int throws an error while using interface_mapping. Could not find a work around for this so using a string type for index.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Added the change

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: The backend api handles the validation of interface mapping.

PR #2057, opened 2024-09-23
AVX-56803 Updating the Interface config and mapping

sairane10:
- Adding the interface mapping support for Dell and ESXI devices in EAT
- Updated the interface config to accept index and removed the interface name. The name will be updated using the interface type and index provided by the user.
- By default, we are setting the interface mapping to use the dell configuration but if the user provides an interface mapping then it is set for ESXI devices

_Example to provide interface mapping only for ESXI devices_
`resource "aviatrix_transit_gateway" "test-edge-transit-15" {
    cloud_type = 262144
    account_name = "edge_aep"
    gw_name = "test-edge-transit-15"
    vpc_id = "site-15"
    gw_size = "SMALL"
    device_id = "7cf4a6fd-a772-4d0d-a655-2d3d436985ff"
    site_id = "site-15"
    local_as_number = 65441
    interfaces {
        gateway_ip = "192.168.20.1"
        ip_address    = "192.168.20.11/24"
        type       = "WAN"
        index = 0
    }

    interfaces {
        gateway_ip = "192.168.21.1"
        ip_address    = "192.168.21.11/24"
        type       = "WAN"
        index = 1
        secondary_private_cidr_list = ["192.168.21.16/29"]
    }

    interfaces {
        dhcp   = true
        type   = "MANAGEMENT"
        index = 0
    }

    interfaces {
        gateway_ip                  = "192.168.22.1"
        ip_address                     = "192.168.22.11/24"
        type                        = "WAN"
        index = 2
    }

    interfaces {
        gateway_ip                  = "192.168.23.1"
        ip_address                     = "192.168.23.11/24"
        type                        = "WAN"
        index = 3
    }

    interface_mapping{
      name  = "eth0"
      type  = "wan"
      index = "0"
    }

    interface_mapping{
      name  = "eth1"
      type  = "wan"
      index = "1"
    }

    interface_mapping{
      name  = "eth2"
      type  = "wan"
      index = "2"
    }

    interface_mapping{
      name  = "eth3"
      type  = "mgmt"
      index = "0"
    }

    interface_mapping{
      name  = "eth4"
      type  = "wan"
      index = "3"
    }
}`

### DIFF
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -757,7 +757,7 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 			"interface_mapping": {
 				Type:        schema.TypeList,
 				Optional:    true,
-				Description: "List of interface names mapped to interface types and indices.",
+				Description: "List of interface names mapped to interface types and indices. Only required for ESXI.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
@@ -772,7 +772,7 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 							ValidateFunc: validation.StringInSlice([]string{"wan", "mgmt"}, false),
 						},
 						"index": {
-							Type:        schema.TypeString,
+							Type:        schema.TypeInt,
 							Required:    true,
 							Description: "Interface index (e.g., '0', '1').",
 						},
@@ -915,7 +915,7 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 
 		if goaviatrix.IsCloudType(cloudType, goaviatrix.EDGENEO) {
 			/*
-				TODO: Add support for user provided interface mapping. Currently, the interface mapping is hardcoded for Dell devices. This will be updated in a future release.
+				TODO: Use the device_id to determine the interface mapping. This change will provide support for other device models interface mapping. For now, we will use the user provided interface mapping for ESXI devices and default values for Dell devices.
 			*/
 			interfaceMapping := map[string][]string{}
 			interfaceMappingInput := d.Get("interface_mapping").([]interface{})
@@ -924,15 +924,15 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 				for _, value := range interfaceMappingInput {
 					mappingMap, ok := value.(map[string]interface{})
 					if !ok {
-						return fmt.Errorf("invalid type for interface mapping, expected a map")
+						return fmt.Errorf("invalid type %T for interface mapping, expected a map", value)
 					}
 					interfaceName, ok1 := mappingMap["name"].(string)
 					interfaceType, ok2 := mappingMap["type"].(string)
-					interfaceIndex, ok3 := mappingMap["index"].(string)
+					interfaceIndex, ok3 := mappingMap["index"].(int)
 					if !ok1 || !ok2 || !ok3 {
 						return fmt.Errorf("invalid interface mapping, 'name', 'type', and 'index' must be strings")
 					}
-					interfaceMapping[interfaceName] = []string{interfaceType, interfaceIndex}
+					interfaceMapping[interfaceName] = []string{interfaceType, strconv.Itoa(interfaceIndex)}
 				}
 			} else {
 				// Set the interface mapping for Dell devices
@@ -1974,6 +1974,7 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
 		// set interface mapping
 		if len(gw.InterfaceMapping) != 0 {
 			var interfaceMapping []map[string]interface{}
+			// sort the interface mapping by interface name. This will maintain the order of the interface mappings in the state
 			sortedInterfaceMappings := sortInterfaceMappingByCustomOrder(gw.InterfaceMapping)
 			for _, intf := range sortedInterfaceMappings {
 				interfaceMap := make(map[string]interface{})
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -797,7 +797,7 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 			return fmt.Errorf("interfaces attribute is required for EAT gateway")
 		}
 		// get the count of WAN and MANAGEMENT interfaces
-		wanCount, mgmtCount, err := countInterfaceTypes(interfaces)
+		wanCount, err := countInterfaceTypes(interfaces)
 		if err != nil {
 			return fmt.Errorf("failed to get the interface count: %v", err)
 		}
@@ -827,11 +827,11 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
 			ifaceIndex = ifaceInfo["index"].(int)
 
 			// get the interface name using the interface type, index and count
-			// getInterfaceName(intfType string, intfIndex, wanCount, mgmtCount int) (string, error)
-			ifaceName, err := getInterfaceName(ifaceType, ifaceIndex, wanCount, mgmtCount)
+			ifaceName, err := getInterfaceName(ifaceType, ifaceIndex, wanCount)
 			if err != nil {
 				return fmt.Errorf("failed to get the interface name: %v", err)
 			}
+
 			// Append values to interface mapping
 			interfaceMapping[ifaceName] = []string{mappedIfaceType, strconv.Itoa(ifaceIndex)}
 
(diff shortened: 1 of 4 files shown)


aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -753,6 +754,31 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 					},
 				},
 			},
+			"interface_mapping": {
```
ssundarrajan, line 757: do we still need this after your recent changes ?

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -871,13 +914,33 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
...
-				"eth2": {"wan", "2"},
-				"eth3": {"mgmt", "0"},
-				"eth4": {"wan", "3"},
+			/*
+				TODO: Add support for user provided interface mapping. Currently, the interface mapping is hardcoded for Dell devices. This will be updated in a future release.
+			*/
+			interfaceMapping := map[string][]string{}
+			interfaceMappingInput := d.Get("interface_mapping").([]interface{})
+			if interfaceMappingInput != nil {
+				// Set the interface mapping for ESXI devices
```
ssundarrajan, line 924: Comments are conflicting, ESXi vs Dell

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Hi @ssundarrajan, yes we need interface mapping for only the esxi devices. By default if the user provides no value for this then we will be setting the interface mapping to dell configuration. This was implemented after my discussion with Purnima and Rui.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: In this implementation, if the user provides the interface mapping then we are assuming its for internal testing on ESXI devices but for production use case, no input is expected from the user and the interface mapping is set for Dell devices by default.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -753,6 +754,31 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 					},
 				},
 			},
+			"interface_mapping": {
+				Type:        schema.TypeList,
```
dnardo, line 758: Does order matter here? Do you want to avoid duplicates?  You may want to consider TypeSet

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -753,6 +754,31 @@ func resourceAviatrixTransitGateway() *schema.Resource {
...
+							Description: "Interface name (e.g., 'eth0', 'eth1').",
+						},
+						"type": {
+							Type:         schema.TypeString,
+							Required:     true,
+							Description:  "Interface type (e.g., 'wan', 'mgmt').",
+							ValidateFunc: validation.StringInSlice([]string{"wan", "mgmt"}, false),
+						},
+						"index": {
+							Type:        schema.TypeString,
```
dnardo: Is there a reason we do string here and not int?
Side beneifit here is you can do something like:

ValidateFunc: validation.IntAtLeast(0),

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -753,6 +754,31 @@ func resourceAviatrixTransitGateway() *schema.Resource {
...
+			"interface_mapping": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "List of interface names mapped to interface types and indices.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "Interface name (e.g., 'eth0', 'eth1').",
```
dnardo, line 766: How flexible is name?  Must it be eth\d+$?  If so I would add a validation function 

ValidateFunc: validation.StringMatch(regexp.MustCompile(`^eth\d+$`)

Or if we think this needs to be flexible leave as is.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -871,13 +914,33 @@ func resourceAviatrixTransitGatewayCreate(d *schema.ResourceData, meta interface
...
-			interfaceMapping := map[string][]string{
-				"eth0": {"wan", "0"},
-				"eth1": {"wan", "1"},
-				"eth2": {"wan", "2"},
-				"eth3": {"mgmt", "0"},
-				"eth4": {"wan", "3"},
+			/*
+				TODO: Add support for user provided interface mapping. Currently, the interface mapping is hardcoded for Dell devices. This will be updated in a future release.
+			*/
+			interfaceMapping := map[string][]string{}
```
dnardo, line 921: So interface mappings will only ever be required for edge is that right?  And is that the long term plan or we plan to remove interface mapping altogether in the future?

dnardo (changes requested):

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: I tried using a typeInt for index but the interface_mapping started throwing an unexpected value error. When I changed the type to string again, the error went away. My assumption is that the TF provider does not like a complex type map inside a list or set.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Updated the comment

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: I need the list ordering for interface mapping and the current interface_mapping works with the example provided in the description.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: for other devices this name might look different so keeping it flexible for now.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Updated to int. Added the changes.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: The plan is to remove interface_mapping in the future and internally create a mapping in the code using the device_id provided by the user.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -3769,3 +3848,62 @@ func resourceAviatrixTransitGatewayDelete(d *schema.ResourceData, meta interface
...
+func getInterfaceName(intfType string, intfIndex, wanCount int) (string, error) {
+	// Check if the interface type is valid
+	if intfType != "WAN" && intfType != "MANAGEMENT" {
+		log.Printf("Invalid interface type %s", intfType)
+		return "", errors.New("invalid interface type " + intfType)
+	}
+	var num int
+	// Determine the interface name based on the type and index
+	switch intfType {
+	case "WAN":
```
dnardo, line 3883: I'm a little bit concerned about the interface mapping here:
With wanCount = 2 and intfIndex = 2 for both WAN and MANAGEMENT interfaces, both interfaces will map to num = 3, which corresponds to eth3. I don't know if that particular use case is valid or will be valid in the future. 

I would say regardless at some point (doesn't need to be this PR) we need to document the logic for interface mapping for what is and is not valid, so we can compare the code here to the actual intentions.  This seem brittle to me.

dnardo (dismissed):

docs/resources/aviatrix_transit_gateway.md
```diff
@@ -400,13 +400,17 @@ The following arguments are supported:
...
-  * `name` - (Required) Interface name e.g. eth0, eth1, eh2 etc.
   * `type` - (Required) Interface type. Valid values are 'WAN' or 'MANAGEMENT'.
+  * `index` - (Required) Interface index. Valid values are 0,1,2 etc.
   * `gateway_ip` - (Optional) The gateway IP address associated with this interface.
   * `ip_address` - (Optional) The static IP address assigned to this interface.
   * `public_ip` - (Optional) The public IP address associated with this interface.
   * `dhcp` - (Optional) Whether DHCP is enabled on this interface. Set the value to true or false. Applicable to only 'MANAGEMENT' type interface.
   * `secondary_private_cidr_list` - (Optional) A list of secondary private CIDR blocks associated with this interface.
+* `interface_mapping` - (Optional) A list of interface names mapped to interface types and indices. Required and valid only for edge transit gateways (AEP). Each interface has the following attributes:
+  * `name` - (Required) Interface name e.g. eth0, eth1, eh2 etc.
```
dnardo:
```suggestion
  * `name` - (Required) Interface name e.g. eth0, eth1, eth2 etc.
```

docs/resources/aviatrix_transit_gateway.md
```diff
@@ -303,33 +303,33 @@ resource "aviatrix_transit_gateway" "test-edge-transit-1" {
     site_id = "site-3"
     interfaces {
         gateway_ip    = "192.168.20.1"
-        name          = "eth0"
         ip_address    = "192.168.20.11/24"
         type          = "WAN"
+        index         = 0
```
dnardo, line 308: Should there also be aep documentation changes?

dnardo (changes requested):

docs/resources/aviatrix_transit_gateway.md
sairane10: AEP documentation change wont be required here.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: The interface mappings are validated in the backend api call. If the mgmt interface or wan interface in the above case is not set then the api throws an error.

dnardo (approved):

PR #2048, opened 2024-09-16
AVX-54790 Adding BGP BFD polling time for resources

sairane10:
Adding the bgp-bfd polling time to the following resources:
 - aviatrix_edge_equinix
 - aviatrix_edge_gateway_selfmanaged
 - avaitrix_edge_platform
 - aviatrix_edge_zededa
 - aviatrix_spoke_gateway
 - aviatrix_edge_spoke_gateway
 - aviatrix_transit_gateway

PR #2089, opened 2024-10-29
AVX-58209 Updating the transit gateway with bgp-bfd polling time

sairane10:
This bug fixes the missing bgp-bfd polling time in transit gateway. This change got overridden while merging conflicts.
https://aviatrix.atlassian.net/browse/AVX-58209
UT is already a part of my previously merged PR - https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2048

### DIFF
--- a/docs/guides/feature-changelist-v3.md
+++ b/docs/guides/feature-changelist-v3.md
@@ -95,7 +95,7 @@ The following logging resources are removed:
 |(deprecated) | firenet | keep_alive_via_lan_interface_enabled | **Yes**; This configuration value no longer has any effect. It will be removed from the Aviatrix provider in the 3.2.0 release |
 
 
-## R3.2.1 (UserConnect-7.2)**
+## R3.2.1 (UserConnect-7.2)
 ### Attribute Change
 | Diff | Resource | Attribute | Action Required? |
 |:----:|:--------:|:---------:|:----------------:|
--- a/docs/guides/feature-changelist-v3.md
+++ b/docs/guides/feature-changelist-v3.md
@@ -93,3 +93,10 @@ The following logging resources are removed:
 |:----:|:--------:|:---------:|:----------------:|
 |(deprecated) | edge_equinix, edge_equinix_ha, edge_csp, edge_csp_ha, edge_neo, edge_neo_ha, edge_platform, edge_platform_ha, edge_zededa, edge_zededa_ha | bandwidth | **Yes**; This configuration value no longer has any effect. It will be removed from the Aviatrix provider in the 3.2.0 release |
 |(deprecated) | firenet | keep_alive_via_lan_interface_enabled | **Yes**; This configuration value no longer has any effect. It will be removed from the Aviatrix provider in the 3.2.0 release |
+
+
+## R3.2.1 (UserConnect-7.2)**
+### Attribute Change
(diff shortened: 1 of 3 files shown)


inc1t3Ful (changes requested):
Please update missing documentation for this new attribute in the transit gateway doc

inc1t3Ful:
I noticed since you changed the `bgp_polling_time` to Int type, did you test how that would affect existing resource?

aka if a transit was created prior to this change with `bgp_polling_time` set (as string), and then when they update their provider with this new change to Int type, will they get a delta/will there be any issues?

sairane10:
@inc1t3Ful This documentation change are already a part of this merged PR - https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2048
I had a transit deployed with bgp_polling_time set as a string value. On applying this change it showed a delta on the existing resource but there no issues encountered.

inc1t3Ful:
@sairane10 On applying the code change, it shows a delta which is expected. But will the existing resource delta be resolved in the TF apply? As in, if on your next TF operation (apply), will that delta still be shown? 

I assume after that users would have to normally just change their infra code from "50" to 50?

sairane10:
@inc1t3Ful On my TF apply operation the delta was show as I had to change the value to int instead of a string. The users will have to manually apply this change.

inc1t3Ful:
> @inc1t3Ful This documentation change are already a part of this merged PR - #2048 I had a transit deployed with bgp_polling_time set as a string value. On applying this change it showed a delta on the existing resource but there no issues encountered.

Can we also update the R3.x Feature Changelist doc and ensure that this data type change is noted? Please go create a table for R3.2.1 (I assume it is for that version release), and list the "Diff" column as "(changed)" , where "Action Required" is "Yes; ... "  and provide note/instruction on what the user should do.

You can also refer to the R2.x Feature Changelist for reference if you are unsure about the wording. 
- https://registry.terraform.io/providers/AviatrixSystems/aviatrix/latest/docs/guides/feature-changelist-v2#boolean-standardization

sairane10:
@inc1t3Ful Added the feature changelist document update.

docs/guides/feature-changelist-v3.md
```diff
@@ -93,3 +93,10 @@ The following logging resources are removed:
 |:----:|:--------:|:---------:|:----------------:|
 |(deprecated) | edge_equinix, edge_equinix_ha, edge_csp, edge_csp_ha, edge_neo, edge_neo_ha, edge_platform, edge_platform_ha, edge_zededa, edge_zededa_ha | bandwidth | **Yes**; This configuration value no longer has any effect. It will be removed from the Aviatrix provider in the 3.2.0 release |
 |(deprecated) | firenet | keep_alive_via_lan_interface_enabled | **Yes**; This configuration value no longer has any effect. It will be removed from the Aviatrix provider in the 3.2.0 release |
+
+
+## R3.2.1 (UserConnect-7.2)**
```
inc1t3Ful:
```suggestion
## R3.2.1 (UserConnect-7.2)
```

inc1t3Ful (changes requested):

inc1t3Ful (approved):
LGTM

PR #2051, opened 2024-09-18
AVX-54790 Adding BGP BFD config to external connection

sairane10:
Updating the following resource with bgp_bfd config
- aviatrix_edge_spoke_external_device_conn
- avaitrix_transit_external_device_conn

Example of the connection resource with BGP-BFD enabled
```
`resource "aviatrix_transit_external_device_conn" "transit-1-transit-2" {
  vpc_id            = "vpc-048eb7341becf3c1b"
  connection_name   = "transit-1-transit-2"
  gw_name           = "udxjl-aws-transit-1"
  connection_type   = "bgp"
  tunnel_protocol    = "IPsec"
  bgp_local_as_num  = "65075"
  bgp_remote_as_num = "65076"
  remote_gateway_ip = "3.19.43.179"
  local_tunnel_cidr  = "169.254.10.1/30"
  remote_tunnel_cidr = "169.254.10.2/30"
  pre_shared_key = "psk12"
  enable_bfd = true
  bgp_bfd {
    transmit_interval = 500
    receive_interval = 500
    multiplier = 5
  }
}

resource "aviatrix_transit_external_device_conn" "transit-2-transit-1" {
  vpc_id            = "vpc-00951e22f3e587165"
  connection_name   = "transit-2-transit-1"
  gw_name           = "udxjl-aws-transit-2"
  connection_type   = "bgp"
  tunnel_protocol    = "IPsec"
  bgp_local_as_num  = "65076"
  bgp_remote_as_num = "65075"
  remote_gateway_ip = "3.140.172.87"
  local_tunnel_cidr  = "169.254.10.2/30"
  remote_tunnel_cidr = "169.254.10.1/30"
  pre_shared_key = "psk12"
}

resource "aviatrix_edge_spoke_external_device_conn" "edge_bgp_lan" {
  site_id                  = "SC"
  connection_name          = "cloudn-bgp-lan"
  gw_name                  = "Edge-182"
  bgp_local_as_num         = "65182"
  bgp_remote_as_num        = "65220"
  local_lan_ip             = "10.220.86.182"
  remote_lan_ip            = "10.220.86.100"
  enable_bfd               = true
  bgp_bfd {
    transmit_interval = 400
    receive_interval = 400
    multiplier = 5
  }
}`
```

PR #2094, opened 2024-11-05
AVX-58386 Adding BGP BFD config to spoke connection

sairane10:
- Updating the spoke external connection with bgp_bfd config
- Adding the check for enable_bfd only for bgp connections
Link - https://aviatrix.atlassian.net/browse/AVX-58386

```
resource "aviatrix_spoke_external_device_conn" "test_spoke_conn" {
	vpc_id            = "vpc-0c9ec6eeb879174bb"
	connection_name   = "test_spoke_conn"
	gw_name           = "cnsvf-aws-spoke-1"
	connection_type   = "bgp"
	bgp_local_as_num  = "65078"
	bgp_remote_as_num = "65079"
	remote_gateway_ip = "35.167.229.132"
        enable_bfd = true
        bgp_bfd {
          transmit_interval = 400
          receive_interval = 400
          multiplier = 5
        }
}
```

### DIFF
--- a/aviatrix/resource_aviatrix_spoke_external_device_conn.go
+++ b/aviatrix/resource_aviatrix_spoke_external_device_conn.go
@@ -728,33 +728,11 @@ func resourceAviatrixSpokeExternalDeviceConnCreate(d *schema.ResourceData, meta 
 				if !ok {
 					return fmt.Errorf("expected bgp_bfd to be a map, but got %T", bfd0)
 				}
-				transmitInterval := defaultBfdTransmitInterval
-				receiveInterval := defaultBfdReceiveInterval
-				multiplier := defaultBfdMultiplier
-				if value, ok := bfd1["transmit_interval"].(int); ok {
-					transmitInterval = value
-				}
-				if value, ok := bfd1["receive_interval"].(int); ok {
-					receiveInterval = value
-				}
-				if value, ok := bfd1["multiplier"].(int); ok {
-					multiplier = value
-				}
-				bfd2 := &goaviatrix.BgpBfdConfig{
-					TransmitInterval: transmitInterval,
-					ReceiveInterval:  receiveInterval,
-					Multiplier:       multiplier,
-				}
-				externalDeviceConn.BgpBfdConfig = append(externalDeviceConn.BgpBfdConfig, bfd2)
+				externalDeviceConn.BgpBfdConfig = goaviatrix.CreateBgpBfdConfig(bfd1)
 			}
 		} else {
 			// set the bgp bfd config using the default values
-			bfd := &goaviatrix.BgpBfdConfig{
-				TransmitInterval: defaultBfdTransmitInterval,
-				ReceiveInterval:  defaultBfdReceiveInterval,
-				Multiplier:       defaultBfdMultiplier,
-			}
-			externalDeviceConn.BgpBfdConfig = append(externalDeviceConn.BgpBfdConfig, bfd)
+			externalDeviceConn.BgpBfdConfig = &defaultBfdConfig
 		}
 		err := client.EditConnectionBgpBfd(externalDeviceConn)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -382,6 +382,43 @@ func resourceAviatrixSpokeExternalDeviceConn() *schema.Resource {
...
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Enable BGP BFD connection.",
+			},
+			"bgp_bfd": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "BGP BFD configuration details applied to a BGP session.",
+				MaxItems:    1,
```
jpeach, line 395: Why is this a list with max one? Could it be an optional Map instead? The code that extract this doesn't seem to care about the MaxItems limit?

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -1500,12 +1500,12 @@ func resourceAviatrixSpokeGatewayRead(d *schema.ResourceData, meta interface{})
 	if gw.EnableBgp {
 		d.Set("learned_cidrs_approval_mode", gw.LearnedCidrsApprovalMode)
 		d.Set("bgp_polling_time", gw.BgpPollingTime)
-		d.Set("bgp_bdf_polling_time", gw.BgpBfdPollingTime)
+		d.Set("bgp_bfd_polling_time", gw.BgpBfdPollingTime)
```
jpeach: Are there tests to catch this kind of error?

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -670,6 +707,61 @@ func resourceAviatrixSpokeExternalDeviceConnCreate(d *schema.ResourceData, meta
...
+					receiveInterval = value
+				}
+				if value, ok := bfd1["multiplier"].(int); ok {
+					multiplier = value
+				}
+				bfd2 := &goaviatrix.BgpBfdConfig{
+					TransmitInterval: transmitInterval,
+					ReceiveInterval:  receiveInterval,
+					Multiplier:       multiplier,
+				}
```
jpeach: Suggest factoring this into a function that takes the terraform config and returns the BgpBfdConfig struct, or error.

You ought to be able to write tests to exercise the error paths of that.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -670,6 +707,61 @@ func resourceAviatrixSpokeExternalDeviceConnCreate(d *schema.ResourceData, meta
...
+				externalDeviceConn.BgpBfdConfig = append(externalDeviceConn.BgpBfdConfig, bfd2)
+			}
+		} else {
+			// set the bgp bfd config using the default values
+			bfd := &goaviatrix.BgpBfdConfig{
+				TransmitInterval: defaultBfdTransmitInterval,
+				ReceiveInterval:  defaultBfdReceiveInterval,
+				Multiplier:       defaultBfdMultiplier,
+			}
+			externalDeviceConn.BgpBfdConfig = append(externalDeviceConn.BgpBfdConfig, bfd)
```
jpeach: Why is `BgpBfdConfig` a slice when the only API call (edit_connection_bgp_bfd) explicitly only sends the values of the first element?

aviatrix/resource_aviatrix_spoke_external_device_conn.go
jpeach: There's also half a dozen other places where you could use this function.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -1113,6 +1235,74 @@ func resourceAviatrixSpokeExternalDeviceConnUpdate(d *schema.ResourceData, meta
...
+	}
+	if enableBfd {
+		// get the new BGP BFD config
+		bgpBfdConfig, ok := d.Get("bgp_bfd").([]interface{})
+		if !ok {
+			return fmt.Errorf("expected bgp_bfd to be a list of maps, but got %T", d.Get("bgp_bfd"))
+		}
+		var bgpBfdConfigList []*goaviatrix.BgpBfdConfig
+		// Update the BGP BFD config if bfd is enabled and config has changed
+		if len(bgpBfdConfig) > 0 && d.HasChange("bgp_bfd") {
```
jpeach: I'm not sure I follow this. If you have the config, I think you should use the values from that whether or not HasChange is true. If HasChange is false, is forcing the defaults the right things to do?

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -1113,6 +1235,74 @@ func resourceAviatrixSpokeExternalDeviceConnUpdate(d *schema.ResourceData, meta
...
+			ConnectionName: d.Get("connection_name").(string),
+			EnableBfd:      d.Get("enable_bfd").(bool),
+			BgpBfdConfig:   bgpBfdConfigList,
+		}
+		err := client.EditConnectionBgpBfd(externalDeviceConn)
+		if err != nil {
+			return fmt.Errorf("could not update BGP BFD config: %v", err)
+		}
+	} else {
+		if d.HasChange("enable_bfd") {
```
jpeach: Shouldn't we always force the terraform desired state onto the controller? If the BFD was enabled out of band but not in terraform, then would a terraform apply set it back?

jpeach (commented):
Had a couple of questions. Would be good to refactor so there is one function that extracts the BFD config in a canonical way.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
sairane10: I wanted to provide support for bgp_bfd config that looks like this
```
bgp_bfd {
     transmit_interval = 300
     receive_interval   = 300
     multiplier              = 3
}
```
TF map does not support complex key - values which in our case bgp_bfd value is. Hence the choice was made to select the list type with a max item list of 1. For a given external connection there will only be 1 bgp_bfd config.
This implementation is an extension of this merged PR - https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2051

aviatrix/resource_aviatrix_spoke_gateway.go
sairane10: Yes I have added test to resource_avaitrix_spoke_gateway_test.go

aviatrix/resource_aviatrix_spoke_external_device_conn.go
sairane10: updated

aviatrix/resource_aviatrix_spoke_external_device_conn.go
sairane10: updated bGPbfdConfig

aviatrix/resource_aviatrix_spoke_external_device_conn.go
sairane10: Updated the logic to follow this:
if bfd is enabled and bgp_bfd config is changed then set the new config
else if bfd is enabled but no bgp_bfd config is provided then use the default
else if bfd is disabled then remove the bgp_bfd config
else no change.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
sairane10: I have tweaked the bgp_bfd config. Open to suggestions on how I can improve it further.

sairane10: Updated the code to consider bgp_bfd changes

sairane10:
@jpeach I have resolved the comments on this PR. Please take a look at the pr again.

jpeach (approved):

PR #2107, opened 2024-11-14
AVX-58677 Megaport Edge as a Spoke

sairane10:
Adding the resource to support megaport edge as a spoke.
Link - https://aviatrix.atlassian.net/browse/AVX-58677

```
resource "aviatrix_edge_megaport" "spokeMegaport" {
    account_name = "megaportacct1"
    gw_name = "spokeMegaportFabric"
    site_id = "testMegaportSite1"
    ztp_file_download_path  = "ztp"
    local_as_number = "65445"
    prepend_as_path = ["65445"]
    interfaces {
        gateway_ip = "10.220.14.1"
        ip_address = "10.220.14.10/24"
        type       = "LAN"
        index      = 0
    }

    interfaces {
        enable_dhcp   = true
        type   = "MANAGEMENT"
        index  = 0
    }

    interfaces {
        gateway_ip    = "192.168.99.1"
        ip_address    = "192.168.99.14/24"
        type          = "WAN"
        index         = 0
        wan_public_ip     = "67.207.104.19"
    }

    interfaces {
        gateway_ip    = "192.168.88.1"
        ip_address    = "192.168.88.14/24"
        type          = "WAN"
        index         = 1
        wan_public_ip     = "67.71.12.148"
    }

    interfaces {
        gateway_ip  = "192.168.77.1"
        ip_address  = "192.168.77.14/24"
        type        = "WAN"
        index       = 2
        wan_public_ip   = "67.72.12.149"
    }
}
```

### DIFF
--- a/aviatrix/resource_aviatrix_edge_megaport_test.go
+++ b/aviatrix/resource_aviatrix_edge_megaport_test.go
@@ -1,0 +1,136 @@
+package aviatrix
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+
+	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccAviatrixEdgeMegaport_basic(t *testing.T) {
+	if os.Getenv("SKIP_EDGE_Megaport") == "yes" {
+		t.Skip("Skipping Edge Megaport test as SKIP_EDGE_Megaport is set")
+	}
+
+	resourceName := "aviatrix_edge_Megaport.test"
+	accountName := "acc-" + acctest.RandString(5)
+	edgeMegaportUsername := "megaport-user-" + acctest.RandString(5)
+	gwName := "gw-" + acctest.RandString(5)
+	siteId := "site-" + acctest.RandString(5)
+	path, _ := os.Getwd()
+
+	resource.Test(t, resource.TestCase{
+		PreCheck: func() {
+			testAccPreCheck(t)
+		},
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckEdgeMegaportDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccEdgeMegaportBasic(accountName, edgeMegaportUsername, gwName, siteId, path),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckEdgeMegaportExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "gw_name", gwName),
+					resource.TestCheckResourceAttr(resourceName, "site_id", siteId),
+					resource.TestCheckResourceAttr(resourceName, "interfaces.0.ip_address", "10.230.5.32/24"),
+					resource.TestCheckResourceAttr(resourceName, "interfaces.1.ip_address", "10.230.3.32/24"),
+					resource.TestCheckResourceAttr(resourceName, "interfaces.2.ip_address", "172.16.15.162/20"),
+					resource.TestCheckResourceAttr(resourceName, "bgp_polling_time", "50"),
+					resource.TestCheckResourceAttr(resourceName, "bgp_neighbor_status_polling_time", "5"),
+				),
+			},
+			{
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"ztp_file_download_path"},
+			},
+		},
+	})
+}
+
+func testAccEdgeMegaportBasic(edgeMegaportUsername, accountName, gwName, siteId, path string) string {
+	return fmt.Sprintf(`
+resource "aviatrix_account" "test" {
+	account_name          = "%s"
+	cloud_type            = 524288
+	edge_Megaport_username = "%s"
+}
+resource "aviatrix_edge_Megaport" "test" {
+	account_name                       = aviatrix_account.test_account.account_name
+	gw_name                            = "%s"
+	site_id                            = "%s"
+	ztp_file_download_path             = "%s"
+	bgp_polling_time                   = 50
+	bgp_neighbor_status_polling_time   = 5
+	
+	interfaces {
+		name          = "eth0"
+		type          = "WAN"
+		ip_address    = "10.230.5.32/24"
+		gateway_ip    = "10.230.5.100"
+		wan_public_ip = "64.71.24.221"
+	}
+	
+	interfaces {
+		name       = "eth1"
+		type       = "LAN"
+		ip_address = "10.230.3.32/24"
+	}
+	
+	interfaces {
+		name        = "eth2"
+		type        = "MANAGEMENT"
+		enable_dhcp = false
+		ip_address  = "172.16.15.162/20"
+		gateway_ip  = "172.16.0.1"
+	}
+}
+ `, accountName, edgeMegaportUsername, gwName, siteId, path)
+}
+
+func testAccCheckEdgeMegaportExists(resourceName string) resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		rs, ok := s.RootModule().Resources[resourceName]
+		if !ok {
+			return fmt.Errorf("edge megaport not found: %s", resourceName)
+		}
+		if rs.Primary.ID == "" {
+			return fmt.Errorf("no edge megaport id is set")
+		}
+
+		client := testAccProvider.Meta().(*goaviatrix.Client)
+
+		edgeSpoke, err := client.GetEdgeMegaport(context.Background(), rs.Primary.Attributes["gw_name"])
+		if err != nil {
+			return err
+		}
+		if edgeSpoke.GwName != rs.Primary.ID {
+			return fmt.Errorf("could not find edge megaport")
+		}
+		return nil
+	}
+}
+
+func testAccCheckEdgeMegaportDestroy(s *terraform.State) error {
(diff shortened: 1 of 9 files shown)


goaviatrix/edge_megaport.go
```diff
@@ -0,0 +1,262 @@
...
+	"strings"
+)
+
+type EdgeMegaport struct {
+	Action                             string `json:"action,omitempty"`
+	CID                                string `json:"CID,omitempty"`
+	AccountName                        string `json:"account_name,omitempty"`
+	GwName                             string `json:"name,omitempty"`
+	SiteId                             string `json:"site_id,omitempty"`
+	ZtpFileDownloadPath                string
```
rpanzer-aviatrix: Why does this not have a json tag?
If the idea is to not marshal this, `json:"-"` might be better.

goaviatrix/edge_megaport.go
```diff
@@ -0,0 +1,262 @@
...
+	Result string `json:"results"`
+	Reason string `json:"reason"`
+}
+
+func (c *Client) CreateEdgeMegaport(ctx context.Context, edgeMegaport *EdgeMegaport) error {
+	edgeMegaport.Action = "create_megaport_instance"
+	edgeMegaport.CID = c.CID
+	edgeMegaport.NoProgressBar = true
+
+	interfaces, err := json.Marshal(edgeMegaport.InterfaceList)
```
rpanzer-aviatrix, line 158: What is the rationale behind serialising this to base64(json) instead of having it just in the json?

docs/resources/aviatrix_edge_megaport.md
```diff
@@ -0,0 +1,119 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+### Required
+* `account_name` - (Required) Edge Megaport account name.
+* `gw_name` - (Required) Edge Megaport name.
```
rpanzer-aviatrix: Does `gw` in `gw_name` stand for `gateway`?
Should the description mention this term?

docs/resources/aviatrix_edge_megaport.md
```diff
@@ -0,0 +1,119 @@
...
+* `site_id` - (Required) Site ID.
+* `ztp_file_download_path` - (Required) The folder path where the ZTP file will be downloaded.
+* `interfaces` - (Required) WAN/LAN/MANAGEMENT interfaces.
+  * `name` - (Required) Interface name.
+  * `type` - (Required) Type. Valid values: WAN, LAN, or MANAGEMENT.
+  * `bandwidth` - (Optional) The rate of data can be moved through the interface, requires an integer value. Unit is in Mb/s.
+  * `enable_dhcp` - (Optional) Enable DHCP. Valid values: true, false. Default value: false.
+  * `wan_public_ip` - (Optional) WAN public IP.
+  * `ip_address` - (Optional) Interface static IP address.
```
rpanzer-aviatrix, line 78: All other "ip"s are `<prefix>-ip`. Only this is `ip-<suffix>`.
Would it make sense to align the form with the other properties?

docs/resources/aviatrix_edge_megaport.md
```diff
@@ -0,0 +1,119 @@
...
+  * `gateway_ip` - (Optional) Gateway IP.
+  * `dns_server_ip` - (Optional) Primary DNS server IP.
+  * `secondary_dns_server_ip` - (Optional) Secondary DNS server IP.
+  * `enable_vrrp` - (Optional) Enable VRRP. Valid values: true, false. Default value: false.
+  * `vrrp_virtual_ip` - (Optional) VRRP virtual IP.
+  * `tag` - (Optional) Tag.
```
rpanzer-aviatrix, line 84: There is no example for tags, nor is there a documentation of the schema.
It might be easier to add a tag to the example above.

docs/resources/aviatrix_edge_megaport.md
```diff
@@ -0,0 +1,119 @@
...
+* `bgp_polling_time` - (Optional) BGP route polling time in seconds. Valid values are between 10 and 50. Default value: 50.
+* `bgp_neighbor_status_polling_time` - (Optional) BGP neighbor status polling time in seconds. Valid values are between 1 and 10. Default value: 5.
+* `bgp_hold_time` - (Optional) BGP hold time in seconds. Valid values are between 12 and 360. Default value: 180.
+* `enable_edge_transitive_routing` - (Optional) Switch to enable Edge transitive routing. Valid values: true, false. Default value: false.
+* `enable_jumbo_frame` - (Optional) Switch to enable jumbo frame. Valid values: true, false. Default value: false.
+* `latitude` - (Optional) Latitude of Edge Megaport. Valid values are between -90 and 90. Example: "47.7511".
+* `longitude` - (Optional) Longitude of Edge Megaport. Valid values are between -180 and 180. Example: "120.7401".
+* `rx_queue_size` - (Optional) Ethernet interface RX queue size. Once set, can't be deleted or disabled. Valid values: "1K", "2K", "4K".
+* `vlan` - (Required) VLAN configuration.
```
rpanzer-aviatrix: Is this "(Required)" correct?
It is in the section "Optional" and the examples don't contain this property.

aviatrix/resource_aviatrix_edge_megaport_test.go
```diff
@@ -0,0 +1,136 @@
...
+				ResourceName:            resourceName,
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"ztp_file_download_path"},
+			},
+		},
+	})
+}
+
+func testAccEdgeMegaportBasic(edgeMegaportUsername, accountName, gwName, siteId, path string) string {
```
rpanzer-aviatrix: This test does not work for me and there are multiple validation errors.
It would be great if the tests would work.


After trying to fix the test configuration I saw this error:
```
time="2024-12-10T16:50:18+01:00" level=info msg="Found Aviatrix Account megaport-user-zaoc4"
    resource_aviatrix_edge_megaport_test.go:28: Step 1/2 error: Error running apply: exit status 1
        
        Error: Provider produced inconsistent result after apply
        
        When applying changes to aviatrix_edge_megaport.test, provider
        "provider[\"registry.terraform.io/hashicorp/aviatrix\"]" produced an
        unexpected new value: Root object was present, but now absent.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
```
I think even if I have an incompatible controller I should never see such an error.

goaviatrix/edge_megaport.go
```diff
@@ -0,0 +1,262 @@
...
+
+	outFile, err := os.Create(fileName)
+	if err != nil {
+		return err
+	}
+
+	_, err = outFile.WriteString(data.Result)
+	if err != nil {
+		return err
+	}
```
rpanzer-aviatrix, line 193: Is the user supposed to do sth with this file?
What happens if terraform runs in a CD pipeline?
Would it make sense to expose such things as computed properties?

rpanzer-aviatrix (commented):
It would be great to have the test working.

goaviatrix/edge_megaport.go
sairane10: The ztp file download path is where the cloud_init file is stored locally. This will not require marshaling.

docs/resources/aviatrix_edge_megaport.md
sairane10: Yes, I will update the description to specify gateway. The edge megaport follows the edge equinix code for gateway creation.

docs/resources/aviatrix_edge_megaport.md
sairane10: The ip_address follows the naming present for other edge interfaces like AEP and Equinix so I have kept it the same for Megaport.

docs/resources/aviatrix_edge_megaport.md
sairane10: vlan is an optional attribute. Updated the document.

goaviatrix/edge_megaport.go
sairane10: This is the cloud_init file which is required to bring up the Megaport Edge Gateway. The user's will have to onboard a megaport account, create MVE with internet connection enabled on the Megaport portal. Cloud_init file is required to successfully order an MVE which will be provided by the TF resource.

docs/resources/aviatrix_edge_megaport.md
sairane10: Updated vlan to optional field as it is not required in interfaces

goaviatrix/edge_megaport.go
sairane10: It was a design decision made by the edge_csp team to encode it to base64 so user can pass on the data as a single string. They do it for all the APIs that take bulk of data in a single argument for e.g vlan, nat etc

aviatrix/resource_aviatrix_edge_megaport_test.go
sairane10: I have updated the UT and you won't encounter the above error.

goaviatrix/edge_megaport.go
rpanzer-aviatrix: Without a json tag it is marshalled.
If you don't want it to be marshalled you would need to define `json:"-"`.
See https://go.dev/play/p/h9oC2qLDQ3d

goaviatrix/edge_megaport.go
rpanzer-aviatrix: I see that the same logic is used in edge_equinix, so it seems to be accepted.
I still believe that it would be better to expose the cloud-init contents as a computed property.
Then the user can still write it to a local file, or grab it from their CICD pipeline.

goaviatrix/edge_megaport.go
sairane10: Updated the json tag

rpanzer-aviatrix (approved):
LGTM
There is a bug currently in the controller so that it doesn't return all the interfaces registered with terraform. Therefore the plan is always dirty after apply. Once this gets fixed the test should also pass.

PR #2108, opened 2024-11-26
AVX-59033 Adding the BFD refresh functionality

sairane10:
- The bgp_bfd config was not not getting set in terraform. Updated the read back code to address this issue. Now if the bgp_bfd is disabled using the api then the TF code detects this change and updates the external resource connection.
Link - https://aviatrix.atlassian.net/browse/AVX-59033

### DIFF
--- a/aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
+++ b/aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
@@ -350,7 +350,7 @@ func resourceAviatrixEdgeSpokeExternalDeviceConnCreate(ctx context.Context, d *s
 			}
 		} else {
 			// set the bgp bfd config using the default values
-			externalDeviceConn.BgpBfdConfig = &defaultBfdConfig
+			externalDeviceConn.BgpBfdConfig = defaultBfdConfig
 		}
 		err := client.EditConnectionBgpBfd(externalDeviceConn)
 		if err != nil {
@@ -454,7 +454,7 @@ func resourceAviatrixEdgeSpokeExternalDeviceConnRead(ctx context.Context, d *sch
 		return diag.Errorf("expected enable_bfd to be a boolean, but got %T", d.Get("enable_bfd"))
 	}
 	d.Set("enable_bfd", enable_bfd)
-	if conn.EnableBfd && conn.BgpBfdConfig != nil {
+	if conn.EnableBfd {
 		var bgpBfdConfig []map[string]interface{}
 		bfd := conn.BgpBfdConfig
 		bfdMap := make(map[string]interface{})
@@ -541,7 +541,7 @@ func resourceAviatrixEdgeSpokeExternalDeviceConnUpdate(ctx context.Context, d *s
 				GwName:         d.Get("gw_name").(string),
 				ConnectionName: d.Get("connection_name").(string),
 				EnableBfd:      d.Get("enable_bfd").(bool),
-				BgpBfdConfig:   &bgpBfd,
+				BgpBfdConfig:   bgpBfd,
 			}
 			err := client.EditConnectionBgpBfd(externalDeviceConn)
 			if err != nil {
--- a/aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
+++ b/aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
@@ -448,12 +448,7 @@ func resourceAviatrixEdgeSpokeExternalDeviceConnRead(ctx context.Context, d *sch
 	if conn.BgpRemoteAsNum != 0 {
 		d.Set("bgp_remote_as_num", strconv.Itoa(conn.BgpRemoteAsNum))
 	}
-
-	enable_bfd, ok := d.Get("enable_bfd").(bool)
-	if !ok {
-		return diag.Errorf("expected enable_bfd to be a boolean, but got %T", d.Get("enable_bfd"))
-	}
-	d.Set("enable_bfd", enable_bfd)
+	d.Set("enable_bfd", conn.EnableBfd)
 	if conn.EnableBfd {
 		var bgpBfdConfig []map[string]interface{}
 		bfd := conn.BgpBfdConfig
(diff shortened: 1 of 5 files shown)


aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
```diff
@@ -448,19 +448,13 @@ func resourceAviatrixEdgeSpokeExternalDeviceConnRead(ctx context.Context, d *sch
...
 		d.Set("bgp_remote_as_num", strconv.Itoa(conn.BgpRemoteAsNum))
 	}
-
-	enable_bfd, ok := d.Get("enable_bfd").(bool)
-	if !ok {
-		return diag.Errorf("expected enable_bfd to be a boolean, but got %T", d.Get("enable_bfd"))
-	}
-	d.Set("enable_bfd", enable_bfd)
-	if conn.EnableBfd && conn.BgpBfdConfig != nil {
+	d.Set("enable_bfd", conn.EnableBfd)
```
stewartbutler, line 451: We just set this blindly to whatever the config is? 

If `d.Get("enable_bfd") ` would return true, but `conn.EnableBfd` is unset (default false), is the expected behavior to force it to false to match `conn.EnableBfd`?

aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
```diff
@@ -448,19 +448,13 @@ func resourceAviatrixEdgeSpokeExternalDeviceConnRead(ctx context.Context, d *sch
...
+	// set the bgp_bfd config details only if the user has enabled BFD and provided the config details. For default values, the config is not set
+	bgpBfdConfig := d.Get("bgp_bfd").([]interface{})
```
stewartbutler: I think you need to check this cast -- `if bgpBfdConfig, ok := d.Get("bgp_bfd").([]interface{}); !ok { <handle error> }`

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -986,21 +986,13 @@ func resourceAviatrixSpokeExternalDeviceConnRead(d *schema.ResourceData, meta in
...
-			return fmt.Errorf("expected enable_bfd to be a boolean, but got %T", d.Get("enable_bfd"))
-		}
-		if enable_bfd && conn.ConnectionType != "bgp" {
-			return fmt.Errorf("BFD is only supported for BGP connection type")
-		}
-		d.Set("enable_bfd", enable_bfd)
-		if conn.EnableBfd && conn.BgpBfdConfig != nil {
+		d.Set("enable_bfd", conn.EnableBfd)
+		// set the bgp_bfd config details only if the user has enabled BFD and provided the config details. For default values, the config is not set
+		bgpBfdConfig := d.Get("bgp_bfd").([]interface{})
```
stewartbutler: Cast check?

goaviatrix/transit_external_device_conn.go
```diff
@@ -377,7 +384,7 @@ func TransitExternalDeviceConnPh1RemoteIdDiffSuppressFunc(k, old, new string, d
 	return false
 }
 
-func CreateBgpBfdConfig(bfd map[string]interface{}) *BgpBfdConfig {
+func CreateBgpBfdConfig(bfd map[string]interface{}) BgpBfdConfig {
```
stewartbutler, line 387: Any reason we are switching away from pointer here?

stewartbutler (changes requested):

goaviatrix/transit_external_device_conn.go
```diff
@@ -65,14 +65,14 @@ type ExternalDeviceConn struct {
...
-	BgpMd5KeyChanged       bool          `form:"bgp_md5_key_changed,omitempty"`
-	BgpBfdConfig           *BgpBfdConfig `form:"bgp_bfd,omitempty"`
-	EnableBfd              bool          `form:"enable_bfd,omitempty"`
+	BgpMd5Key              string       `form:"bgp_md5_key,omitempty"`
+	BackupBgpMd5Key        string       `form:"backup_bgp_md5_key,omitempty"`
+	AuthType               string       `form:"auth_type,omitempty"`
+	EnableEdgeUnderlay     bool         `form:"edge_underlay,omitempty"`
+	RemoteCloudType        string       `form:"remote_cloud_type,omitempty"`
+	BgpMd5KeyChanged       bool         `form:"bgp_md5_key_changed,omitempty"`
+	BgpBfdConfig           BgpBfdConfig `form:"bgp_bfd_params,omitempty"`
```
stewartbutler, line 74: Will this break backward compat?

stewartbutler: The change of the form keys, specifically

aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -1088,21 +1088,13 @@ func resourceAviatrixTransitExternalDeviceConnRead(d *schema.ResourceData, meta
...
-			return fmt.Errorf("expected enable_bfd to be a boolean, but got %T", d.Get("enable_bfd"))
-		}
-		if conn.ConnectionType != "bgp" && enable_bfd {
-			return fmt.Errorf("BFD is only supported for BGP connection")
-		}
-		d.Set("enable_bfd", enable_bfd)
-		if enable_bfd && conn.BgpBfdConfig != nil {
+		d.Set("enable_bfd", conn.EnableBfd)
+		// set the bgp_bfd config details only if the user has enabled BFD and provided the config details. For default values, the config is not set
+		bgpBfdConfig := d.Get("bgp_bfd").([]interface{})
```
stewartbutler: Cast check needed

aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
sairane10: The api call returns the conn.EnableBfd which we want to set inorder to correctly reflect the correction status. Yes if the d.Get("enable_bfd") value does not match with what is being returned by conn.EnableBfd then an update is triggered by TF. We always want to set the value with conn.EnableBfd.

goaviatrix/transit_external_device_conn.go
sairane10: This is a newly added param for 7.2.b so it won't break the functionality. This PR is a bug resolution for bfd refresh state which was caused by incorrect referencing of the controller params.

goaviatrix/transit_external_device_conn.go
sairane10: Since we are using this as a read only value and creating a new structure based on the user provided input, so decided to remove the pointers.

aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
sairane10: Added the check

aviatrix/resource_aviatrix_spoke_external_device_conn.go
sairane10: done

aviatrix/resource_aviatrix_transit_external_device_conn.go
sairane10: done

stewartbutler (approved):

PR #2110, opened 2024-12-02
AVX-59225 Updating the check on bgp_bfd config when bfd is disabled

sairane10:
Updating the create function to throw an error when bfd is disabled and bgp-bfd config is provided for the following resources
* spoke_external_device_conn
* edge_spoke_external_device_conn
* transit_external_device_conn
Link - https://aviatrix.atlassian.net/browse/AVX-59225


inc1t3Ful (approved):
LGTM

PR #2114, opened 2024-12-03
AVX-59196 Updating interface config to omit null values

sairane10:
Interface validations were failing because of the default values present for interface configs. Omitting null values while setting the interface config resolve this error

```
Error: could not create Edge Platform: rest API create_edge_csp_gateway POST failed: [AVXERR-EDGE-0011] Error in EdgeGateway creation: [AVXERR-EDGE-0005] Parameter error: Invalid field subinterfaces in interface eth0.; vrrp_state and virtual_ip are supported only for LAN interfaces.; IP address should not be present for interface eth2 when DHCP is true.; Invalid field subinterfaces in interface eth1.; Spoke mode requires at least one WAN, and exactly one LAN interface.; Edge gateway has no management interface config.

resource "aviatrix_edge_platform" "edge_platform_4" {
    account_name = "edge_aep"
    gw_name = "aep-edge-5"
    site_id = "site-21"
    device_id = "187c781f-7dd4-4b54-8be5-a0b8b96aa616"
    gw_size = "small"
    management_egress_ip_prefix_list = [
        "145.40.88.208/29"
    ]
    local_as_number = "65448"
    wan_interface_names = [
        "eth0"
    ]
    lan_interface_names = [
        "eth1"
    ]
    management_interface_names = [
        "eth2"
    ]
    interfaces {
        name = "eth2"
        type = "MANAGEMENT"
        ip_address = null
        enable_dhcp = true
    }

    interfaces {
        name = "eth1"
        type = "LAN"
        ip_address = "192.168.12.31/24"
    }

    interfaces {
        name = "eth0"
        type = "WAN"
        ip_address = "192.168.20.11/24"
        wan_public_ip = "145.40.88.210"
        gateway_ip = "192.168.20.1"
    }
}
```


aviatrix/resource_aviatrix_edge_platform.go
```diff
@@ -153,7 +153,7 @@ func resourceAviatrixEdgePlatform() *schema.Resource {
 			"bgp_neighbor_status_polling_time": {
 				Type:         schema.TypeInt,
 				Optional:     true,
-				Default:      defaultBgpPollingTime,
+				Default:      defaultBgpNeighborStatusPollingTime,
```
stewartbutler, line 156: This drops from 50 to 5 -- is that intentional? Pretty big jump.

stewartbutler (approved):
LGTM, just sanity checking the one thing.

avx-rchung (approved):

aviatrix/resource_aviatrix_edge_platform.go
sairane10: The bgp_neighbor_status_polling_time should be between 1 to 10. The default value indicated here was incorrect so updated it.

PR #2116, opened 2024-12-04
AVX-59353 Updating the state for default bgp_bfd config

sairane10:
Setting the state for bgp_bfd config if default values are used. If enable_bfd is set to true then bgp_bfd state will always be set for external connection.

Updates are made to the following resources
* aviatrix_edge_spoke_external_conn
* aviatrix_spoke_external_conn
* aviatrix_transit_external_conn

**Link** - https://aviatrix.atlassian.net/browse/AVX-59353


inc1t3Ful:
Is the expectation that `bfd_config {}` will only be set if enable_bfd is enabled? 
If `enable_bfd` is false/unset/default, what is the expected behavior?

sairane10:
@inc1t3Ful Yes bgp_bfd config should only be set if enable_bfd is set to true. For all other values of enable_bfd, the config will not be set.

inc1t3Ful (approved):

PR #2118, opened 2024-12-05
AVX-59353 Updating the bgp_bfd config state

sairane10:
- Adding the DiffSuppressFunc to bgp_bfd config. If enable_bfd is set to true and bgp_bfd config is set to default values then no resource change will be detected while reapplying the resource.
The default bgp_bfd values are stored in the state file to track changes. 
**Link** - https://aviatrix.atlassian.net/browse/AVX-59353


tommy-carver (commented):
Can we write a unit test for this?

sairane10:
@tommy-carver added the UT

tommy-carver (approved):

PR #2123, opened 2024-12-11
AVX-59753 Adding checks for bgp_neighbor_status_polling_time for backward compatibilty

sairane10:
- Adding the check for bgp_neighbor_status_polling_time to avoid api call from older controller version and latest TF provider version. 
- The api call will only take place if the bgp_neighbor_status_polling_time is set in the resource to a non-default value.
- Adding the diff suppress function for bgp_neighbor_status_polling_time where value of 0 is set for the parameter in the api. This will allow backward compatibility with the older 7.2 versions of the controller where the api call does not exist.

### DIFF
--- a/goaviatrix/spoke_vpc.go
+++ b/goaviatrix/spoke_vpc.go
@@ -3,8 +3,6 @@ package goaviatrix
 import (
 	"fmt"
 	"strings"
-
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 // Spoke gateway simple struct to hold spoke details
@@ -514,9 +512,4 @@ func (c *Client) DisableGlobalVpc(gateway *Gateway) error {
 		"gateway_name": gateway.GwName,
 	}
 	return c.PostAPI(form["action"], form, BasicCheck)
-}
-
-// diff suppress function for bgp_neighbor_status_polling_time. Suppress the diff if old value is 0
-func DiffSuppressFuncBgpNeighborStatusPollingTime(k, old, new string, d *schema.ResourceData) bool {
-	return old == "0"
 }
--- a/goaviatrix/spoke_vpc.go
+++ b/goaviatrix/spoke_vpc.go
@@ -3,6 +3,8 @@ package goaviatrix
 import (
 	"fmt"
 	"strings"
+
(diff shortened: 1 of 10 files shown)


tommy-carver (commented):
Few questions: 
1. What happens here if the desired bgp status polling time is the default. Do we never set that? Or is the API on the controller automatically set the default and the client function here only sets non-default values.
2. Is it possible to add tests for this.

sairane10:
@tommy-carver, to answer your question
1. Yes the defaults are set by the api in the controller. The TF client is only looking to set non default values for bgp_neighbor_status_polling_time
2. The UT for this param already exists. However the version change of the controller calling the invalid api cannot be mocked here.

goaviatrix/spoke_vpc.go
```diff
@@ -513,3 +515,8 @@ func (c *Client) DisableGlobalVpc(gateway *Gateway) error {
 	}
 	return c.PostAPI(form["action"], form, BasicCheck)
 }
+
+// diff suppress function for bgp_neighbor_status_polling_time. Suppress the diff if old value is 0
+func DiffSuppressFuncBgpNeighborStatusPollingTime(k, old, new string, d *schema.ResourceData) bool {
```
tommy-carver: Can we just write an inline function for this? Seems overkill to have this function.

sairane10: sure will add that change.

tommy-carver (commented):
@sairane10 I dont mean tests to verify the controller backwards compat, cause that is difficult. I mean tests that verify that

```bgp_neighbor_status_polling_time := val.(int)
			if bgp_neighbor_status_polling_time >= 1 && bgp_neighbor_status_polling_time != defaultBgpNeighborStatusPollingTime {
				err := client.SetBgpBfdPollingTime(gateway, bgp_neighbor_status_polling_time)
				if err != nil {
					return fmt.Errorf("could not set bgp neighbor status polling time: %v", err)
				}
			}
```
does the right thing, aka call the API only when we actually need to. I dont know if this is possible, but if it is it would be nice to detect that this change actually works outside of manual testing.

sairane10:
@tommy-carver I have added the UT for this change

tommy-carver (approved):

PR #2130, opened 2025-01-06
AVX-57244 Adding support for EAT Equinix

sairane10:
- Adding the ztp_file_download_path param to the transit gateway where the cloud init file will be stored locally.
- Sorting the eip_map keys in the setEipMapDetails function
- Adding UT for Equinix EAT

### DIFF
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -4078,9 +4078,12 @@ func getEipMapDetails(eipMap []interface{}, wanCount int) (string, error) {
 
 func getTransitHaGatewayDetails(d *schema.ResourceData, wanCount int, cloudType int) (*goaviatrix.TransitHaGateway, error) {
 	transitHaGw := &goaviatrix.TransitHaGateway{
-		PrimaryGwName: d.Get("gw_name").(string),
-		GwName:        d.Get("gw_name").(string) + "-hagw",
-		InsaneMode:    "yes",
+		PrimaryGwName:       d.Get("gw_name").(string),
+		GwName:              d.Get("gw_name").(string) + "-hagw",
+		VpcID:               d.Get("vpc_id").(string),
+		ZtpFileDownloadPath: d.Get("ztp_file_download_path").(string),
+		CloudType:           d.Get("cloud_type").(int),
+		InsaneMode:          "yes",
 	}
 	peerBackupPortType, ok := d.GetOk("peer_backup_port_type")
 	if !ok {
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -107,7 +107,7 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 			"ztp_file_download_path": {
 				Type:     schema.TypeString,
 				Optional: true,
-				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+				DiffSuppressFunc: func(_, old, _ string, _ *schema.ResourceData) bool {
 					return old != ""
 				},
 				Description: "The location where the ZTP file will be stored locally.",
@@ -1855,7 +1855,7 @@ func resourceAviatrixTransitGatewayRead(d *schema.ResourceData, meta interface{}
 		d.Set("bgp_lan_ip_list", nil)
 		d.Set("ha_bgp_lan_ip_list", nil)
 		if gw.DeviceID != "" {
-			d.Set("device_id", gw.DeviceID)
+			_ = d.Set("device_id", gw.DeviceID)
 		}
 		// set interfaces
 		if len(gw.Interfaces) != 0 {
@@ -3813,9 +3813,17 @@ func resourceAviatrixTransitGatewayDelete(d *schema.ResourceData, meta interface
 			backoff *= 2
 		}
 		if cloudType == goaviatrix.EDGEEQUINIX {
-			err := deleteZtpFile(gateway.GwName, d.Get("vpc_id").(string), d.Get("ztp_file_download_path").(string))
+			vpcID, ok := d.Get("vpc_id").(string)
+			if !ok {
+				return fmt.Errorf("vpc_id is not a string")
+			}
+			ztpFileDownloadPath, ok := d.Get("ztp_file_download_path").(string)
+			if !ok {
+				return fmt.Errorf("ztp_file_download_path is not a string")
+			}
+			err := deleteZtpFile(gateway.GwName, vpcID, ztpFileDownloadPath)
 			if err != nil {
-				return fmt.Errorf("failed to delete ZTP file: %s", err)
+				return fmt.Errorf("failed to delete ZTP file: %w", err)
 			}
 		}
 	}
@@ -3825,24 +3833,32 @@ func resourceAviatrixTransitGatewayDelete(d *schema.ResourceData, meta interface
 		return fmt.Errorf("failed to delete Aviatrix Edge Transit Gateway: %s", err)
 	}
 	if cloudType == goaviatrix.EDGEEQUINIX {
-		err := deleteZtpFile(gateway.GwName, d.Get("vpc_id").(string), d.Get("ztp_file_download_path").(string))
+		vpcID, ok := d.Get("vpc_id").(string)
+		if !ok {
+			return fmt.Errorf("vpc_id is not a string")
+		}
+		ztpFileDownloadPath, ok := d.Get("ztp_file_download_path").(string)
+		if !ok {
+			return fmt.Errorf("ztp_file_download_path is not a string")
+		}
+		err := deleteZtpFile(gateway.GwName, vpcID, ztpFileDownloadPath)
 		if err != nil {
-			return fmt.Errorf("failed to delete ZTP file: %s", err)
+			return fmt.Errorf("failed to delete ZTP file: %w", err)
 		}
 	}
 
 	return nil
 }
 
-func deleteZtpFile(gatewayName, vpcId, ztpFileDownloadPath string) error {
-	fileName := ztpFileDownloadPath + "/" + gatewayName + "-" + vpcId + "-cloud-init.txt"
+func deleteZtpFile(gatewayName, vpcID, ztpFileDownloadPath string) error {
+	fileName := ztpFileDownloadPath + "/" + gatewayName + "-" + vpcID + "-cloud-init.txt"
 	// check if the file exists
 	if _, err := os.Stat(fileName); os.IsNotExist(err) {
 		return nil
 	}
 	err := os.Remove(fileName)
 	if err != nil {
-		return fmt.Errorf("could not remove the ztp file: %v", err)
+		return fmt.Errorf("could not remove the ztp file: %w", err)
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -3798,24 +3812,44 @@ func resourceAviatrixTransitGatewayDelete(d *schema.ResourceData, meta interface
...
+		}
+	}
 
 	return nil
 }
 
+func deleteZtpFile(gatewayName, vpcId, ztpFileDownloadPath string) error {
+	fileName := ztpFileDownloadPath + "/" + gatewayName + "-" + vpcId + "-cloud-init.txt"
+	err := os.Remove(fileName)
+	if err != nil {
```
stewartbutler: Might want to change this to something like
```suggestion
	if err != nil && !errors.Is(err, fs.ErrNotExist) {
```

That way, if the file is already deleted, the function will still return without error.

stewartbutler: Also, I'd add some unit tests for this function to confirm that the behavior is sane.

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -4293,14 +4338,25 @@ func setInterfaceMappingDetails(interfaceMapping []goaviatrix.InterfaceMapping)
 
 func setEipMapDetails(eipMap map[string][]goaviatrix.EipMap, ifNameTranslation map[string]string) ([]map[string]interface{}, error) {
 	eipMapList := make([]map[string]interface{}, 0)
-	for interfaceName, eipList := range eipMap {
-		// get the interface type and index from the interface name
+	keys := make([]string, 0, len(eipMap))
```
stewartbutler, line 4366: Needs test coverage for this function

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -4084,6 +4127,8 @@ func getTransitHaGatewayDetails(d *schema.ResourceData, wanCount int, cloudType
 			return nil, fmt.Errorf("failed to get the interface mapping details: %v", err)
 		}
 		transitHaGw.InterfaceMapping = interfaceMapping
+		// set device_id for AEP ha gateway
+		transitHaGw.DeviceID = d.Get("ha_device_id").(string)
```
stewartbutler: Check the cast result

goaviatrix/transit_ha_gateway.go
```diff
@@ -41,6 +46,23 @@ type BackupLinkInterface struct {
...
 	transitHaGateway.Action = "create_multicloud_ha_gateway"
+	var data CreateEdgeEquinixResp
+	resp, err := c.PostAPIContext2HaGw(context.Background(), nil, transitHaGateway.Action, transitHaGateway, BasicCheck)
+	if err != nil {
+		return "", err
+	}
+	// create the ZTP file for Equinix Edge transit gateway
+	if transitHaGateway.CloudType == EDGEEQUINIX {
+		fileName := transitHaGateway.ZtpFileDownloadPath + "/" + transitHaGateway.GwName + "-" + transitHaGateway.VpcID + "-cloud-init.txt"
+		outFile, err := os.Create(fileName)
```
stewartbutler: This will truncate the file if it already exists; is that fine?

goaviatrix/transit_vpc.go
```diff
@@ -173,7 +176,26 @@ type TransitGatewayBgpLanIpInfo struct {
...
+	var data CreateEdgeEquinixResp
+	err := c.PostAPIContext2(context.Background(), &data, gateway.Action, gateway, BasicCheck)
+	if err != nil {
+		return err
+	}
+	// create the ZTP file for Equinix edge transit gateway
+	if gateway.CloudType == EDGEEQUINIX {
+		fileName := gateway.ZtpFileDownloadPath + "/" + gateway.GwName + "-" + gateway.VpcID + "-cloud-init.txt"
+
+		outFile, err := os.Create(fileName)
```
stewartbutler: This will truncate the file if it already exists; is that fine?

goaviatrix/transit_ha_gateway.go
stewartbutler: Also, you use this pattern in a few places. Why not create a helper function and test that, then just use it here?

stewartbutler (changes requested):

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: The UT for this exists in the test file 
`TestSetEipMapDetails`

goaviatrix/transit_ha_gateway.go
sairane10: Updated the code. It will delete any a file if it exists with this name and create a new one.

aviatrix/resource_aviatrix_transit_gateway.go
sairane10: Updated

goaviatrix/transit_vpc.go
sairane10: Modified the code to use a helper function and not truncate if a file already exists

aviatrix/resource_aviatrix_transit_gateway_test.go
```diff
@@ -1050,3 +1186,33 @@ func TestSetInterfaceDetails(t *testing.T) {
...
+	gatewayName := "test-gateway"
+	vpcId := "vpc-123456"
+	ztpFileDownloadPath := tempDir
+	fileName := filepath.Join(ztpFileDownloadPath, fmt.Sprintf("%s-%s-cloud-init.txt", gatewayName, vpcId))
+	// Create a temporary file to simulate the ztp file
+	if err := os.WriteFile(fileName, []byte("test content"), 0644); err != nil {
+		t.Fatalf("Failed to create temporary file: %v", err)
+	}
+
+	// Ensure the file exists before calling the function
```
stewartbutler, line 1205: So the issue here is that it is unclear what you expect the behavior of this function to be. There are two options:

1. If a ZTP file is missing and you run DeleteZtpFile, it fails and returns an error
2. If a ZTP file is missing and you run DeleteZtpFile, it detects that the file is already deleted and returns early with no error.

What's your intent? Because right now you are testing case 1, which may cause an unexpected failure if the user deleted the ztp file out of band, or it doesn't already exist.

goaviatrix/transit_vpc.go
```diff
@@ -734,3 +728,35 @@ func (c *Client) DisableTransitPreserveAsPath(transitGateway *TransitVpc) error
...
+		if err := os.Remove(filePath); err != nil {
+			return fmt.Errorf("failed to delete the existing file: %v", err)
+		}
+	} else if !os.IsNotExist(err) {
+		// Some other error occurred while checking for the file
+		return fmt.Errorf("failed to check file existence: %v", err)
+	}
+
+	// Create a new file
+	outFile, err := os.Create(filePath)
```
stewartbutler, line 753: As I stated, truncation is fine so long as you know what that behavior is going to be. Running stat and removing the file is unnecessary.

goaviatrix/transit_vpc.go
```diff
@@ -734,3 +728,35 @@ func (c *Client) DisableTransitPreserveAsPath(transitGateway *TransitVpc) error
 	}
 	return c.PostAPI(action, data, BasicCheck)
 }
+
+// createOrReplaceFile deletes a file if it exists, then creates a new one and writes the given content.
+func createOrReplaceFile(filePath, content string) error {
```
stewartbutler: This function needs tests

stewartbutler (changes requested):
Few more comments

aviatrix/resource_aviatrix_transit_gateway_test.go
sairane10: The expected behavior is that if the DeleteZtpFile function is called then it should delete the the file if exists. If there is no such file to delete then the function should return without any error.

goaviatrix/transit_vpc.go
sairane10: In cloud-init file will be created and used only during the transit creation call. I have updated the code to remove the rest of the checks.

goaviatrix/transit_vpc.go
sairane10: Added UT for this function

aviatrix/resource_aviatrix_transit_gateway_test.go
stewartbutler: That's not what you have implemented, though -- if the file does not exist, it currently throws an error. See https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2130/files#diff-c4c3016716f5fe166ee12877e0e39e7edd4234ee70bd3b244d17acf9182f43d5R3859 -- if you run `os.Remove(filename)` on a nonexistent file, it returns ENOENT as an error. There was a previous conversation to that effect.

stewartbutler: Try it in the test, you should see what I mean.

goaviatrix/utils_test.go
```diff
@@ -62,3 +63,108 @@ func TestMapContains(t *testing.T) {
...
+			filePath:    "/invalid/path/test-file.txt",
+			content:     "This content should not be written.",
+			expectedErr: "failed to create the file",
+		},
+		{
+			name:        "Error writing to file (empty content)",
+			filePath:    "test-file.txt",
+			content:     "",
+			expectedErr: "",
+		},
```
stewartbutler, line 91: Add a test case for file truncation, just to make sure it behaves as you expect. Precreate some test file with some content, create a ztp file with same path with new content, verify the file has the new content.

aviatrix/resource_aviatrix_transit_gateway_test.go
```diff
@@ -98,6 +99,7 @@ func TestAccAviatrixTransitGateway_basic(t *testing.T) {
 	skipGwGCP := os.Getenv("SKIP_TRANSIT_GATEWAY_GCP")
 	skipGwOCI := os.Getenv("SKIP_TRANSIT_GATEWAY_OCI")
 	skipGwAEP := os.Getenv("SKIP_TRANSIT_GATEWAY_AEP")
+	skipGwEQUINIX := os.Getenv("SKIP_TRANSIT_GATEWAY_EQUINIX")
```
stewartbutler, line 102: Not a problem in this PR, but we need to revisit this pattern. A better way of handling this down the line is:

- Create a set of helper functions that take a `testing.T` input, and skip if a condition is true.
- Use `t.Run()` for each of these edge cases, and use the helper function in each subtest to skip that particular subtest if we have disabled it
- Split integration and unit tests into separate files and run integration tests only when a particular golang flag is set

aviatrix/resource_aviatrix_transit_gateway.go
stewartbutler: Still not resolved -- based on earlier comment, you still need to check for `ErrNotExist` if you want this to work as you expect. `os.Remove` will return `ErrNotExist` (`syscall.ENOENT`) if the file does not exist.

aviatrix/resource_aviatrix_transit_gateway_test.go
sairane10: Added the changes to the deleteZtpFile and units test. The UT has a test case where it tries to delete an non existant file and the function does not throw an error

aviatrix/resource_aviatrix_transit_gateway_test.go
sairane10: Yeah I will handle this better in the coming PR's

goaviatrix/utils_test.go
sairane10: Added a new truncation test case here

sairane10:
@stewartbutler I have resolved most of the comments except the reformatting of existing unit tests and integration test using a helper function. Will add that change in the upcoming PR's

stewartbutler (approved):

PR #2141, opened 2025-01-16
AVX-60611: Remove the aviatrix_dns_profile resource

jpeach:
The upstream DNS profile feature never reached a production release and is now being removed. Remove the corresponding "aviatrix_dns_profile" since no production users should have deployed this. Deprecate the "dns_profile_name" field for edge resources so that we don't break and edge configurations that may have inadvertantly set this (ineffective) field.


dnardo (approved):
Thanks James, do you mind backporting this to the 8.0 branch also

PR #2160, opened 2025-02-13
AVX-58676  Adding support for Edge Transit Gateway Peering

sairane10:
Adding the following changes for Megaport peerings

Types of peerings

1. CSP Transit to CSP Transit
2. EAT to EAT 
3. EAT to CSP transit 
4. CSP transit to EAT 

Newly added attributes to schema:
- gateway1_logical_ifnames
- gateway2_logical_ifnames

These attributes are same for all edge gateways and will replace src_wan_interfaces and dst_wan_interfaces which were planned for 8.0 release

```
resource "aviatrix_transit_gateway_peering" "test_transit_gateway_peering" {
  transit_gateway_name1                       = "edge-transit-megaport-2"
  transit_gateway_name2                       = "edge-transit-megaport-3"
  enable_peering_over_private_network = true
  jumbo_frame = false
  insane_mode = true
  gateway1_logical_ifnames = ["wan1"]
  gateway2_logical_ifnames = ["wan1"]
}

resource "aviatrix_transit_gateway_peering" "test_transit_gateway_peering_2" {
  transit_gateway_name1                       = "edge-transit-megaport-2"
  transit_gateway_name2                       = "aws-transit"
  enable_peering_over_private_network = true
  jumbo_frame = false
  insane_mode = true
  gatew
```ay1_logical_ifnames = ["wan1"]
}

### CODE REVIEW

File: aviatrix/resource_aviatrix_transit_gateway_peering.go [hunk 1]
```diff
@@ -89,7 +90,6 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
-				ForceNew: true,
```
stewartbutler: Why removed, just unused? Will this cause breakage for existing customers?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -136,12 +136,6 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
 					"Only valid when the two transit gateways are each launched in Insane Mode and in the same cloud type. " +
 					"Available as of provider version R2.22.2+.",
 			},
-			"over_private_network": {
```
stewartbutler, line 139: Same here, why are we removing this? Will this cause breakage for existing customers?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -154,15 +148,21 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
 				Default:     false,
 				Description: "Enable HPE mode for peering with Edge Transit",
 			},
-			"src_wan_interfaces": {
-				Type:        schema.TypeString,
+			"gateway1_logical_ifnames": {
```
stewartbutler, line 153: Why the name change from `src/dst_wan_interfaces` to `gatewayx_logical_ifnames`? Will renaming this field cause breakage in existing deployments?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -171,126 +171,107 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
...
 
-		var gw1Tgws []string
-		for _, tgw := range d.Get("gateway1_excluded_tgw_connections").([]interface{}) {
-			gw1Tgws = append(gw1Tgws, tgw.(string))
-		}
-		var gw2Tgws []string
-		for _, tgw := range d.Get("gateway2_excluded_tgw_connections").([]interface{}) {
-			gw2Tgws = append(gw2Tgws, tgw.(string))
-		}
+	transitGatewayPeering.EnableOverPrivateNetwork, ok = d.Get("enable_peering_over_private_network").(bool)
```
stewartbutler, line 191: Default false is OK for all of these?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+	var ok bool
+	for i, key := range keys {
+		if results[i], ok = d.Get(key).(bool); !ok {
+			return nil, fmt.Errorf("%s is required and must be a boolean", key)
+		}
+	}
+	return results, nil
+}
+
+func setExcludedResources(d *schema.ResourceData, transitGatewayPeering *goaviatrix.TransitGatewayPeering) error {
```
stewartbutler: Test for this?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+	gw2Tgws, err := getStringListFromResource(d, "gateway2_excluded_tgw_connections")
+	if err != nil {
+		return err
+	}
+	transitGatewayPeering.Gateway2ExcludedTGWConnections = strings.Join(gw2Tgws, ",")
+
+	return nil
+}
+
+func getGatewayDetails(client *goaviatrix.Client, gatewayName string) (*goaviatrix.Gateway, error) {
```
stewartbutler: Quick test?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+		// Call the client method to edit the transit connection.
+		err = client.EditTransitConnectionASPathPrepend(transGwPeering, prependASPath)
+		if err != nil {
+			return fmt.Errorf("could not set %s: %w", prependField, err)
+		}
+	}
+	return nil
+}
+
+func setNonEATPeeringOptions(d *schema.ResourceData, transitGatewayPeering *goaviatrix.TransitGatewayPeering) error {
```
stewartbutler: This logic is pretty fragile (referencing by index value), test here to make sure it is consistent in case someone changes any of these bool values? Also there is some odd logic here, I'd like a test with some decent coverage to make sure it is behaving according to some contract.

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+	return nil
+}
+
+// Helper function to validate Single Tunnel Mode
+func validateSingleTunnelMode(transitGatewayPeering *goaviatrix.TransitGatewayPeering, singleTunnelMode bool) error {
+	if singleTunnelMode {
+		if transitGatewayPeering.PrivateIPPeering == "no" {
+			return fmt.Errorf("enable_single_tunnel_mode is only valid when enable_peering_over_private_network is set to true")
+		}
+		transitGatewayPeering.SingleTunnel = "yes"
```
stewartbutler: This looks like it causes a side effect -- are we sure that is what you want? Validate functions are usually read-only.

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+		transitGatewayPeering.PrivateIPPeering = "yes"
+		if insaneMode {
+			return fmt.Errorf("enable_peering_over_private_network conflicts with enable_insane_mode_encryption_over_internet")
+		}
+	}
+	return nil
+}
+
+// Helper function to validate Single Tunnel Mode
+func validateSingleTunnelMode(transitGatewayPeering *goaviatrix.TransitGatewayPeering, singleTunnelMode bool) error {
```
stewartbutler: Test for this, please, with before-after sanity checks on the tested objects to make sure there aren't side effects.

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+		if transitGatewayPeering.PrivateIPPeering == "no" {
+			return fmt.Errorf("enable_single_tunnel_mode is only valid when enable_peering_over_private_network is set to true")
+		}
+		transitGatewayPeering.SingleTunnel = "yes"
+	}
+	return nil
+}
+
+// Helper function to validate Tunnel Count
+func validateTunnelCount(insaneMode bool, tunnelCount int) error {
```
stewartbutler: Test here as well in case someone edits it to have a side effect like the above.

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+	}
+
+	return nil
+}
+
+// validate Private IP Peering
+func validatePrivateIPPeering(transitGatewayPeering *goaviatrix.TransitGatewayPeering, peeringOverPrivate, insaneMode bool) error {
+	transitGatewayPeering.PrivateIPPeering = "no"
+	if peeringOverPrivate {
+		transitGatewayPeering.PrivateIPPeering = "yes"
```
stewartbutler: This sets a value; validate checks are usually read-only. Are the side effects intentional? If so, I'd change the name.

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+	}
+	if tunnelCount != 0 {
+		transitGatewayPeering.TunnelCount = tunnelCount
+	}
+
+	return nil
+}
+
+// validate Private IP Peering
+func validatePrivateIPPeering(transitGatewayPeering *goaviatrix.TransitGatewayPeering, peeringOverPrivate, insaneMode bool) error {
```
stewartbutler: Same as below, I don't trust this; please add a test with a before/after sanity check.

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+func setNonEATPeeringOptions(d *schema.ResourceData, transitGatewayPeering *goaviatrix.TransitGatewayPeering) error {
+	// Validate and set boolean fields
+	boolValues, err := getBooleanValues(d, "enable_max_performance", "enable_insane_mode_encryption_over_internet", "enable_peering_over_private_network", "enable_single_tunnel_mode")
+	if err != nil {
+		return err
+	}
+	noMaxPerformance, insaneMode, peeringOverPrivate, singleTunnelMode := boolValues[0], boolValues[1], boolValues[2], boolValues[3]
+
+	// Set Max Performance and Insane Mode
+	transitGatewayPeering.NoMaxPerformance = !noMaxPerformance
```
stewartbutler: Read a value and then flip it? Why?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+
+func setNonEATPeeringOptions(d *schema.ResourceData, transitGatewayPeering *goaviatrix.TransitGatewayPeering) error {
+	// Validate and set boolean fields
+	boolValues, err := getBooleanValues(d, "enable_max_performance", "enable_insane_mode_encryption_over_internet", "enable_peering_over_private_network", "enable_single_tunnel_mode")
+	if err != nil {
+		return err
+	}
+	noMaxPerformance, insaneMode, peeringOverPrivate, singleTunnelMode := boolValues[0], boolValues[1], boolValues[2], boolValues[3]
+
+	// Set Max Performance and Insane Mode
```
stewartbutler: Why?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+		strValue, ok := v.(string)
+		if !ok {
+			return nil, fmt.Errorf("%s contains non-string elements", key)
+		}
+		result = append(result, strValue)
+	}
+	return result, nil
+}
+
+func setASPathPrepend(d *schema.ResourceData, client *goaviatrix.Client, prependField, transitGatewayName1, transitGatewayName2 string) error {
```
stewartbutler: What is this doing?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+		}
+
+		// Set up the Transit Gateway Peering struct with the appropriate names.
+		transGwPeering := &goaviatrix.TransitGatewayPeering{
+			TransitGatewayName1: transitGatewayName1,
+			TransitGatewayName2: transitGatewayName2,
+		}
+
+		// Call the client method to edit the transit connection.
+		err = client.EditTransitConnectionASPathPrepend(transGwPeering, prependASPath)
```
stewartbutler: Why prepend rather than append?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+
+func convertToInterfaceSlice(strs []string) []interface{} {
+	result := make([]interface{}, len(strs))
+	for i, v := range strs {
+		result[i] = v
+	}
+	return result
+}
+
+func getStringListFromResource(d *schema.ResourceData, key string) ([]string, error) {
```
stewartbutler: Why a list of values? Are keys not needed? Why return a []string rather than a map?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+	cloudType int,
+	gatewayDetails *goaviatrix.Gateway,
+	gatewayPrefix string,
+	transitGatewayPeering *goaviatrix.TransitGatewayPeering,
+) error {
+	if len(logicalIfNames) == 0 {
+		return nil
+	}
+
+	if goaviatrix.IsCloudType(cloudType, goaviatrix.EDGEEQUINIX|goaviatrix.EDGENEO) {
```
stewartbutler: I'd change this to a switch, to make it clearer and easier to maintain. Also make a comment about using bitwise or here, and that cloudType is a bit mask.
```
switch {
case goaviatrix.IsCloudType(cloudType, goaviatrix.EDGEEQUINIX|goaviatrix.EDGENEO):
   something...
   return nil
case goaviatrix.IsCloudType(cloudType, goaviatrix.EDGEMEGAPORT):
   something...
   return nil
default:
    default behavior
    return nil
}
```

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
...
+		// Process logical interface names for Equinix/NEO cloud types
+		reversedInterfaceNames := ReverseIfnameTranslation(gatewayDetails.IfNamesTranslation)
+		wanInterfacesStr, err := SetWanInterfaces(convertToInterfaceSlice(logicalIfNames), reversedInterfaceNames)
+		if err != nil {
+			return fmt.Errorf("failed to set %s WAN interfaces to create edge peering: %w", gatewayPrefix, err)
+		}
+
+		if gatewayPrefix == "gateway1" {
+			transitGatewayPeering.SrcWanInterfaces = wanInterfacesStr
```
stewartbutler: Why do this weird renaming shuffle if the underlying object still uses Src/Dst?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -171,126 +171,107 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
...
+	if !ok {
+		return fmt.Errorf("enable_peering_over_private_network is required for edge gateway peering")
+	}
+	transitGatewayPeering.EnableJumboFrame, ok = d.Get("jumbo_frame").(bool)
+	if !ok {
+		return fmt.Errorf("jumbo_frame is required for edge gateway peering")
+	}
+	transitGatewayPeering.EnableInsaneMode, ok = d.Get("insane_mode").(bool)
+	if !ok {
+		return fmt.Errorf("insane_mode is required for edge gateway peering")
```
stewartbutler, line 201: Then why is it an option?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -171,126 +171,107 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
...
-		for _, tgw := range d.Get("gateway2_excluded_tgw_connections").([]interface{}) {
-			gw2Tgws = append(gw2Tgws, tgw.(string))
-		}
+	transitGatewayPeering.EnableOverPrivateNetwork, ok = d.Get("enable_peering_over_private_network").(bool)
+	if !ok {
+		return fmt.Errorf("enable_peering_over_private_network is required for edge gateway peering")
+	}
+	transitGatewayPeering.EnableJumboFrame, ok = d.Get("jumbo_frame").(bool)
+	if !ok {
+		return fmt.Errorf("jumbo_frame is required for edge gateway peering")
```
stewartbutler, line 197: Then why is it an option?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -171,126 +171,107 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
...
-		defer resourceAviatrixTransitGatewayPeeringReadIfRequired(d, meta, &flag)
+	if err := setExcludedResources(d, transitGatewayPeering); err != nil {
+		return err
+	}
 
-		err := client.CreateTransitGatewayPeering(transitGatewayPeering)
-		if err != nil {
-			return fmt.Errorf("failed to create Aviatrix Transit Gateway peering: %s", err)
+	// options only supported for non EAT peerings
+	if !goaviatrix.IsCloudType(gateway1CloudType, goaviatrix.EdgeRelatedCloudTypes) && !goaviatrix.IsCloudType(gateway2CloudType, goaviatrix.EdgeRelatedCloudTypes) {
```
stewartbutler: Make this condition a helper function, so people don't have to think so hard about the compounded conditionals

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -171,126 +171,107 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
...
-			}
+	log.Printf("[INFO] Creating Aviatrix Transit Gateway peering: %#v", transitGatewayPeering)
+	d.SetId(transitGatewayPeering.TransitGatewayName1 + "~" + transitGatewayPeering.TransitGatewayName2)
+	defer func() {
+		if err := resourceAviatrixTransitGatewayPeeringReadIfRequired(d, meta, &flag); err != nil {
+			log.Printf("[ERROR] Failed to read Aviatrix Transit Gateway peering: %v", err)
 		}
+	}()
+
+	err = client.CreateTransitGatewayPeering(context.Background(), transitGatewayPeering)
```
stewartbutler: is context.Background appropriate here, or should this have a timeout?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -171,126 +171,107 @@ func resourceAviatrixTransitGatewayPeering() *schema.Resource {
...
-			var prependASPath []string
-			for _, v := range d.Get("prepend_as_path2").([]interface{}) {
-				prependASPath = append(prependASPath, v.(string))
-			}
-			transGwPeering := &goaviatrix.TransitGatewayPeering{
-				TransitGatewayName1: d.Get("transit_gateway_name2").(string),
-				TransitGatewayName2: d.Get("transit_gateway_name1").(string),
-			}
+	// Set AS Path prepend for gateway1 (prepend_as_path1)
+	if err := setASPathPrepend(d, client, "prepend_as_path1", transitGatewayName1, transitGatewayName2); err != nil {
```
stewartbutler, line 273: Why prepend?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -336,72 +317,86 @@ func resourceAviatrixTransitGatewayPeeringRead(d *schema.ResourceData, meta inte
...
-		if transitGatewayPeering.InsaneModeOverInternet {
-			d.Set("tunnel_count", transitGatewayPeering.TunnelCount)
+			return fmt.Errorf("could not set prepend_as_path2: %w", err)
 		}
+	}
+	if err := d.Set("enable_peering_over_private_network", transitGatewayPeering.PrivateIPPeering == "yes"); err != nil {
+		return fmt.Errorf("failed to set enable_peering_over_private_network: %w", err)
+	}
+
+	enableSingleTunnel := transitGatewayPeering.PrivateIPPeering == "yes" && transitGatewayPeering.SingleTunnel == "yes"
```
stewartbutler, line 388: What happens if SingleTunnel is yes but PrivateIPPeering is not? Is that error handled?

aviatrix/resource_aviatrix_transit_gateway_peering.go
```diff
@@ -497,3 +496,244 @@ func resourceAviatrixTransitGatewayPeeringDelete(d *schema.ResourceData, meta in
 
 	return nil
 }
+
```
stewartbutler: Pull everything below this line into a file like `resource_aviatrix_transit_gateway_peering_helpers.go`, and have their associated tests in `resource_aviatrix_transit_gateway_peering_helpers_test.go`. That way it is easier to reason about the helper functions independent of the business logic.

docs/resources/aviatrix_transit_gateway_peering.md
```diff
@@ -38,11 +38,11 @@ resource "aviatrix_transit_gateway_peering" "test_transit_gateway_peering" {
 resource "aviatrix_transit_gateway_peering" "test_edge_transit_gateway_peering" {
   transit_gateway_name1   = "test-edge-transit-1"
   transit_gateway_name2   = "test-edge-transit-2"
-  over_private_network    = true
+  enable_peering_over_private_network  = true
```
stewartbutler, line 41: Will these renames cause breakage in existing customers?

stewartbutler (changes requested):

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: over_private_network was added to support EAT peerings by me. This attribute is not available to any users yet as it was not a part of any release. The functionality of this attribute is achieved by reusing `enable_peering_over_private_network` which maps to `over_private_network` param while making the create api call.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: src_wan_interfaces and dst_wan_interfaces are not part of any release yet. These attributes were added to support Edge gateways interfaces. Since we introduced logical interface names for megaport, I am using this name for all edge gateways and translating them before the api calls.
For aep and equinix peerings, I convert the logical_ifnames from wan0 -> eth0, wan1-> eth1 etc. Megaport api's support logical interfaces directly so I don't do the translation for it.

sairane10: These will be added for the first time in 8.0 release.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: This is not a required attribute and should default to false if not set by the users. 
Refer: Transit to Transit peering in https://aviatrix.atlassian.net/wiki/x/iQIadw

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: I found it painful writing a test around functions using resource schema directly. @stewartbutler is there an example I can refer to for this UT?

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: @stewartbutler, the logic in this function is something I have not tried to refactor or change. I have just improved the readability by breaking down the larger create call into smaller chunks. The edge transit peering logic was bloating the create call and would have been hard to maintain as is.

docs/resources/aviatrix_transit_gateway_peering.md
sairane10: over_private_network attribute is not yet release to any customers. It will be released during 8.0.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Its required for EAT to EAT peerings. For non EAT peerings, jumbo frame is not required

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Its required for EAT to EAT peerings. For non EAT peerings, insane mode is not required. Non eat peerings use enable_insane_mode_encryption_over_internet

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Yes that error is handled in this function `setNonEATPeeringOptions`

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: If the cloud type is Equinix or NEO, we need to convert the logical interface names from wan0, wan1 to eth0, eth1. The api for these edge types does not support logical interface names.
If the cloud type is Megaport, we need to set the logical interface names as the api for this edge type supports logical interface names. Megaport api's throw an error if src_wan_interfaces and dst_wan_interfaces are set while making the create api call
The reverseIfNameTranslation function does the this conversion for different gateway cloud types.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: this function edits the path prepend for transit peering. I kept the original code as is and moved it to a separate function
Path prepend is used to route propagate transit peering just like we do for a cross bgp neighbor. It will add one hop by default, we can add more hops to control the path selection. Similar to what we do for bgp neighbors.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: I think thats a question for routing team. I tried to keep the original implementation of prepend_as_path as is.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Yes updated the `setNonEATPeeringOptions` to do readonly validations and values for other options are directly set in this function

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: updated the code

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Added the test for this

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: added UT for this

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: updated. Added the forceNew back to the code.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Updated the code and added UT for it based on your feedback changes.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Removed this helper function and moved the logic back to `setNonEATPeeringOptions` to improve readability

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Removed this helper function and moved the code to `setNonEATPeeringOptions`

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: I have renamed the attribute to avoid confusion in my helper function. The TF schema uses enble_max_performance but the `get_transit_peering` api returns an attribute called no_max_performance and does not an entry for enable_max_performance. 
Please refer to the updated `setNonEATPeeringOptions` 
I have tried to keep the original logic as and didn't want to change the existing attribute used through out the code.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Updated the code. I had added this to keep the cyclomatic complexity of this function below 10. But as you pointed out, it might cause isses so updated the helper function `setNonEATPeeringOptions`

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: while reading the values from the TF schema, it returns it as []interfaces{} rather than []string. The api expects these list values to be []string so I created the helper function for the conversion

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Updated the code to use switch statements.

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Added the helper function for check

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Added a timeout based context similar to spoke-transit attachments

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Updated the code and moved the helper functions to a separate file

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: added a comment for this already. Kept the original code as is. Just moved it to a helper function

aviatrix/resource_aviatrix_transit_gateway_peering.go
sairane10: Is there an example I can refer to for mocking client and making the api call?

stewartbutler (approved):
Looks good, thanks for breaking it up and adding those additional tests.

PR #2165, opened 2025-02-13
AVX-56050: fix HA eip resuse on azure when calling transit resource update

sw-avtx:
The Update function for transit gateway resource was incorrectly deciding on how to set the re-use EIP field for Azure HA GW. It was looking at GwSize on the local var transitHaGw before it was ever set so it was impossible for the EIP field to ever be set in an update call where the HA GW would be created.

To fix this I moved the block down slightly so I could leverage `newHaGwEnabled` and `changeHaGw` which already exist.

Validated on my setup.

### DIFF
--- a/aviatrix/resource_aviatrix_transit_gateway.go
+++ b/aviatrix/resource_aviatrix_transit_gateway.go
@@ -2530,25 +2530,6 @@ func resourceAviatrixTransitGatewayUpdate(d *schema.ResourceData, meta interface
 			transitHaGw.Eip = haEip
 		}
 
-		haAzureEipName, haAzureEipNameOk := d.GetOk("ha_azure_eip_name_resource_group")
-		if goaviatrix.IsCloudType(gateway.CloudType, goaviatrix.AzureArmRelatedCloudTypes) {
-			if haEip != "" && transitHaGw.GwSize != "" {
-				// No change will be detected when ha_eip is set to the empty string because it is computed.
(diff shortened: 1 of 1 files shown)


aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -2599,6 +2580,28 @@ func resourceAviatrixTransitGatewayUpdate(d *schema.ResourceData, meta interface
...
+				if !assertOk {
+					return fmt.Errorf("failed to create HA Transit Gateway: 'ha_azure_eip_name_resource_group' must be a string")
+				}
+				transitHaGw.Eip = fmt.Sprintf("%s:%s", haAzureEipNameString, haEip)
+			}
+		} else if haAzureEipNameOk {
+			return fmt.Errorf("failed to create HA Spoke Gateway: 'azure_eip_name_resource_group' must be empty when cloud_type is not one of Azure (8), AzureGov (32) or AzureChina (2048)")
+		}
+
+		if !d.HasChange("ha_subnet") && d.HasChange("ha_insane_mode_az") {
```
tommy-carver: Any particular reason this block was moved?

tommy-carver (dismissed):

aviatrix/resource_aviatrix_transit_gateway.go
sw-avtx: accident good catch thanks.

tommy-carver (dismissed):

tommy-carver (approved):

PR #2181, opened 2025-03-10
AVX-62274 Updating the sort order for interfaces in EAT

sairane10:
- Fixing the sort order of interfaces using the order provided by the user
- The older code for updating the eip_map was sending the data as form string. For Megaport, we require eip_map to be provided as a json objects instead. In this PR, I have updated the eip_map code to use the API v2 version which accepts json object. This ensures that the eip_map gets updated correctly and the original object structure is passed directly for the update.

### DIFF
--- a/goaviatrix/transit_vpc.go
+++ b/goaviatrix/transit_vpc.go
@@ -293,7 +293,7 @@ func (c *Client) UpdateEdgeGateway(gateway *TransitVpc) error {
 	return c.PostAPI(action, form, BasicCheck)
 }
 
-func (c *Client) UpdateTransitGateway(ctx context.Context, gateway *TransitVpc) error {
+func (c *Client) UpdateEdgeGatewayV2(ctx context.Context, gateway *TransitVpc) error {
 	gateway.CID = c.CID
 	gateway.Action = "update_edge_gateway"
 	return c.PostAPIContext2(ctx, nil, gateway.Action, gateway, BasicCheck)
--- a/goaviatrix/transit_vpc.go
+++ b/goaviatrix/transit_vpc.go
@@ -2,6 +2,7 @@ package goaviatrix
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"log"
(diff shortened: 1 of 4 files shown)


tommy-carver:
Can you elaborate more on what this PR is doing? I understand the update of the sort order for the interface, but the other changes I don't understand the purpose and "Updating the logical eip map update using the api version 2. The new version works with json object." confuses me slightly.

sairane10:
@tommy-carver, the older code for updating the eip_map was sending the data as form string. For Megaport, we require eip_map to be provided as a json objects instead. In this PR, I have updated the eip_map code to use the API v2 version which accepts json object. This ensures that the eip_map gets updated correctly and the original object structure is passed directly for the update.

goaviatrix/transit_vpc.go
```diff
@@ -292,6 +293,12 @@ func (c *Client) UpdateEdgeGateway(gateway *TransitVpc) error {
 	return c.PostAPI(action, form, BasicCheck)
 }
 
+func (c *Client) UpdateTransitGateway(ctx context.Context, gateway *TransitVpc) error {
```
tommy-carver: This function is called update transit gateway, but calls update edge gateway. Since the change you mentioned switched us to using API v2, can we call this relevantly, UpdateEdgeGatewayV2?

goaviatrix/transit_vpc.go
```diff
@@ -13,11 +14,11 @@ import (
 // Gateway simple struct to hold gateway details
 type TransitVpc struct {
 	AccountName                  string `form:"account_name,omitempty" json:"account_name,omitempty"`
-	Action                       string `form:"action,omitempty"`
-	CID                          string `form:"CID,omitempty"`
+	Action                       string `form:"action,omitempty" json:"action,omitempty"`
+	CID                          string `form:"CID,omitempty" json:"CID,omitempty"`
 	CloudType                    int    `form:"cloud_type,omitempty" json:"cloud_type,omitempty"`
 	DnsServer                    string `form:"dns_server,omitempty" json:"dns_server,omitempty"`
-	GwName                       string `form:"gw_name,omitempty" json:"vpc_name,omitempty"`
```
tommy-carver, line 20: Are there ramifications towards removing this?

tommy-carver (approved):

PR #2186, opened 2025-03-23
Add a simple cache for the app-domains API.

jpeach:
Since there is no way to query the state of a single entry from the app-domains API, add a very conservative cache that mitigates refreshing a lot of smart or web groups. We deliberately serialize access the the API endpoint, dropping the cache after any modification or a conservative 5sec timeout.


stewartbutler (dismissed):

jpeach:
Internal testing on this patch was successful.

PR #2188, opened 2025-03-27
AVX-62741: add a simple cache for the app-domains API

jpeach:
Since there is no way to query the state of a single entry from the app-domains API, add a very conservative cache that mitigates refreshing a lot of smart or web groups. We deliberately serialize access the the API endpoint, dropping the cache after any modification or a conservative 5sec timeout.

### DIFF
--- a/goaviatrix/appdomain_cache.go
+++ b/goaviatrix/appdomain_cache.go
@@ -1,0 +1,121 @@
+package goaviatrix
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"maps"
+	"slices"
+	"sync"
+	"time"
+)
+
+type AppdomainGroup struct {
+	UUID     string          `json:"uuid"`
+	Name     string          `json:"name"`
+	Selector json.RawMessage `json:"selector"`
+}
+
+type AppdomainCache struct {
+	lock      sync.Mutex
+	updatedAt time.Time
+
+	cache map[string]AppdomainGroup
+}
+
+func (a *AppdomainCache) expired() bool {
+	const cacheTime = 5 * time.Second
+	return time.Since(a.updatedAt) > cacheTime
+}
+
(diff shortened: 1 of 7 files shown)


goaviatrix/appdomain_cache.go
```diff
@@ -0,0 +1,121 @@
+package goaviatrix
+
+import (
```
dnardo, line 3: How did we test this?  I'm just wondering if did any tests where we add entries continuously.
For instance with accounts we had something like this:

```
resource "aviatrix_account" "aws_account" {
  count               = 500
  account_name = "aws-${count.index +1}"
  cloud_type = 1
....
  ]
}
```

And we saw some strange corner cases with the TF state while doing that.  You may not have these issues, but I just want to make sure we did some sort of test with bulk updates.

I wonder also how much the cache time plays a role here also.

jpeach: The field/support folks tested this with the terraform attached to AVX-60981.

dnardo (approved):

kspoelstra-aviatrix:
LGTM, @jpeach , I think this is an important part of the solution, I'm not quite sure about the edge cases with terraform, do we want to make this configurable in case we run into unforeseen edgecases? 

For large deployments we are going to support using multiple terraform  in 8.1.0.  The cache should not be an issue as users should have their own smartgroups and not collide.
We also do have a change coming where we can query the smartgroups individually, taking away the large request. But still doing multiple requests and we're subject to latency there. 

I don't think we have any OOB smartgroup creation triggered by other terraform which could give us issues here . (@shinrich , do you know of any non smartgroup operations which create smartgroups? (Enabling DCF?))

jpeach:
> do we want to make this configurable in case we run into unforeseen edgecases

We don't have a way to do that today. As you note, the right medium term fix is to improve the API :)

PR #2201, opened 2025-04-11
AVX-62599: Using the new GET method to get app_domain

avtx-akonnur:
Removing the caching method for getting the API as we can now get it directly from the API.


jpeach (commented):
Is this done or are you planning followup PRs to remove the caching code and add webgroup support?

avtx-akonnur:
> Is this done or are you planning followup PRs to remove the caching code and add webgroup support?

I was thinking we could leave the caching logic for list all app domains, it might help it be faster right? What are your thoughts?

goaviatrix/smart_group.go
```diff
@@ -202,21 +202,18 @@ func (c *Client) CreateSmartGroup(ctx context.Context, smartGroup *SmartGroup) (
 }
 
 func (c *Client) GetSmartGroup(ctx context.Context, uuid string) (*SmartGroup, error) {
-	g, err := c.appdomainCache.Get(ctx, c, uuid)
-	if err != nil {
-		return nil, err
-	}
+	endpoint := fmt.Sprintf("app-domains/%s", uuid)
```
kspoelstra-aviatrix, line 208: @jpeach , @avtx-akonnur , do we want to keep the appdomain cache in the other functions?

avtx-akonnur: Removed the caching logic completely

goaviatrix/smart_group.go
```diff
@@ -198,57 +197,54 @@ func makeSmartGroupForm(smartGroup *SmartGroup) map[string]interface{} {
...
-		result := SmartGroupResult{
-			UUID: g.UUID,
-			Name: g.Name,
-		}
-
-		if err := json.Unmarshal(g.Selector, &result.Selector); err != nil {
-			return nil, err
+	var smartGroups []*SmartGroup
+	for _, smartGroupResult := range data.SmartGroups {
+		if smartGroupResult.UUID != "" {
```
jpeach, line 244: Is UUID == "" a legal value? What does it mean?

jpeach (approved):
Looks reasonable, but I don't see the API in cloudx-local/daemon/v2_5/schemas/microseg_appdomain.yaml. Suggest we update that and add apigroup tests.

PR #2203, opened 2025-04-14
AVX-62963: Removed input validation for type key, this will be handled by the API

avtx-akonnur:
Removing the input validation for type key in external group creation, the input validation gets complicated as it needs to be avoided if there is an external keyword mentioned, I think it's best if we let the API handle this validation check instead of having it in the TF.


kasunkarunaratne (approved):
lgtm, I think API level validation here is sufficient

tommy-carver (commented):
Can we update relevant docs as well?

tommy-carver (dismissed):

tommy-carver (approved):

PR #2209, opened 2025-04-24
AVX-60903 TF support for DCF Default Action Policy

xc-6:
This PR allows users to modify the DCF `default action policy` action and logging settings in TF.

When the DCF feature flag is enabled, the default action policy is set to {"action": "PERMIT", "logging": false}, with both the source (src) and destination (dst) set to Anywhere. This ensures that the traffic in the existing system remains uninterrupted.

Users can specify their desired action and logging settings during TF creation or update. When a user sets the action to "DENY", all traffic will be denied since both the source (src) and destination (dst) are set to Anywhere on the `default_action_policy`. To ensure necessary traffic is allowed, users should create corresponding PERMIT rules based on their architecture design.

When TF is destroyed, the default action policy reverts to {"action": "PERMIT", "logging": false} instead of being deleted. 
The rule is only deleted when the DCF feature flag is turned off.

### Test 
```

  # aviatrix_distributed_firewalling_zero_trust_rule.test-xc will be updated in-place
  ~ resource "aviatrix_distributed_firewalling_default_action_policy" "test-xc" {
      + action  = "DENY"
        id      = "34-42-176-139"
        # (1 unchanged attribute hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

aviatrix_distributed_firewalling_default_action_policy.test-xc: Modifying... [id=34-42-176-139]
aviatrix_distributed_firewalling_default_action_policy.test-xc: Modifications complete after 0s [id=34-42-176-139]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```


goaviatrix/distributed_firewalling_zero_trust_rule.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (c *Client) UpdateDistributedFirewallingZeroTrust(ctx context.Context, request *DistributedFirewallingZeroTrustRule) error {
+	endpoint := "microseg/edit-zero-trust"
+	return c.PutAPIContext25(ctx, endpoint, request)
+}
+
+func (c *Client) GetDistributedFirewallingZeroTrustRule(ctx context.Context) (map[string]string, error) {
+	postRuleListUUID := "defa11a1-3000-5000-0000-000000000000"
+	zeroTrustRuleUUID := "defa11a1-0000-0000-0000-000000000000"
```
jpeach: Are these well-known constants? Could you link to a reference to help future maintainers?

aviatrix/provider.go
```diff
@@ -111,6 +111,7 @@ func Provider() *schema.Provider {
 			"aviatrix_distributed_firewalling_intra_vpc":                      resourceAviatrixDistributedFirewallingIntraVpc(),
 			"aviatrix_distributed_firewalling_origin_cert_enforcement_config": resourceAviatrixDistributedFirewallingOriginCertEnforcementConfig(),
 			"aviatrix_distributed_firewalling_policy_list":                    resourceAviatrixDistributedFirewallingPolicyList(),
+			"aviatrix_distributed_firewalling_zero_trust_rule":                resourceAviatrixDistributedFirewallingZeroTrustRule(),
```
jpeach: We should add documentation for the new resource. Not quite sure how docs work, but poke around in docs/resources and see what you can do.

aviatrix/resource_aviatrix_distributed_Firewalling_zero_trust_rule.go
```diff
@@ -0,0 +1,151 @@
...
+	return resourceAviatrixDistributedFirewallingZeroTrustRuleRead(ctx, d, meta)
+}
+
+func resourceAviatrixDistributedFirewallingZeroTrustRuleCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, ok := meta.(*goaviatrix.Client)
+	if !ok {
+		return diag.Errorf("failed to assert meta as *goaviatrix.Client")
+	}
+
+	action, ok := d.Get("action").(string)
```
jpeach: Lots of the code in the provider just does a type assertion without the error handling, which seems OK to me since if it failed it would mean the code is broken pretty badly.

Probably ok if you keep this too, if other code is doing the same pattern.

aviatrix/resource_aviatrix_distributed_Firewalling_zero_trust_rule.go
```diff
@@ -0,0 +1,151 @@
...
+	return resourceAviatrixDistributedFirewallingZeroTrustRuleRead(ctx, d, meta)
+}
+
+func resourceAviatrixDistributedFirewallingZeroTrustRuleRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, ok := meta.(*goaviatrix.Client)
+	if !ok {
+		return diag.Errorf("failed to assert meta as *goaviatrix.Client")
+	}
+
+	if d.Id() != strings.Replace(client.ControllerIP, ".", "-", -1) {
```
jpeach: https://pkg.go.dev/strings#ReplaceAll

aviatrix/resource_aviatrix_distributed_Firewalling_zero_trust_rule.go
```diff
@@ -0,0 +1,151 @@
...
+		return diag.Errorf("ID: %s does not match controller IP. Please provide correct ID for importing", d.Id())
```
jpeach:
```suggestion
		return diag.Errorf("ID: %s does not match controller IP %q: please provide correct ID for importing", d.Id(), client.ControllerIP)
```

aviatrix/resource_aviatrix_distributed_Firewalling_zero_trust_rule.go
```diff
@@ -0,0 +1,151 @@
...
+			d.SetId("")
+			return nil
+		}
+		return diag.Errorf("failed to read the zero trust rule: %s", err)
+	}
+
+	if err := d.Set("action", postRuleList["action"]); err != nil {
+		return diag.Errorf("failed to set 'action': %v", err)
+	}
+	if err := d.Set("logging", postRuleList["logging"]); err != nil {
```
jpeach: Does `d.Set` implicitly convert this string back into a bool?

goaviatrix/distributed_firewalling_zero_trust_rule.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	// Iterate through the policies to find the rule with the matching UUID
+	for _, policy := range response.DcfPolicies.Policies {
+		if policy.UUID == zeroTrustRuleUUID {
+			return map[string]string{
+				"Action":  policy.Action,
+				"Logging": strconv.FormatBool(policy.Logging),
+				"UUID":    policy.UUID,
+			}, nil
```
jpeach: Why don't you return a DistributedFirewallingZeroTrustRule here?

goaviatrix/distributed_firewalling_zero_trust_rule.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+type DistributedFirewallingZeroTrustRule struct {
+	Action  string `json:"action"`
+	Logging bool   `json:"logging"`
+	UUID    string `json:"uuid,omitempty"`
+}
+
+func (c *Client) UpdateDistributedFirewallingZeroTrust(ctx context.Context, request *DistributedFirewallingZeroTrustRule) error {
+	endpoint := "microseg/edit-zero-trust"
```
jpeach: If you can PUT on this endpoint, why can't you GET instead of the searching business below?

aviatrix/resource_aviatrix_distributed_Firewalling_zero_trust_rule.go
xc-6: Thanks. Just updated the code to convert it back to bool

aviatrix/provider.go
xc-6: Thanks! Just update it on docs/resources/aviatrix_distributed_firewalling_default_action_policy.md

goaviatrix/distributed_firewalling_default_action_policy.go
```diff
@@ -0,0 +1,35 @@
...
+	var response DistributedFirewallingDefaultActionPolicy
+
+	err := c.GetAPIContext25(ctx, &response, endpoint, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get distributed firewalling default action policy: %w", err)
+	}
+
+	result := map[string]string{
+		"action":  response.Action,
+		"logging": strconv.FormatBool(response.Logging),
```
jpeach: We need to go back to the cloudn API and define the response type. Then we can revisit this PR.

xc-6: Thanks. Cloudn API update PR: https://github.com/AviatrixDev/cloudn/pull/45176/files

goaviatrix/distributed_firewalling_default_action_policy.go
```diff
@@ -0,0 +1,34 @@
...
+	Logging bool   `json:"logging"`
+}
+
+func (c *Client) UpdateDistributedFirewallingDefaultActionPolicy(ctx context.Context, request *DistributedFirewallingDefaultActionPolicy) error {
+	endpoint := "microseg/default-action-policy"
+	return c.PutAPIContext25(ctx, endpoint, request)
+}
+
+func (c *Client) GetDistributedFirewallingDefaultActionPolicy(ctx context.Context) (*DistributedFirewallingDefaultActionPolicy, error) {
+	endpoint := "microseg/default-action-policy"
```
tommy-carver: Can we make this a const and use the value instead of setting it twice?

docs/resources/aviatrix_distributed_firewalling_default_action_policy.md
```diff
@@ -0,0 +1,37 @@
...
+subcategory: "Secured Networking"
+layout: "aviatrix"
+page_title: "Aviatrix: aviatrix_distributed_firewalling_default_action_policy"
+description: |-
+  Creates and manages an Aviatrix Distributed-firewalling default action policy
+---
+
+# aviatrix_distributed_firewalling_default_action_policy
+
+The **aviatrix_distributed_firewalling_default_action_policy** resource handles the creation and management of Distributed-firewalling default action policy. Available as of Provider 3.2.2+.
```
tommy-carver, line 11: Can we elaborate more on exactly what this is? Something like "This sets the default action policy of any newly created distributed firewalls to be this action. If not set, the default is {default}"

xc-6: Good point. Just updated. Thanks

tommy-carver (approved):

PR #2211, opened 2025-05-01
AVX-54439: Add terraform changes for BGP Communities

triwas:
This commit adds terraform changes for BGP Communities. API details can be found at: https://aviatrix.atlassian.net/wiki/spaces/AVXENG/pages/2745237507/BGP+Communities+APIs.


aviatrix/resource_aviatrix_controller_bgp_communities_auto_cloud_config.go
```diff
@@ -0,0 +1,145 @@
...
+	if !ok {
+		return diag.Errorf("failed to assert meta as *goaviatrix.Client")
+	}
+
+	if d.HasChange("community_prefix") {
+		autoCloud, ok := d.Get("auto_cloud_enabled").(bool)
+		if !ok {
+			return diag.Errorf("failed to assert auto_cloud_enabled as bool")
+		}
+		if autoCloud {
```
dnardo, line 118: If a user flips auto_cloud_enabled = false while leaving the prefix unchanged, `DisableControllerBgpCommunitiesAutoCloud` is never called, so the resource drifts.  Is that ok?

aviatrix/resource_aviatrix_controller_bgp_communities_auto_cloud_config.go
```diff
@@ -0,0 +1,145 @@
...
+			return nil
+		}
+		return diag.Errorf("could not get controller BGP communities auto cloud config: %v", err)
+	}
+
+	err = d.Set("community_prefix", commPrefix)
+	if err != nil {
+		return diag.Errorf("failed to set community prefix: %v", err)
+	}
+	if commPrefix > 0 {
```
dnardo, line 92: We don't  set  `auto_cloud_enabled = false` explicitly when the controller returns a prefix `0`. In other words, You never call `d.Set("auto_cloud_enabled", false)`. Terraform will think the attribute is unknown, you should probably set this attribute explicitly.

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -655,6 +667,32 @@ func resourceAviatrixSpokeGatewayCreate(d *schema.ResourceData, meta interface{}
...
 
+	setComm := false
+	commSendCurr, commAcceptCurr, err := client.GetGatewayBgpCommunities(gateway.GwName)
+	acceptComm, ok := d.Get("bgp_accept_communities").(bool)
+	if !ok {
+		return fmt.Errorf("failed to assert bgp_accept_communities as a boolean")
+	}
+	if acceptComm != commAcceptCurr || err != nil {
+		setComm = true
+		err := client.SetGatewayBgpCommunitiesSend(gateway.GwName, acceptComm)
```
dnardo: Should this be calling an API called `SetGatewayBgpCommunitiesAccept`?  Or do we only have the Send API and both branches call this same Send?   If this is how its supposed to be, comments would be helpful although its unclear how we differentiate between send and accept from this API alone with no additional parameters.

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -655,6 +667,32 @@ func resourceAviatrixSpokeGatewayCreate(d *schema.ResourceData, meta interface{}
...
+		if err != nil {
+			return fmt.Errorf("failed to set send BGP communities for gateway %s: %w", gateway.GwName, err)
```
dnardo: Should this be failed to set accept?

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -1787,6 +1838,32 @@ func resourceAviatrixSpokeGatewayUpdate(d *schema.ResourceData, meta interface{}
...
 	}
 
+	commSendCurr, commAcceptCurr, err := client.GetGatewayBgpCommunities(gateway.GwName)
+	if d.HasChange("bgp_accept_communities") {
+		acceptComm, ok := d.Get("bgp_accept_communities").(bool)
+		if !ok {
+			return fmt.Errorf("failed to assert bgp_accept_communities as a boolean")
+		}
+		if acceptComm != commAcceptCurr || err != nil {
+			err := client.SetGatewayBgpCommunitiesSend(gateway.GwName, acceptComm)
```
dnardo: same question as above for both this call and the error message below and the places where this is done below.

aviatrix/resource_aviatrix_transit_external_device_conn.go
```diff
@@ -506,6 +528,51 @@ func resourceAviatrixTransitExternalDeviceConnCreate(d *schema.ResourceData, met
 		DisableActivemesh:      d.Get("disable_activemesh").(bool),
 	}
 
+	sendComm, ok := d.Get("connection_bgp_send_communities").(string)
```
dnardo, line 531: Comments around intentions would be helpful:

```
Detect whether the user wants to change the set of BGP communities sent on a given connection; if so....
```

aviatrix/resource_aviatrix_controller_bgp_communities_auto_cloud_config.go
triwas: Updated this, thanks

aviatrix/resource_aviatrix_controller_bgp_communities_auto_cloud_config.go
triwas: Updated if statement on line 108

aviatrix/resource_aviatrix_spoke_gateway.go
triwas: No, this was a typo, should be `SetGatewayBgpCommunitiesAccept`

aviatrix/resource_aviatrix_transit_external_device_conn.go
triwas: Added comment

dnardo (approved):

PR #2197, opened 2025-04-03
AVX-61735 Adding custom interface mapping to HA Self Managed Spoke

sairane10:
Adding the self-managed ha spoke gateway with custom interface mapping logic
The custom interface mapping supports PCI ID's and mac address. 
This PR is an extension to the spoke gateway PR - https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2210

```
resource "aviatrix_edge_gateway_selfmanaged_ha" "edge_self_managed_ha" {
    primary_gw_name = "eas_self_managed_1"
    ztp_file_type = "cloud-init"
    ztp_file_download_path = "ztp"
    site_id = "eas-site-1"
    management_egress_ip_prefix_list = [ 
        "162.43.140.85/32"
    ]
    interfaces {
        name = "eth0"
        type = "WAN"
        ip_address = "192.168.19.12/24"
        gateway_ip = "192.168.19.1"
    }

    interfaces {
        name = "eth1"
        type = "LAN"
        ip_address = "192.168.20.12/24"
        gateway_ip = "192.168.20.1"
    }

    interfaces {
        name = "eth2"
        type = "MANAGEMENT"
        enable_dhcp = true
    }

    custom_interface_mapping {
      logical_ifname = "wan0"
      identifier_type = "system-assigned"
      identifier_value = "auto"
    }

    custom_interface_mapping {
      logical_ifname = "lan0"
      identifier_type = "mac"
      identifier_value = "00:0c:29:63:82:b2"
    }

    custom_interface_mapping {
      logical_ifname = "mgmt0"
      identifier_type = "pci"
      identifier_value = "pci@0000:04:00.0"
    }
}
```

PR #2237, opened 2025-05-20
AVX-63900 Reverting the changes for custom interface mapping in self managed HA gateway

sairane10:
- Reverting the custom interface mapping changes from this PR - https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2197
- Keeping the selfmanaged field in const.go file to support the creation of selfmanaged EAT
- Adding the golang lint changes for HA self-managed gateways

This custom mapping will be useful for self-managed platforms like ESXi and KVM. When a VM is deployed on these platforms, the interfaces might not follow PCI ordering, which has caused issues in the past. Determining the actual interface pattern and configuring the correct interface_mapping for such platforms has been a challenge. Using MAC addresses or PCI bus IDs as the basis for mapping would provide a more deterministic and flexible way to discover and configure interfaces, potentially even dynamically at runtime, instead of doing it all at boot time.

However, we have decided not to include this in version 8.1 for the following reasons:

In 8.1, cloud_init not only includes edge metadata but also contains the ZTP (Zero Touch Provisioning) code that brings up the edge and transitions it to phase-2. Adding support for this custom mapping—along with the code to handle it—would increase the size of the cloud_init file, potentially exceeding 16 KB. This would require an updated implementation by the Edge team to support containerized ESXi/KVM gateways. Specifically, the interface mapping would need to be split—initially bringing up only the management interface and deferring the rest of the interface configuration to phase-2. This would involve careful planning and rework, which is not feasible for 8.1 as we are already past CF1.

Furthermore, starting in 8.1, even self-managed edges support image upgrades via the config. This means a new cloud_init file can be downloaded and used without requiring deletion and recreation of the edge configuration in the controller. However, this custom mapping based on MAC or PCI IDs adds complexity during image upgrades—especially for self-managed edges. When moving an edge to a different host, the MAC and PCI IDs of the interfaces may change. Therefore, a new workflow is needed to collect updated MAC or PCI bus IDs during the image upgrade process. Supporting this would require additional configuration logic and new APIs, which again is out of scope for 8.1 since we're already past CF1.

```
resource "aviatrix_edge_gateway_selfmanaged_ha" "edge_self_managed_ha" {
    primary_gw_name = "eas_self_managed_1"
    ztp_file_type = "cloud-init"
    ztp_file_download_path = "ztp"
    site_id = "eas-site-1"
    management_egress_ip_prefix_list = [ 
        "162.43.140.85/32"
    ]
    interfaces {
        name = "eth0"
        type = "WAN"
        ip_address = "192.168.19.12/24"
        gateway_ip = "192.168.19.1"
    }

    interfaces {
        name = "eth1"
        type = "LAN"
        ip_address = "192.168.20.12/24"
        gateway_ip = "192.168.20.1"
    }

    interfaces {
        name = "eth2"
        type = "MANAGEMENT"
        enable_dhcp = true
    }
}
```

### DIFF
--- a/aviatrix/resource_aviatrix_edge_gateway_selfmanaged_ha.go
+++ b/aviatrix/resource_aviatrix_edge_gateway_selfmanaged_ha.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"log"
 	"os"
-	"regexp"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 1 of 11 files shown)


sairane10:
@avx-rchung I have updated the Jira description as well as the notes here to support the removal of custom interface mapping

avx-rchung (approved):

PR #2247, opened 2025-06-04
AVX-59186: terraform changes for route approval

allwync:
Prior to Release 8.1, users were only allowed to set up "exact prefix match" values for the list of approved_learned_cidrs.  In 8.1, we have expanded that to take optional "ge" and "le" values in order to let users specify a range of prefixes.  We also refactored the APIs to remove unnecessary duplication, for example, we used to have separate APIs for spoke/ transit that were doing the same thing.  In this PR we modify the Aviatrix Terraform Provider to use these these new APIs.


aviatrix/utils.go
```diff
@@ -426,3 +427,85 @@ func sortSpokeInterfacesByCustomOrder(interfaces []goaviatrix.MegaportInterface,
 	})
 	return interfaces
 }
+
+// ValidateCIDRRule validates that the string is a valid CIDR rule in the format:
+// a.b.c.d/x [ge y] [le z] where x <= y <= z <= 32
+func ValidateCIDRRule(i interface{}, k string) (warnings []string, errors []error) {
+	cidrRuleStr, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected type of %s to be string", k))
```
dnardo: No need to append errors here:
Just:

`return nil, []error{fmt.Errorf("%q must be a string, got %T", k, v)}`

aviatrix/utils.go
```diff
@@ -426,3 +427,85 @@ func sortSpokeInterfacesByCustomOrder(interfaces []goaviatrix.MegaportInterface,
...
+		return warnings, errors
+	}
+
+	// Helper function to add error
+	invalidRule := func(message string) {
```
dnardo: This doesn't look necessary since you are always returning on error so I don't think you are actually accumulating errors.

aviatrix/utils.go
```diff
@@ -426,3 +427,85 @@ func sortSpokeInterfacesByCustomOrder(interfaces []goaviatrix.MegaportInterface,
...
+		return
+	}
+
+	// Validate the CIDR part
+	cidrPart := parts[0]
+	ip, ipNet, err := net.ParseCIDR(cidrPart)
+	if err != nil {
+		invalidRule(fmt.Sprintf("invalid CIDR %s", cidrPart))
+		return
+	}
```
dnardo: You can combine this:

```
ip, netCIDR, err := net.ParseCIDR(fields[0])
	if err != nil || ip.To4() == nil {
		return nil, []error{fmt.Errorf("invalid IPv4 CIDR %q", fields[0])}
	}
```

aviatrix/utils.go
```diff
@@ -426,3 +427,85 @@ func sortSpokeInterfacesByCustomOrder(interfaces []goaviatrix.MegaportInterface,
...
+	// Ensure the CIDR is normalized (using existing functions from net package)
+	normalizedIP := ip.Mask(ipNet.Mask)
+	if !ip.Equal(normalizedIP) {
+		invalidRule(fmt.Sprintf("CIDR %s is not normalized, should be %s%s",
+			cidrPart, normalizedIP.String(), cidrPart[strings.LastIndex(cidrPart, "/"):]))
+		return
+	}
+
+	// Extract prefix length
+	prefixLen, _ := ipNet.Mask.Size()
```
dnardo: I believe some of this can be simplified 

```
if fields[0] != netCIDR.String() {
		return nil, []error{fmt.Errorf("CIDR %q is not canonical; use %q",
			fields[0], netCIDR.String())}
	}
	prefixLen, _ := netCIDR.Mask.Size()

	if len(fields) == 1 {
		return nil, nil
	}
```

and then this last part can be

```
	var ge, le *int
	for i := 1; i < len(fields); i += 2 {
		n, err := strconv.Atoi(fields[i+1])
		if err != nil {
			return nil, []error{fmt.Errorf("invalid length %q in %q", fields[i+1], s)}
		}
		switch fields[i] {
		case "ge":
			ge = &n
		case "le":
			le = &n
		default:
			return nil, []error{fmt.Errorf("unknown qualifier %q in %q", fields[i], s)}
		}
		if n < prefixLen || n > 32 {
			return nil, []error{fmt.Errorf("length %d out of range for %q", n, s)}
		}
	}
	if ge != nil && le != nil && *ge > *le {
		return nil, []error{fmt.Errorf("ge length %d > le length %d in %q", *ge, *le, s)}
	}
```

aviatrix/utils.go
```diff
@@ -426,3 +427,85 @@ func sortSpokeInterfacesByCustomOrder(interfaces []goaviatrix.MegaportInterface,
 	})
 	return interfaces
 }
+
+// ValidateCIDRRule validates that the string is a valid CIDR rule in the format:
+// a.b.c.d/x [ge y] [le z] where x <= y <= z <= 32
+func ValidateCIDRRule(i interface{}, k string) (warnings []string, errors []error) {
```
dnardo: I would avoid this named parameter stuff, to me its unintuitive

allwync: Done

aviatrix/utils.go
allwync: Done.

aviatrix/utils.go
allwync: Removed

aviatrix/utils.go
allwync: Done

aviatrix/utils.go
allwync: Done

dnardo (approved):

PR #2259, opened 2025-06-12
AVX-63679 Terraform changes for VPC creation in v6

avx-shrutianand:


goaviatrix/vpc.go
```diff
@@ -25,6 +25,10 @@ type Vpc struct {
 	PublicRoutesOnly       bool
 	ResourceGroup          string `json:"resource_group,omitempty"`
 	PrivateModeSubnets     bool
+	EnableIpv6             bool     `form:"enable_ipv6,omitempty" json:"enable_ipv6,omitempty"`
+	VpcIpv6Cidr            string   `form:"vpc_ipv6_cidr,omitempty" json:"vpc_ipv6_cidr,omitempty"`
+	IPV6AccessType         string   `form:"ipv6_access_type,omitempty" json:"ipv6_access_type,omitempty"`
```
avx-siyer, line 30: Is there a manual page or description to be updated where the options are available for the user to see - for example - access-type - Internal/External, and what's the difference between v6-cidr and v6-cidr-list?

avx-siyer: Similarly, how can we indicate which clouds are supported?

PR #2276, opened 2025-07-11
AVX-65811: Add per connection BGP community resources for spoke gw.

triwas:
Spoke gateway communities resources were missing. These changes are similar to transit gateway changes in `aviatrix/resource_aviatrix_transit_external_device_conn.go`

### DIFF
--- a/aviatrix/resource_aviatrix_spoke_external_device_conn.go
+++ b/aviatrix/resource_aviatrix_spoke_external_device_conn.go
@@ -1452,17 +1452,7 @@ func resourceAviatrixSpokeExternalDeviceConnUpdate(d *schema.ResourceData, meta 
 		}
 	}
 
-	if d.HasChange("connection_bgp_send_communities") || d.HasChange("connection_bgp_send_communities_additive") || d.HasChange("connection_bgp_send_communities_block") {
-		connName, ok := d.Get("connection_name").(string)
-		if !ok {
-			return fmt.Errorf("failed to assert connection_name as string")
-		}
-
-		gwName, ok := d.Get("gw_name").(string)
-		if !ok {
-			return fmt.Errorf("failed to assert gw_name as string")
-		}
-
+	if d.HasChanges("connection_bgp_send_communities", "connection_bgp_send_communities_additive", "connection_bgp_send_communities_block") {
 		// Detect whether the user wants to change the set of BGP communities sent on a given connection
 		// if so, update the connection with the new set of communities, either additively or as a replacement
 		// or block the communities entirely, depending on the user's choice
--- a/aviatrix/resource_aviatrix_spoke_external_device_conn.go
+++ b/aviatrix/resource_aviatrix_spoke_external_device_conn.go
@@ -426,12 +426,34 @@ func resourceAviatrixSpokeExternalDeviceConn() *schema.Resource {
 				Default:     true,
 				Description: "Enable multihop on BGP connection.",
 			},
+			"connection_bgp_send_communities": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: "Connection based additional BGP communities to be sent",
+			},
+			"connection_bgp_send_communities_additive": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				ForceNew:    true,
+				Description: "Do additive operation instead of replacement operation",
+			},
+			"connection_bgp_send_communities_block": {
+				Type:        schema.TypeBool,
+				Optional:    true,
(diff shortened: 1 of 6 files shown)


sairane10:
@triwas, please add the attributes to the .md files for this resource.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -464,6 +486,41 @@ func resourceAviatrixSpokeExternalDeviceConnCreate(d *schema.ResourceData, meta
 		EnableBgpMultihop:      d.Get("enable_bgp_multihop").(bool),
 	}
 
+	sendComm, ok := d.Get("connection_bgp_send_communities").(string)
+	if !ok {
```
sairane10, line 491: I think this will panic if the attribute is not present in the resource. Since the value is optional we need to handle this case. Can you verify this change by creating a resource without any bgp communities attribute set?

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -464,6 +486,41 @@ func resourceAviatrixSpokeExternalDeviceConnCreate(d *schema.ResourceData, meta
...
+		return fmt.Errorf("failed to assert connection_bgp_send_communities as string")
+	}
+	blockComm, ok := d.Get("connection_bgp_send_communities_block").(bool)
```
sairane10, line 494: same as above. This will panic as well if the attribute is not present in the reosurce.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -1375,6 +1452,47 @@ func resourceAviatrixSpokeExternalDeviceConnUpdate(d *schema.ResourceData, meta
 		}
 	}
 
+	if d.HasChange("connection_bgp_send_communities") || d.HasChange("connection_bgp_send_communities_additive") || d.HasChange("connection_bgp_send_communities_block") {
```
sairane10: you can replace this with a check like this
if d.HasChanges("connection_bgp_send_communities", "connection_bgp_send_communities_additive", "connection_bgp_send_communities_block") {

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -1375,6 +1452,47 @@ func resourceAviatrixSpokeExternalDeviceConnUpdate(d *schema.ResourceData, meta
...
+		connName, ok := d.Get("connection_name").(string)
+		if !ok {
+			return fmt.Errorf("failed to assert connection_name as string")
+		}
+
+		gwName, ok := d.Get("gw_name").(string)
+		if !ok {
+			return fmt.Errorf("failed to assert gw_name as string")
+		}
```
sairane10: The gw_name and conn_name values are already extracted above at the start of the update function. You reuse them instead of re extracting from the schema.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
```diff
@@ -1375,6 +1452,47 @@ func resourceAviatrixSpokeExternalDeviceConnUpdate(d *schema.ResourceData, meta
...
+
+		sendAdditive, ok := d.Get("connection_bgp_send_communities_additive").(bool)
+		if !ok {
+			return fmt.Errorf("failed to assert connection_bgp_send_communities_additive as bool")
+		}
+
+		sendBlock, ok := d.Get("connection_bgp_send_communities_block").(bool)
+		if !ok {
+			return fmt.Errorf("failed to assert connection_bgp_send_communities_block as bool")
+		}
```
sairane10, line 1473: You will have to handle the panic here if any of the attributes are not provided by the user.

aviatrix/resource_aviatrix_spoke_external_device_conn.go
triwas: I did check without the communities attributes set and it does not panic. Same with the block attribute. I also checked the same for transit code which is exactly the same. Wouldn't the `if !ok` handle that in any case?

aviatrix/resource_aviatrix_spoke_external_device_conn.go
triwas: Same comment as above

aviatrix/resource_aviatrix_spoke_external_device_conn.go
triwas: Verified that this will also not panic, leaving as is.

sairane10 (approved):

PR #2277, opened 2025-07-15
AVX-52106: Update function missing fields for edge spoke ext connection

triwas:
Testing: verified with edge spoke gateway that BGP session gets established after providing the MD5 key either in the beginning or while updating the terraform.


aviatrix/resource_aviatrix_edge_spoke_external_device_conn.go
```diff
@@ -620,6 +620,18 @@ func resourceAviatrixEdgeSpokeExternalDeviceConnUpdate(ctx context.Context, d *s
 	if externalDeviceConn.EnableEdgeUnderlay && d.HasChanges("bgp_md5_key", "backup_bgp_md5_key") {
 		edgeExternalDeviceConn := goaviatrix.EdgeExternalDeviceConn(*externalDeviceConn)
 
+		bgpMD5Key, ok := d.Get("bgp_md5_key").(string)
+		if !ok {
+			return diag.Errorf("failed to assert bgp_md5_key as string")
```
dnardo: How about something like, since users will get see this error:

```
return diag.Errorf("invalid value for 'bgp_md5_key': expected a string, but got a %T (value: %v)", val, val)
```

Here and below

triwas: Sure!

dnardo (approved):

PR #2281, opened 2025-07-25
AVX-60902: Adding TF for DCF deployment policy API

avtx-akonnur:
This PR adds TF support for deployment policy API.
The TF create: does a POST request to create the deployment policy based on users set of Providers.
The TF destroy: does a POST request to reset the providers list back to default list which is `{
		"GCP",
		"AWS",
		"AWS-GOV",
		"AZURE-GOV",
		"AZURE",
		"AVX-TEST",
	}`
this uses the set_defaults parameter in the API to reset providers back to default.

Added a unit test as well.


aviatrix/resource_aviatrix_distributed_firewalling_deployment_policy.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func resourceAviatrixDistributedFirewallingDeploymentPolicyDelete(ctx context.Context, _ *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client, ok := meta.(*goaviatrix.Client)
+	if !ok {
+		return diag.Errorf("failed to assert meta as *goaviatrix.Client")
+	}
+	dummyProviders := []string{
+		"GCP",
+		"AWS",
```
dnardo, line 143: In the documentation it says the defaults are 
 `["GCP", "AWS", "AWS-GOV", "AZURE-GOV", "AZURE", "AVX-TEST"]`

But when we reset we don't set back to the defaults we set the dummy values. Its not clear to me why we do that.  What is the difference between dummy and default?

aviatrix/resource_aviatrix_distributed_firewalling_deployment_policy.go
```diff
@@ -0,0 +1,132 @@
...
+				Type:     schema.TypeBool,
+				ForceNew: true,
+				Optional: true,
+				Default:  false,
+				Description: "Set to true to reset the deployment policy to default values. " +
+					"Set to false to create a new deployment policy with the specified providers.",
+			},
+		},
+	}
+}
```
dnardo: According to the docs:

```
set_defaults: true = "deployment policy will be enabled in default CSPs"
set_defaults: false = "enable DCF only in the CSPs mentioned in the providers list"
```

If set_defaults=true means "use system defaults and ignore the providers list", then why force users to provide a providers list that gets ignored?  I'm probably missing something here.

In other words: When a user sets `set_defaults=true`, do they still need to provide a meaningful providers list, or is it truly ignored by the API?

docs/resources/aviatrix_distributed_firewalling_deployment_policy.md
```diff
@@ -0,0 +1,46 @@
+---
+subcategory: "Secured Networking"
+layout: "aviatrix"
+page_title: "Aviatrix: aviatrix_distributed_firewalling_deployment_policy"
+description: |-
+  Creates and manages an Aviatrix Distributed-firewalling deployment policy
+---
+
+# aviatrix_distributed_firewalling_default_action_rule
```
dnardo: copy pasta?

You mean: aviatrix_distributed_firewalling_deployment_policy

aviatrix/resource_aviatrix_distributed_firewalling_deployment_policy.go
avtx-akonnur: We don't set it to the dummy values, we set it to the default. When set_defaults is set to True the dummy ones are ignored by the API.
providers can't be empty, so we just set it to some dummy values as they will be ignored.

aviatrix/resource_aviatrix_distributed_firewalling_deployment_policy.go
avtx-akonnur: It is truly ignored by the API, so if set_defaults is true, it doesn't matter what the user adds in providers argument.

docs/resources/aviatrix_distributed_firewalling_deployment_policy.md
avtx-akonnur: oops will fix it, my bad

aviatrix/resource_aviatrix_distributed_firewalling_deployment_policy.go
dnardo: ok, that's not intuitive, can you add a comment - something like: // These dummy values are required but will be ignored by the API when SetDefaults=true

avtx-akonnur: Added the comment. that makes sense.

aviatrix/resource_aviatrix_distributed_firewalling_deployment_policy.go
dnardo: To me this seems awkward to me:

Since users could write something like this: 
```
resource "aviatrix_distributed_firewalling_deployment_policy" "test" {
  providers = ["ANYTHING", "GOES", "HERE"]  # Completely ignored
  set_defaults = true
}
```

You could make this optional:

```
"providers": {
    Type:     schema.TypeSet,
    Optional: true,  // Make it optional
    Elem: &schema.Schema{Type: schema.TypeString},
},
```

Then add some custom validation?

```
...
    setDefaults := d.Get("set_defaults").(bool)
    providers := d.Get("providers").(*schema.Set)
    
    // When set_defaults=false, providers is required
    if !setDefaults && providers.Len() == 0 {
        return errors.New("providers is required when set_defaults is false")
    }
    
    // Optional: warn when set_defaults=true but providers provided
    if setDefaults && providers.Len() > 0 {
        // Could log a warning that providers will be ignored
    }
...
```

If this is what we want that's fine, but I just want to point out that it may be confusing for users

avtx-akonnur: Thanks a lot, this makes more sense and is intuitive for the user. I have updated it now.

dnardo (approved):

PR #2282, opened 2025-08-03
AVX-66122 terraform support for VPC creation

avx-shrutianand:
This PR adds IPv6 support for VPC creation in the Aviatrix Terraform provider.

### goaviatrix/vpc.go:
- Added IPv6 fields to Vpc struct:
  - EnableIpv6 (bool) - Flag to enable IPv6
  - VpcIpv6Cidr (string) - Single IPv6 CIDR for Azure
  - Ipv6AccessType (string) - Access type for GCP (INTERNAL/EXTERNAL)

- Updated CreateVpc function to handle IPv6 parameters
- Added conditional logic for different cloud providers (Azure vs GCP)

### aviatrix/resource_aviatrix_vpc.go:
- Added IPv6 schema fields:
  - enable_ipv6 (bool)
  - vpc_ipv6_cidr (string)
  - ipv6_access_type (string)
- Added validation for IPv6 CIDR blocks
-  Support for AWS,AZURE,GCP
- Updated Create, Read, and Update functions to handle IPv6 fields
- Added proper error handling for unsupported configurations

### DIFF
--- a/aviatrix/provider.go
+++ b/aviatrix/provider.go
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-var supportedVersions = []string{"8.0"}
+var supportedVersions = []string{"8.0", "8.2"}
 
 // Provider returns a schema.Provider for Aviatrix.
 func Provider() *schema.Provider {
--- a/aviatrix/resource_aviatrix_vpc.go
+++ b/aviatrix/resource_aviatrix_vpc.go
@@ -1,11 +1,12 @@
 package aviatrix
 
 import (
+	"context"
 	"fmt"
-	"log"
 	"strings"
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
@@ -228,13 +229,14 @@ func resourceAviatrixVpc() *schema.Resource {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     false,
-				Description: "Enable IPv6 for the VPC. Only supported for Azure (8), AzureGov (32), AzureChina (2048) and GCP (4).",
+				Description: "Enable IPv6 for the VPC. Only supported for AWS (1), Azure (8)and GCP (4).",
 			},
 			"vpc_ipv6_cidr": {
 				Type:         schema.TypeString,
 				Optional:     true,
+				Computed:     true,
 				ForceNew:     true,
-				Description:  "IPv6 CIDR for the VPC. Required when enable_ipv6 is true for Azure. For GCP, required when ipv6_access_type is INTERNAL.",
+				Description:  "IPv6 CIDR for the VPC. Required when enable_ipv6 is true for Azure.For GCP, required when ipv6_access_type is INTERNAL. For AWS, this is auto-allocated and computed.",
 				ValidateFunc: validation.IsCIDR,
 			},
 			"ipv6_access_type": {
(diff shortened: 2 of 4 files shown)


aviatrix/resource_aviatrix_vpc.go
```diff
@@ -597,6 +672,22 @@ func resourceAviatrixVpcUpdate(d *schema.ResourceData, meta interface{}) error {
 		}
 	}
 
+	if d.HasChange("enable_ipv6") {
+		log.Printf("[INFO] IPv6 enablement change detected.")
```
rpanzer-aviatrix: Would it make sense to use tflog instead?
https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-logging

aviatrix/resource_aviatrix_vpc.go
```diff
@@ -624,4 +715,4 @@ func resourceAviatrixVpcDelete(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	return nil
-}
+}
```
rpanzer-aviatrix: Nit: I think we generally prefer our sources to finish with EOL.
At least the linters in cloudn check that.
Would be good to have that here too.

goaviatrix/vpc.go
```diff
@@ -77,6 +85,12 @@ func (c *Client) CreateVpc(vpc *Vpc) error {
 		"account_name": vpc.AccountName,
 		"pool_name":    vpc.Name,
 	}
+	if vpc.EnableIpv6 {
+		form["enable_ipv6"] = "true"
+	}
+	if vpc.CloudType == Azure && vpc.EnableIpv6 {
```
rpanzer-aviatrix, line 89: CloudType Azure is really only Azure, isn't it?
What about AzureGov and AzureChina?

rpanzer-aviatrix (commented):
It would be good to have a test for these changes.

goaviatrix/vpc.go
avx-shrutianand: Yes for now ipv6 is supported only for Azure

aviatrix/provider.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-var supportedVersions = []string{"8.0"}
+var supportedVersions = []string{"8.0", "8.2"}
```
rpanzer-aviatrix, line 12: Please don't merge this change.
This should be done as a separate PR with a separate ticket.
Only change it for locally for your local testing.

goaviatrix/vpc.go
rpanzer-aviatrix: So Azure China and Azure Gov are explicitly not supported?
https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2282/files#diff-34db6fe9fed9cd8010051108f54299de197ba2b8080acb69761592f1917ebfbfR369 seems to allow setting a IPv6 CIDR for these Azure Gov and Azure China, and here that property is ignored?
Wouldn't it be better then to also disallow setting this for anything but Azure itself so that customers are aware that they cannot set this property?

aviatrix/resource_aviatrix_vpc_test.go
```diff
@@ -129,10 +129,204 @@ func TestAccAviatrixVpc_basic(t *testing.T) {
...
+			},
+			Providers:    testAccProviders,
+			CheckDestroy: testAccCheckVpcDestroy,
+			Steps: []resource.TestStep{
+				{
+					Config: testAccVpcConfigIPv6AWS(rName),
+					Check: resource.ComposeTestCheckFunc(
+						testAccCheckVpcExists(resourceName, &vpc),
+						resource.TestCheckResourceAttr(resourceName, "name", fmt.Sprintf("tfg-%s-ipv6", rName)),
+						resource.TestCheckResourceAttr(resourceName, "account_name", ""),
```
rpanzer-aviatrix, line 159: The test currently fails here.
It would be good if I could run the test without modifying the code, but just by setting an environment variable.

avx-pkandoi:
closing since different PR is raised and merged

PR #2167, opened 2025-02-14
AVX-61260: Terraform for EaT external connection

jzhuavtx:
Changes to aviatrix_transit_external_device_conn resource for Edge Transit site2cloud connections:

* `enable_edge_underlay` - (Optional) Enable BGP over WAN underlay. Valid values: true, false. Default value: false.
* `disable_activemesh` - (Optional) Switch to disable ActiveMesh mode. Only valid for Edge Transit gateway BGPoIPSec and BGPoGRE connections. Valid values: true, false. Default value: false.
* `tunnel_src_ip` - (Optional) Local WAN interface IP used for Tunnel Source IP. Required for Edge Transit gateway BGPoIPSec and BGPoGRE connections.

UTed with:
EdgeTransit BGPoIPsec (2-to-2, 2-to-1)
EdgeTransit BGPoGRE (disable_activemesh, 2-to-2, 2-to-1), create with jumbo enabled, edit jumbo.
EdgeTransit BGPoLAN underlay (2-to-2, 1-to-1)
AWSTransit BGPoIPSec (1-to-2, 1-to-1)

```
resource "aviatrix_transit_external_device_conn" "eat-2-bgpoipsec-2" {
  vpc_id                    = "tsite-2"
  connection_name           = "eat-2-bgpoipsec-2"
  gw_name                   = "e2e-edge-transit-2"
  connection_type           = "bgp"
  tunnel_protocol           = "IPsec"
  bgp_local_as_num          = "65402"
  bgp_remote_as_num         = "65151"
  remote_gateway_ip         = "3.140.40.45"
  pre_shared_key         = "aviatrix,aviatrix"
  direct_connect            = true
  enable_jumbo_frame            = false
  disable_activemesh = false
  local_tunnel_cidr         = "169.254.22.54/30, 169.254.238.22/30"
  remote_tunnel_cidr        = "169.254.22.53/30, 169.254.238.21/30"
  ha_enabled                = true
  backup_local_tunnel_cidr  = "169.254.33.94/30, 169.254.165.254/30"
  backup_remote_tunnel_cidr = "169.254.33.93/30, 169.254.165.253/30"
  backup_bgp_remote_as_num  = "65151"
  backup_remote_gateway_ip  = "18.223.219.22"
  backup_direct_connect     = true
  backup_pre_shared_key         = "aviatrix,aviatrix"
  tunnel_src_ip  = "192.168.20.117,192.168.23.16"
}

[...]

PR #2298, opened 2025-08-26
Fixing the enable_ha change for external device connection

sairane10:
Enable HA value was enabled when there were two remote connections. This logic is only applicable for EAT external device connection. 
PR - https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2167/files#diff-a74471d814210d0b5fa34660c7cedc4422bccbfc1ddd00cc8b2e24d0c7796db6R268-R276

For spoke external device connection, enable HA is disabled for 2 remote connections. Updating the `get_site2cloud_conn_detail` to handle the logic for spoke and EAT device connections differently

### DIFF
--- a/goaviatrix/transit_external_device_conn.go
+++ b/goaviatrix/transit_external_device_conn.go
@@ -199,206 +199,28 @@ func (c *Client) GetExternalDeviceConnDetail(externalDeviceConn *ExternalDeviceC
 	}
 
 	externalDeviceConnDetail := data.Results.Connections
-	if len(externalDeviceConnDetail.ConnectionName) != 0 {
-		if len(externalDeviceConnDetail.VpcID) != 0 {
-			externalDeviceConn.VpcID = externalDeviceConnDetail.VpcID[0]
-		}
-
-		externalDeviceConn.ConnectionName = externalDeviceConnDetail.ConnectionName[0]
-		externalDeviceConn.GwName = externalDeviceConnDetail.GwName
-
-		if externalDeviceConnDetail.BgpStatus == "enabled" || externalDeviceConnDetail.BgpStatus == "Enabled" {
-			bgpLocalAsNumber, _ := strconv.Atoi(externalDeviceConnDetail.BgpLocalAsNum)
-			externalDeviceConn.BgpLocalAsNum = bgpLocalAsNumber
-			bgpRemoteAsNumber, _ := strconv.Atoi(externalDeviceConnDetail.BgpRemoteAsNum)
-			externalDeviceConn.BgpRemoteAsNum = bgpRemoteAsNumber
-			externalDeviceConn.ConnectionType = "bgp"
-			if len(externalDeviceConnDetail.Tunnels) >= 1 {
-				tunnelProtocol := externalDeviceConnDetail.Tunnels[0].TunnelProtocol
-				// LAN tunnel protocol is defined in the backend as "N/A(LAN)".
-				// Here we clean that up to be just "LAN" for Terraform users.
-				if strings.Contains(tunnelProtocol, "LAN") {
-					tunnelProtocol = "LAN"
-				}
-				externalDeviceConn.TunnelProtocol = tunnelProtocol
-			}
-		} else {
-			externalDeviceConn.RemoteSubnet = externalDeviceConnDetail.RemoteSubnet
-			externalDeviceConn.ConnectionType = "static"
-		}
-		externalDeviceConn.RemoteGatewayIP = strings.Split(externalDeviceConnDetail.RemoteGatewayIP, ",")[0]
-		externalDeviceConn.BgpSendCommunities = externalDeviceConnDetail.BgpSendCommunities
-
-		// GRE and LAN tunnels cannot set Algorithms
-		if externalDeviceConn.TunnelProtocol != "GRE" && externalDeviceConn.TunnelProtocol != "LAN" {
-			if externalDeviceConnDetail.Algorithm.Phase1Auth[0] == "SHA-256" &&
-				externalDeviceConnDetail.Algorithm.Phase2Auth[0] == "HMAC-SHA-256" &&
-				externalDeviceConnDetail.Algorithm.Phase1DhGroups[0] == "14" &&
-				externalDeviceConnDetail.Algorithm.Phase2DhGroups[0] == "14" &&
-				externalDeviceConnDetail.Algorithm.Phase1Encrption[0] == "AES-256-CBC" &&
-				externalDeviceConnDetail.Algorithm.Phase2Encrption[0] == "AES-256-CBC" {
-				externalDeviceConn.CustomAlgorithms = false
-				externalDeviceConn.Phase1Auth = ""
-				externalDeviceConn.Phase2Auth = ""
-				externalDeviceConn.Phase1DhGroups = ""
-				externalDeviceConn.Phase2DhGroups = ""
-				externalDeviceConn.Phase1Encryption = ""
-				externalDeviceConn.Phase2Encryption = ""
-			} else {
-				externalDeviceConn.CustomAlgorithms = true
-				externalDeviceConn.Phase1Auth = externalDeviceConnDetail.Algorithm.Phase1Auth[0]
-				externalDeviceConn.Phase2Auth = externalDeviceConnDetail.Algorithm.Phase2Auth[0]
-				externalDeviceConn.Phase1DhGroups = externalDeviceConnDetail.Algorithm.Phase1DhGroups[0]
-				externalDeviceConn.Phase2DhGroups = externalDeviceConnDetail.Algorithm.Phase2DhGroups[0]
-				externalDeviceConn.Phase1Encryption = externalDeviceConnDetail.Algorithm.Phase1Encrption[0]
-				externalDeviceConn.Phase2Encryption = externalDeviceConnDetail.Algorithm.Phase2Encrption[0]
-			}
-		}
-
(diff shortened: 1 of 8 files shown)


goaviatrix/transit_external_device_conn.go
```diff
@@ -299,13 +299,10 @@ func (c *Client) GetExternalDeviceConnDetail(externalDeviceConn *ExternalDeviceC
...
 							// two external devices, remote has HA
 							// activemesh is disabled, 2 straight tunnels only
-							externalDeviceConn.LocalTunnelCidr = externalDeviceConnDetail.LocalTunnelCidr
-							externalDeviceConn.BackupLocalTunnelCidr = externalDeviceConnDetail.BackupLocalTunnelCidr
-							externalDeviceConn.RemoteTunnelCidr = externalDeviceConnDetail.RemoteTunnelCidr
-							externalDeviceConn.BackupRemoteTunnelCidr = externalDeviceConnDetail.BackupRemoteTunnelCidr
-							externalDeviceConn.BackupRemoteGatewayIP = strings.Split(externalDeviceConnDetail.RemoteGatewayIP, ",")[1]
-							externalDeviceConn.HAEnabled = "enabled"
-							externalDeviceConn.BackupBgpRemoteAsNum = backupBgpRemoteAsNumber
+							externalDeviceConn.LocalTunnelCidr = externalDeviceConnDetail.LocalTunnelCidr + "," + externalDeviceConnDetail.BackupLocalTunnelCidr
```
shanxu2017: this change will break EaT support

sairane10: So is the previous logic applicable for tunnel_type Transit_Remote_Policy, Transit_Remote_Routed and Transit_BGP.
I can add a check in the logic for the above tunnel types and set the values accordingly

sairane10: I have updated the logic to check for transit tunnel types

goaviatrix/transit_external_device_conn.go
```diff
@@ -299,13 +300,20 @@ func (c *Client) GetExternalDeviceConnDetail(externalDeviceConn *ExternalDeviceC
...
 							// two external devices, remote has HA
 							// activemesh is disabled, 2 straight tunnels only
-							externalDeviceConn.LocalTunnelCidr = externalDeviceConnDetail.LocalTunnelCidr
-							externalDeviceConn.BackupLocalTunnelCidr = externalDeviceConnDetail.BackupLocalTunnelCidr
-							externalDeviceConn.RemoteTunnelCidr = externalDeviceConnDetail.RemoteTunnelCidr
-							externalDeviceConn.BackupRemoteTunnelCidr = externalDeviceConnDetail.BackupRemoteTunnelCidr
-							externalDeviceConn.BackupRemoteGatewayIP = strings.Split(externalDeviceConnDetail.RemoteGatewayIP, ",")[1]
-							externalDeviceConn.HAEnabled = "enabled"
-							externalDeviceConn.BackupBgpRemoteAsNum = backupBgpRemoteAsNumber
+							if strings.HasPrefix(externalDeviceConnDetail.TunnelType, "Transit") {
```
shanxu2017: This is not right. tunnel_type for EaT is also Transit_xxx. 
The right way is to check the local gateway is edge or not.

sairane10: Updated the code to use the local gateway details.

goaviatrix/transit_external_device_conn.go
```diff
@@ -623,3 +453,244 @@ func (c *Client) ConnectionBGPSendCommunities(bgpSendCommunities *BgpSendCommuni
...
+	}
+}
+
+// populateDirectConnectInfo populates direct connect information
+func populateDirectConnectInfo(externalDeviceConn *ExternalDeviceConn, externalDeviceConnDetail EditExternalDeviceConnDetail) {
+	if externalDeviceConnDetail.DirectConnect {
+		externalDeviceConn.DirectConnect = "enabled"
+	} else {
+		externalDeviceConn.DirectConnect = "disabled"
+	}
```
dnardo: There is a lot of this sort of thing where we enable or disable - Suggest a new function
```
const (
	Enabled  = "enabled"
	Disabled = "disabled"
)

func on(b bool) string {
	if b {
		return Enabled
	}
	return Disabled
}
```

usage would then be 
```
externalDeviceConn.DirectConnect = on(externalDeviceConnDetail.DirectConnect)
```

You could apply this here and elswhere we have this pattern in BackupDirectConnect,  EnableEdgeSegmentation, and EnableIkev2

goaviatrix/transit_external_device_conn.go
```diff
@@ -623,3 +453,244 @@ func (c *Client) ConnectionBGPSendCommunities(bgpSendCommunities *BgpSendCommuni
...
+	// Example2: "conn_bgp_send_communities": "block"
+	// We need to parse this field to set the BgpSendCommunities, BgpSendCommunitiesAdditive and BgpSendCommunitiesBlock fields
+	if externalDeviceConnDetail.BgpSendCommunities != "" {
+		commList := strings.Split(externalDeviceConnDetail.BgpSendCommunities, " ")
+		if len(commList) > 0 && commList[0] == "block" {
+			externalDeviceConn.BgpSendCommunitiesBlock = true
+		} else if len(commList) > 0 && commList[0] == "additive" {
+			externalDeviceConn.BgpSendCommunitiesAdditive = true
+			commList = commList[1:]
+			externalDeviceConn.BgpSendCommunities = strings.Join(commList, " ")
```
dnardo: Optional: This reads a little simpler to me and returns early on error

```
	if externalDeviceConnDetail.BgpSendCommunities == "" {
		return
	}
	parts := strings.Fields(externalDeviceConnDetail.BgpSendCommunities)
	if len(parts) == 0 {
		return
	}
	switch parts[0] {
	case "block":
		externalDeviceConn.BgpSendCommunitiesBlock = true
	case "additive":
		externalDeviceConn.BgpSendCommunitiesAdditive = true
		externalDeviceConn.BgpSendCommunities = strings.Join(parts[1:], " ")
	}
```

goaviatrix/transit_external_device_conn.go
```diff
@@ -623,3 +453,244 @@ func (c *Client) ConnectionBGPSendCommunities(bgpSendCommunities *BgpSendCommuni
...
+}
+
+// configureLocalGatewayHAEnabled configures HA when local gateway has HA enabled
+func configureLocalGatewayHAEnabled(externalDeviceConn *ExternalDeviceConn, externalDeviceConnDetail EditExternalDeviceConnDetail, localGateway *Gateway, backupBgpRemoteAsNumber int) {
+	if len(externalDeviceConnDetail.Tunnels) == 2 {
+		remoteIP := strings.Split(externalDeviceConnDetail.RemoteGatewayIP, ",")
+		if len(remoteIP) == 2 {
+			if remoteIP[0] == remoteIP[1] {
+				// one external device, no remote HA
+				externalDeviceConn.LocalTunnelCidr = externalDeviceConnDetail.LocalTunnelCidr + "," + externalDeviceConnDetail.BackupLocalTunnelCidr
```
dnardo: how about: 

```
externalDeviceConn.LocalTunnelCidr = strings.Join(
    []string{externalDeviceConnDetail.LocalTunnelCidr, externalDeviceConnDetail.BackupLocalTunnelCidr},
    ",",
)
```

or you could just move this out to its own function
func join2(a, b string) string { return strings.Join([]string{a, b}, ",") }

externalDeviceConn.LocalTunnelCidr = join2(d.LocalTunnelCidr, d.BackupLocalTunnelCidr)

goaviatrix/transit_external_device_conn.go
```diff
@@ -623,3 +453,244 @@ func (c *Client) ConnectionBGPSendCommunities(bgpSendCommunities *BgpSendCommuni
...
+	if externalDeviceConnDetail.HAEnabled == "enabled" {
+		configureLocalGatewayHAEnabled(externalDeviceConn, externalDeviceConnDetail, localGateway, backupBgpRemoteAsNumber)
+	} else {
+		configureLocalGatewayHADisabled(externalDeviceConn, externalDeviceConnDetail, backupBgpRemoteAsNumber)
+	}
+}
+
+// configureLocalGatewayHAEnabled configures HA when local gateway has HA enabled
+func configureLocalGatewayHAEnabled(externalDeviceConn *ExternalDeviceConn, externalDeviceConnDetail EditExternalDeviceConnDetail, localGateway *Gateway, backupBgpRemoteAsNumber int) {
+	if len(externalDeviceConnDetail.Tunnels) == 2 {
```
dnardo: optional:
To me this reads better as a switch,

extract the len of tunnels first:

```
nTunnels := len(d.Tunnels)
...
switch nTunnels {
	case 2:
        ....
    case 4: 
        ....
```

you could have an inner switch on the len of ips as well.

goaviatrix/transit_external_device_conn.go
sairane10: Updated

goaviatrix/transit_external_device_conn.go
sairane10: Done.

goaviatrix/transit_external_device_conn.go
sairane10: Added the switch case

goaviatrix/transit_external_device_conn.go
sairane10: done

dnardo (approved):

PR #2299, opened 2025-08-27
AVX-68228 terraform support for IPv6 AWS and Azure VPC and Gateway

avx-pkandoi:
AVX-68228 terraform support for IPv6 AWS and Azure VPC and Multi cloud Gateway

Added support for IPv6 VPC for AWS and Azure.
Added support for IPv6 spoke/transit Gateway for AWS and Azure.

### DIFF
--- a/aviatrix/resource_aviatrix_vpc.go
+++ b/aviatrix/resource_aviatrix_vpc.go
@@ -3,11 +3,11 @@ package aviatrix
 import (
 	"fmt"
 	"log"
+	"net"
 	"strings"
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
 func resourceAviatrixVpc() *schema.Resource {
@@ -232,12 +232,38 @@ func resourceAviatrixVpc() *schema.Resource {
 				Description: "Enable IPv6 for the VPC. Only supported for AWS (1), Azure (8).",
 			},
 			"vpc_ipv6_cidr": {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Computed:     true,
-				ForceNew:     true,
-				Description:  "IPv6 CIDR for the VPC. Required when enable_ipv6 is true for Azure (8).",
-				ValidateFunc: validation.IsCIDR,
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				ForceNew:    true,
+				Description: "IPv6 CIDR for the VPC. Required when enable_ipv6 is true for Azure (8).",
+				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
+					v := val.(string)
+					ip, ipnet, err := net.ParseCIDR(v)
+					if err != nil {
+						errs = append(errs, fmt.Errorf("%q must be a valid IPv6 CIDR, got: %s", key, v))
+						return
+					}
+
+					// Ensure IPv6
+					if ip.To4() != nil {
+						errs = append(errs, fmt.Errorf("%q must be an IPv6 CIDR, got IPv4: %s", key, v))
+						return
+					}
+
+					// Ensure it's the network address, not a host address
+					if !ip.Equal(ipnet.IP) {
+						errs = append(errs, fmt.Errorf("%q must be a network CIDR, not a host IP (%s)", key, v))
+					}
+
+					// Reject /128 (single host)
+					ones, _ := ipnet.Mask.Size()
+					if ones == 128 {
+						errs = append(errs, fmt.Errorf("%q cannot be /128, must be a valid IPv6 network range", key))
+					}
+
+					return
+				},
 			},
 		},
 	}
--- a/aviatrix/resource_aviatrix_vpc.go
+++ b/aviatrix/resource_aviatrix_vpc.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
 func resourceAviatrixVpc() *schema.Resource {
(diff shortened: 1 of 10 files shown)


aviatrix/resource_aviatrix_vpc.go
```diff
@@ -223,6 +224,21 @@ func resourceAviatrixVpc() *schema.Resource {
...
+				Default:     false,
+				Description: "Enable IPv6 for the VPC. Only supported for AWS (1), Azure (8).",
+			},
+			"vpc_ipv6_cidr": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
+				Description:  "IPv6 CIDR for the VPC. Required when enable_ipv6 is true for Azure (8).",
+				ValidateFunc: validation.IsCIDR,
```
dnardo: keep in mind that this accepts both ipv4 and ipv6, you may want a separate v6 only validation here

aviatrix/resource_aviatrix_vpc.go
```diff
@@ -329,6 +345,25 @@ func resourceAviatrixVpcCreate(d *schema.ResourceData, meta interface{}) error {
...
+		vpc.EnableIpv6 = true
+		log.Printf("[INFO] Enabling IPv6 in VPC: %#v", vpc)
+
+		// Handle ipv6_access_type for Azure
+		if goaviatrix.IsCloudType(vpc.CloudType, goaviatrix.AzureArmRelatedCloudTypes) {
+			if vpcIpv6Cidr, ok := d.GetOk("vpc_ipv6_cidr"); ok {
+				vpc.VpcIpv6Cidr = vpcIpv6Cidr.(string)
+			} else {
+				return fmt.Errorf("error creating vpc: valid vpc_ipv6_cidr is required when enable_ipv6 is true for Azure")
+			}
```
dnardo, line 389: You could consider moving this to 
CustomizeDiff: in the schema above 

https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/customizing-differences

```
CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, i interface{}) error {
    if d.Get("enable_ipv6").(bool) && isAzure(d) {
        if _, ok := d.GetOk("vpc_ipv6_cidr"); !ok {
            return fmt.Errorf("vpc_ipv6_cidr is required when enable_ipv6 is true for Azure")
        }
    }
    if !d.Get("enable_ipv6").(bool) && d.Get("vpc_ipv6_cidr").(string) != "" {
        return fmt.Errorf("vpc_ipv6_cidr must not be set when enable_ipv6 is false")
    }
    return nil
},
```

aviatrix/resource_aviatrix_vpc.go
```diff
@@ -223,6 +224,21 @@ func resourceAviatrixVpc() *schema.Resource {
...
+				Optional:    true,
+				ForceNew:    true,
+				Default:     false,
+				Description: "Enable IPv6 for the VPC. Only supported for AWS (1), Azure (8).",
+			},
+			"vpc_ipv6_cidr": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ForceNew:     true,
```
dnardo: so forcenew is pretty heavy handed it sounds like, so we should be sure we need it

docs/resources/aviatrix_transit_gateway.md
```diff
@@ -401,6 +401,26 @@ resource "aviatrix_transit_gateway" "edge-transit-test" {
...
+# Create an Aviatrix IPv6 Transit Network Gateway
+resource "aviatrix_transit_gateway" "test_transit_gateway_aws" {
+  cloud_type               = 1
+  account_name             = "devops_aws"
+  gw_name                  = "transit"
+  vpc_id                   = "vpc-abcd1234"
+  vpc_reg                  = "us-east-1"
+  gw_size                  = "t2.micro"
+  subnet                   = "10.1.0.0/24"
+  ha_subnet                = "10.1.0.0/24"
```
dnardo: nit in the example here both subnets are the same is that correct? should they be different?

docs/resources/aviatrix_vpc.md
```diff
@@ -142,6 +165,8 @@ The following arguments are supported:
 * `num_of_subnet_pairs` - (Optional) Number of public subnet and private subnet pair created. Only supported for AWS, Azure provider. Example: 1. Available in provider version R2.17+.
 * `enable_private_oob_subnet` - (Optional) Switch to enable private oob subnet. Only supported for AWS, AWSGov and AWSChina providers. Valid values: true, false. Default value: false. Available as of provider version R2.18+.
 * `resource_group` - (Optional) The name of an existing resource group or a new resource group to be created for the Azure VNet.  A new resource group will be created if left blank. Only available for Azure, AzureGov and AzureChina providers. Available as of provider version R2.19+.
+* `enable_ipv6` - (Optional) To enable IPv6 CIDR in VPC. Only AWS, Azure, AzureGov and AWSGov are supported.
+* `vpc_ipv6_cidr` - (Optional) To configure IPv6 CIDR in Azure VPC. Only Azure, AzureGov are supported.
```
dnardo: maybe add this if I'm correct:
```
On AWS/AWSGov, the IPv6 CIDR is automatically assigned and this field is ignored.
```

aviatrix/resource_aviatrix_vpc.go
avx-pkandoi: yes we need it because VPC can't be updated with new CIDR. so if some one is changing the IPv6 cidr it should be new vpc.

docs/resources/aviatrix_transit_gateway.md
avx-pkandoi: done copy/paste issue. fixed in both places. but technically its possible but not best practise.

aviatrix/resource_aviatrix_vpc.go
avx-pkandoi: Didn't see any example in current code where we used CustomizeDiff in schema.
Maybe we follow above method otherwise schema may clutter?

aviatrix/resource_aviatrix_vpc.go
avx-pkandoi: yes fixed it by adding own function.

aviatrix/resource_aviatrix_vpc.go
dnardo: its more about failing earlier in the plan process if something is wrong.

dnardo (approved):

PR #2319, opened 2025-09-03
AVX-68208: Create a new resource user group membership

dnardo:
The **aviatrix_rbac_group_user_membership** resource allows the creation and management of user membership for Aviatrix  RBAC groups. This resource is authoritative for the group's user membership, meaning it manages the complete set of users that belong to a specific group.

**Note:** There is another related resource, `aviatrix_rbac_group_user_attachment` which manages a single user-to-group attachment per resource. While that resource is still available it would be preferable to use this new resource especially
for larger sets of users or groups.

### DIFF
--- a/aviatrix/resource_aviatrix_rbac_group_user_membership.go
+++ b/aviatrix/resource_aviatrix_rbac_group_user_membership.go
@@ -1,0 +1,166 @@
+package aviatrix
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"slices"
+	"sort"
+	"strings"
+
+	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceAviatrixRbacGroupUserMembership() *schema.Resource {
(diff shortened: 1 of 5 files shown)


aviatrix/resource_aviatrix_rbac_group_user_membership.go
```diff
@@ -0,0 +1,166 @@
...
+		desired := expandStringSet(d.Get("user_names").(*schema.Set))
+		if err := client.SetRbacGroupUsers(group, desired); err != nil {
+			return fmt.Errorf("failed to update users for RBAC group %q: %w", group, err)
+		}
+	}
+
+	return resourceAviatrixRbacGroupUserMembershipRead(d, meta)
+}
+
+func resourceAviatrixRbacGroupUserMembershipDelete(d *schema.ResourceData, meta interface{}) error {
```
tommy-carver, line 117: If this function is for delete of the resource, should we remove all the users from the rbac group, even if `remove_users_on_destroy` is false. IE, that flag says whether to delete or just remove the users from the group.

tommy-carver (commented):
Just one comment

tommy-carver (approved):

PR #2349, opened 2025-09-19
AVX-68970: Add TLS profile resource

kasunkarunaratne:
In 8.2 we are exposing TLS profile creation CRUD APIs. This PR should provide TF support for it.


avtx-akonnur:
Please add docs and test to your TF resource.

avtx-akonnur (dismissed):

goaviatrix/dcf_tls_profile.go
```diff
@@ -0,0 +1,107 @@
...
+	UUID string `json:"uuid"`
+}
+
+type TLSProfilesListResponse struct {
+	// Profiles List of all TLS profiles
+	Profiles []TLSProfileWithID `json:"profiles"`
+}
+
+func (c *Client) GetTLSProfile(ctx context.Context, uuid string) (*TLSProfile, error) {
+	endpoint := fmt.Sprintf("dcf/tls-profile/%s", uuid)
```
stewartbutler: UUID here can be an arbitrary string, can we parse it or otherwise sanitize it to make sure we aren't getting a URL injection?

goaviatrix/dcf_tls_profile.go
```diff
@@ -0,0 +1,107 @@
...
+	err := c.GetAPIContext25(ctx, &listResponse, endpoint, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return &listResponse, nil
+}
+
+func (c *Client) UpdateDCFPolicy(ctx context.Context, uuid string, tlsProfile *TLSProfile) error {
+	endpoint := fmt.Sprintf("dcf/tls-profile/%s", uuid)
```
stewartbutler: Same here

goaviatrix/dcf_tls_profile.go
```diff
@@ -0,0 +1,107 @@
...
+	err := c.PutAPIContext25(ctx, endpoint, tlsProfile)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Client) UpdateTLSProfile(ctx context.Context, uuid string, tlsProfile *TLSProfile) error {
+	endpoint := fmt.Sprintf("dcf/tls-profile/%s", uuid)
```
stewartbutler: Same here

goaviatrix/dcf_tls_profile.go
```diff
@@ -0,0 +1,107 @@
...
+	err := c.PutAPIContext25(ctx, endpoint, tlsProfile)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (c *Client) DeleteTLSProfile(ctx context.Context, uuid string) error {
+	endpoint := fmt.Sprintf("dcf/tls-profile/%s", uuid)
```
stewartbutler: Same here

docs/resources/aviatrix_dcf_tls_profile.md
```diff
@@ -0,0 +1,85 @@
...
+}
+```
+
+```hcl
+# Create an Aviatrix Distributed Firewalling TLS Profile with custom CA bundle
+resource "aviatrix_dcf_tls_profile" "with_ca_bundle" {
+  display_name           = "ca-bundle-tls-profile"
+  certificate_validation = "CERTIFICATE_VALIDATION_ENFORCE"
+  verify_sni            = true
+  ca_bundle_id          = "f15c9890-c8c4-4c1a-a2b5-ef0ab34d2e30"
```
stewartbutler: I don't like the idea of passing around UUIDs like this -- where does this value come from? If it is coming from another resource or data source, that resource/datasource identifier should be the input, not a random UUID string.

aviatrix/resource_aviatrix_dcf_tls_profile_test.go
```diff
@@ -0,0 +1,165 @@
...
+	if !ok {
+		return fmt.Errorf("failed to assert Meta as *goaviatrix.Client")
+	}
+
+	for _, rs := range s.RootModule().Resources {
+		if rs.Type != "aviatrix_dcf_tls_profile" {
+			continue
+		}
+
+		_, err := client.GetTLSProfile(context.Background(), rs.Primary.ID)
```
stewartbutler: I don't think this is a great pattern. Can you try this instead?

1) Create a helper function, `generateAcceptanceCheckFunctionDcfTLSProfileDestroy`, that takes in a `testing.T` and returns a `func(*terraform.State) error`. 
2) In that helper function, create a closure that runs `t.Context()` to get a test-scoped context for use in these calls, and then return the contents of this helper function with this `context.Background` swapped out.

I'm ambivalent about returning an error on casting failure -- if this absolutely should never happen, you might also want to use `testify.Assert` with the `testing.T` instead of returning the `failed to assert Meta as *goaviatrix.Client` error.

Also update the other helper functions here to use a similar pattern.

stewartbutler: Also, is the return value here really irrelevant so long as it isn't a failure, or should you also be doing sanity checks to make sure this matches your expectations?

aviatrix/resource_aviatrix_dcf_tls_profile_test.go
```diff
@@ -0,0 +1,165 @@
...
+resource "aviatrix_dcf_tls_profile" "test" {
+	display_name           = "test-dcf-tls-profile-with-ca"
+	certificate_validation = "CERTIFICATE_VALIDATION_ENFORCE"
+	verify_sni            = true
+	ca_bundle_id          = "test-ca-bundle-uuid"
+}
+`
+}
+
+func testAccCheckDcfTLSProfileExists(n string) resource.TestCheckFunc {
```
stewartbutler: Same as down below, add a `testing.T` parameter to this closure and use that to grab the testing context instead of using Background

aviatrix/resource_aviatrix_dcf_tls_profile.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func resourceAviatrixDCFTLSProfile() *schema.Resource {
+	return &schema.Resource{
+		CreateWithoutTimeout: resourceAviatrixDCFTLSProfileCreate,
+		ReadWithoutTimeout:   resourceAviatrixDCFTLSProfileRead,
+		UpdateWithoutTimeout: resourceAviatrixDCFTLSProfileUpdate,
+		DeleteWithoutTimeout: resourceAviatrixDCFTLSProfileDelete,
```
stewartbutler: These should be CreateContext, etc. The resource logic isn't doing any complicated synchronization logic AFAIK, so this should be context-based with a sane timeout.

aviatrix/resource_aviatrix_dcf_tls_profile.go
```diff
@@ -0,0 +1,179 @@
...
+				Type:        schema.TypeBool,
+				Required:    true,
+				Description: "Toggle to enable advanced SNI verification of client connections.",
+			},
+			"ca_bundle_id": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "UUID of the CA bundle that should be used for origin certificate validation. If not populated the default bundle would be used.",
+			},
+			"uuid": {
```
stewartbutler, line 46: As mentioned below this indicates something wrong with the API design to me -- a user should not have to pass around a UUID in any circumstance unless it is externally generated (e.g. a UUID provided by a third party service).

aviatrix/resource_aviatrix_dcf_tls_profile.go
```diff
@@ -0,0 +1,179 @@
...
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"CERTIFICATE_VALIDATION_NONE", "CERTIFICATE_VALIDATION_LOG_ONLY", "CERTIFICATE_VALIDATION_ENFORCE"}, false),
+				Description:  "Certificate validation mode. Must be one of CERTIFICATE_VALIDATION_NONE, CERTIFICATE_VALIDATION_LOG_ONLY, or CERTIFICATE_VALIDATION_ENFORCE.",
+			},
+			"verify_sni": {
+				Type:        schema.TypeBool,
+				Required:    true,
+				Description: "Toggle to enable advanced SNI verification of client connections.",
+			},
+			"ca_bundle_id": {
```
stewartbutler, line 41: Same as below here -- no UUIDs

aviatrix/resource_aviatrix_dcf_tls_profile.go
```diff
@@ -0,0 +1,179 @@
...
+			"display_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Display name for the TLS profile.",
+			},
+			"certificate_validation": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"CERTIFICATE_VALIDATION_NONE", "CERTIFICATE_VALIDATION_LOG_ONLY", "CERTIFICATE_VALIDATION_ENFORCE"}, false),
+				Description:  "Certificate validation mode. Must be one of CERTIFICATE_VALIDATION_NONE, CERTIFICATE_VALIDATION_LOG_ONLY, or CERTIFICATE_VALIDATION_ENFORCE.",
```
stewartbutler, line 34: Is this restriction coming from the API backend?

aviatrix/resource_aviatrix_dcf_tls_profile.go
kasunkarunaratne: Hm, not sure what you mean. Just following pattern of other resources - see https://github.com/AviatrixSystems/terraform-provider-aviatrix/blob/70e02b4e7f6071701cf091aa99933a82ddc27489/aviatrix/resource_aviatrix_dcf_mwp_policy_list.go#L17

aviatrix/resource_aviatrix_dcf_tls_profile.go
kasunkarunaratne: yes

aviatrix/resource_aviatrix_dcf_tls_profile.go
kasunkarunaratne: right, this is the API. Not something we can change in terraform right. The API here does take in UUID. That is how other APIs like `microseg/policy-list` work. The UI does the display_name -> UUID mapping. For TF for now it will take the UUID directly.

aviatrix/resource_aviatrix_dcf_tls_profile.go
kasunkarunaratne: https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2349/files#r2383376333

docs/resources/aviatrix_dcf_tls_profile.md
kasunkarunaratne: https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2349/files#r2383376333, yes we should use the datasource when we add terraform for dcf trustbunlde operations

goaviatrix/dcf_tls_profile.go
kasunkarunaratne: i added this parsing check when we unmarshal the resource code

aviatrix/resource_aviatrix_dcf_tls_profile_test.go
kasunkarunaratne: looking at other tests this is how it seems to be done

aviatrix/resource_aviatrix_dcf_tls_profile_test.go
kasunkarunaratne: following pattern of other tests, many use context.Background()

aviatrix/resource_aviatrix_dcf_tls_profile.go
stewartbutler: From the docs: ```
	// Most resources should prefer CreateContext with properly implemented
	// operation timeout values, however there are cases where operation
	// synchronization across concurrent resources is necessary in the resource
	// logic, such as a mutex, to prevent remote system errors. Since these
	// operations would have an indeterminate timeout that scales with the
	// number of resources, this allows resources to control timeout behavior.
	```
	
See https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Resource
	
My point is that in other places in the codebase, the `*WithoutTimeout` code is mostly used without consideration of whether it is a good idea, same with using `context.Background()`. Without it, you wind up with stuff surviving beyond the lifespan of their calling resources. IMO, *most* things we do should be done with a timeout, unless there's some architectural reason why that isn't feasible. 

I don't think this API has anything going on that would prevent you from implementing a timeout, here.

aviatrix/resource_aviatrix_dcf_tls_profile.go
stewartbutler: Do we have a list of more than a few approved UUIDs at this point? If not I'd rather you use a display  name -> uuid mapping in the code as well.

aviatrix/resource_aviatrix_dcf_tls_profile_test.go
stewartbutler: Right, I just don't like using `context.Background()` if there's a reasonable way to provide a context that will limit the scope down to the lifespan of the test. We already use the closure pattern elsewhere, this would just use it to give a testing.T-scoped context for this call.

aviatrix/resource_aviatrix_dcf_tls_profile.go
kasunkarunaratne: yeh, UUID can be generated when customer uploads trustbundle. I think this needs to be fixed with datasource for trustbundle when we add terraform for that ? I don't see how we can fix this here

aviatrix/resource_aviatrix_dcf_tls_profile.go
kasunkarunaratne: fixed, cytal pls , thanks

aviatrix/resource_aviatrix_dcf_tls_profile_test.go
kasunkarunaratne: created a context.Withtimeout for one, and t.Context() for the other

docs/resources/aviatrix_dcf_tls_profile.md
avtx-akonnur: Data source is now added for dcf trustbundle which can be used here.

aviatrix/resource_aviatrix_dcf_tls_profile.go
stewartbutler: Is this still true now that those datasources are added?

aviatrix/resource_aviatrix_dcf_tls_profile.go
stewartbutler: Is this still true now that those datasources are added?

aviatrix/resource_aviatrix_dcf_tls_profile.go
stewartbutler: May want a datasource or some other enum here rather than a string type.

stewartbutler (approved):

PR #2350, opened 2025-09-19
AVX-54809: Adding a data source for webgroups to search by name

avtx-akonnur:
The webgroups can now be queried by name using this data source. They will return the Selector and UUID of the webgroups.

Example usage:

```tf
data "aviatrix_web_group" "foo" {
  name = "testing-ad"
}

resource "aviatrix_distributed_firewalling_policy_list" "testing1" {
    policies {
    name                     = "df-policy-1"
    action                   = "DEEP_PACKET_INSPECTION_PERMIT"
    priority                 = 1
    protocol                 = "ANY"
    logging                  = false
    watch                    = false
    exclude_sg_orchestration = true
    src_smart_groups         = [
      "def000ad-0000-0000-0000-000000000000"
    ]
    dst_smart_groups         = [
      "def000ad-0000-0000-0000-000000000000"
    ]
    web_groups = [
      data.aviatrix_web_group.foo.id
    ]
    flow_app_requirement     = "TLS_REQUIRED"
    decrypt_policy           = "DECRYPT_ALLOWED"
    port_ranges {
      hi = 50000
      lo = 49000
    }
  }
}
```

### DIFF
--- a/aviatrix/data_source_aviatrix_dcf_webgroups.go
+++ b/aviatrix/data_source_aviatrix_dcf_webgroups.go
@@ -1,0 +1,103 @@
+package aviatrix
+
+import (
+	"context"
+
+	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceAviatrixDcfWebgroups() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceAviatrixDcfWebgroupsRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the Web Group.",
+			},
+			"selector": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"match_expressions": {
+							Type:     schema.TypeList,
+							Computed: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"snifilter": {
+										Type:        schema.TypeString,
+										Computed:    true,
+										Description: "Server name indicator this expression matches.",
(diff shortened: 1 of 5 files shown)


aviatrix/data_source_aviatrix_dcf_webgroups.go
```diff
@@ -0,0 +1,103 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the Web Group.",
+			},
+			"selector": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"match_expressions": {
```
dnardo, line 25: Does this work ok as a type list?  If you re-apply after applying do you see any ordering issues where terraform  "thinks" there is a change  do to the items being reordered?  This would be the controller for instance returning in a different order. If so this probably should be a set, or you need to sort below.

docs/data-sources/aviatrix_web_group.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Attribute Reference
+
+The following attributes are exported:
+
+* `name` - Name of the Web Group.
+* `uuid` - UUID of the Web Group.
+* `selector` - Block containing match expressions to filter the Web Group.
+        * `match_expressions` - List of match expressions for the Web Group.
+                * `snifilter` - Server name indicator this expression matches.
```
dnardo: Nit: 
```
Server Name Indication (SNI) value this expression matches.
```

aviatrix/data_source_aviatrix_dcf_webgroups.go
avtx-akonnur: Yes, this works fine as the order is returned same from the code. To be consistent with how it is defined in the resource I will have to keep it as a List, but will follow up with this to see if we can change this to a Set without impacting customers. I think for the scope of this PR we can leave it as a List.

dnardo: Ok, I'll mention that I have fixed at least 3 bugs resulting for order being different and forcing churn on the customer side.  Just keep that in mind.

dnardo (approved):

PR #2354, opened 2025-09-24
AVX-43334 Added the TF support for the IPS Profile and Rule Feeds

xc-6:
This PR adds support for IPS Profiles and Rule Feeds, and enables setting an IPS Profile for a VPC.

### Test
```
aviatrix_dcf_ips_rule_feed.custom_threats: Creating...
aviatrix_dcf_ips_rule_feed.malware_rules: Creating...
aviatrix_dcf_ips_rule_feed.custom_threats: Creation complete after 0s [id=37c3a8c7-a2e5-47f2-981a-71646932370e]
aviatrix_dcf_ips_rule_feed.malware_rules: Creation complete after 0s [id=9ea699b3-dbd7-4941-bdca-8c0ec2afdbdd]
aviatrix_dcf_ips_profile.production_profile: Creating...
aviatrix_dcf_ips_profile.production_profile: Creation complete after 0s [id=45865461-46d0-4dcb-87ae-251f411a1d86]
aviatrix_dcf_ips_profile_vpc.vpc_association: Creating...
aviatrix_dcf_ips_profile_vpc.vpc_association: Creation complete after 0s [id=vpc-09853caa138265cbf]

Apply complete! Resources: 4 added, 0 changed, 0 destroyed.
```

```
xcheng@xcheng-vm:~/go/src/github.com/AviatrixSystems/terraform-provider-aviatrix$ TF_ACC=1 go test -v ./aviatrix -run '^TestAccAviatrixDCFIps'
=== RUN   TestAccAviatrixDCFIpsRuleFeed_basic
--- PASS: TestAccAviatrixDCFIpsRuleFeed_basic (11.84s)
=== RUN   TestAccAviatrixDCFIpsProfile_basic
--- PASS: TestAccAviatrixDCFIpsProfile_basic (12.02s)
=== RUN   TestAccAviatrixDCFIpsProfileVpc_basic
--- PASS: TestAccAviatrixDCFIpsProfileVpc_basic (12.85s)
PASS
ok      github.com/AviatrixSystems/terraform-provider-aviatrix/v3/aviatrix 36.748s
```


avtx-akonnur (commented):
This is a nice PR!
Will we have data sources in the future for the rule feed and IPS profile resources since they are referenced by UUIDs?

xc-6:
> This is a nice PR! Will we have data sources in the future for the rule feed and IPS profile resources since they are referenced by UUIDs?

That's a good point! Yes, we should add the data for them. The current API does not support fetching by name yet, but we'll include that in a later PR.

aviatrix/resource_aviatrix_dcf_ips_profile.go
```diff
@@ -0,0 +1,179 @@
...
+							Type:        schema.TypeList,
+							Optional:    true,
+							Description: "List of external rule feed IDs.",
+							Elem:        &schema.Schema{Type: schema.TypeString},
+						},
+						"ignored_sids": {
+							Type:        schema.TypeList,
+							Optional:    true,
+							Description: "List of rule SIDs to ignore.",
+							Elem:        &schema.Schema{Type: schema.TypeInt},
```
dnardo, line 57: We use TypeList here keep in mind that if these come in with a different order, terraform will perceive it as a diff and want to make changes.  You can avoid this by using TypeSet instead.

aviatrix/resource_aviatrix_dcf_ips_profile.go
```diff
@@ -0,0 +1,179 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name of the IPS profile.",
+			},
+			"uuid": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "UUID of the IPS profile.",
+			},
+			"rule_feeds": {
```
dnardo, line 34: Should `rule_feeds` be Computed true also?  `rule_feeds` currently reappears even when omitted if the user were to omit it. If the user committed that stanza would they  not see a perpetual diff?

aviatrix/resource_aviatrix_dcf_ips_profile.go
```diff
@@ -0,0 +1,179 @@
...
+					},
+				},
+			},
+			"intrusion_actions": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "Actions for different severity levels.",
+				Elem: &schema.Schema{
+					Type:         schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{"alert", "alert_and_drop"}, false),
```
dnardo, line 68: We validate the values, but not the valid severerties - should we? 
Meaning can the user just do:
```
intrusion_actions = { 
     flibbertygibbet = "alert" 
}
```

aviatrix/resource_aviatrix_dcf_ips_profile_vpc.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"vpc_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "The VPC ID to assign DCF IPS profiles to.",
+			},
+			"dcf_ips_profiles": {
```
dnardo, line 31: Similar comment about ordering.  If these ever change order on return they will cause a diff.

But also, why are we forcing the user to type in UUIDs?  Shouldn't we have a friendly name? This feels like a bad UX and seems pretty error prone if you are a user.

aviatrix/resource_aviatrix_dcf_ips_rule_feed.go
```diff
@@ -0,0 +1,119 @@
...
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"feed_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Name for the rule feed.",
+			},
+			"file_content": {
```
dnardo, line 30: The full rules blob ends up in plan/state should we mark `file_content` as `sensitive` to avoid leaking it. Also, how large can this get? Consider supporting a file_path (mutually exclusive with file_content) so users can keep rules in a separate file.

aviatrix/resource_aviatrix_dcf_ips_rule_feed.go
```diff
@@ -0,0 +1,119 @@
...
+	d.Set("uuid", ruleFeed.UUID)
+	d.Set("feed_name", ruleFeed.FeedName)
+	d.Set("content_hash", ruleFeed.ContentHash)
+	d.Set("ips_rules", ruleFeed.IpsRules)
+
+	return nil
+}
+
+func resourceAviatrixDCFIpsRuleFeedUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*goaviatrix.Client)
```
dnardo, line 100: Should the update only be done if there are changes to the plan?

``` 
if !d.HasChanges("feed_name", "file_content", "file_path")
...
```

docs/resources/aviatrix_dcf_ips_profile.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+### Required
+- `profile_name` - (Required) Name of the IPS profile. Type: String.
+
+### Optional
+- `rule_feeds` - (Required) Rule feeds configuration block:
```
dnardo: You say (Required) but its' in the Optional section and marked optional i the schema.

aviatrix/resource_aviatrix_dcf_ips_test.go
```diff
@@ -0,0 +1,451 @@
+//revive:disable:var-naming
+package aviatrix
```
dnardo, line 2: Did you run this manually?

docs/resources/aviatrix_dcf_ips_profile.md
xc-6: Oh right. I think the `rule_feeds` should be optional. rule_feeds is required when creating a profile, but not required for updates

aviatrix/resource_aviatrix_dcf_ips_test.go
xc-6: Yes. I have run this manually before. Will try to address some of your comments today if I have time. Thanks for the all good comments.

aviatrix/resource_aviatrix_dcf_ips_profile.go
xc-6: Thanks! Added the validateIntrusionActionsKeys below.
```
xcheng@xcheng-vm:~/src/test_ips$ terraform plan
╷
│ Error: invalid key for intrusion_actions: 'mino'. Allowed keys are: informational, minor, major, critical
```

aviatrix/resource_aviatrix_dcf_ips_profile_vpc.go
xc-6: On UI, the customer will select by name, then API request will send uuid.

On TF, the user can use like this 
```
resource "aviatrix_dcf_ips_profile_vpc" "vpc_association" {
  vpc_id       = "vpc-09853caa138265cbf" # replace with your actual VPC ID
  dcf_ips_profiles = ["${aviatrix_dcf_ips_profile.production_profile.uuid}"]
}
```

But right maybe input by ips_profile_name will be easier to them. Let me talk this with Susan see if we should also update it by name not uuid. If yes, then we need to update the backend API logic at first, then update the TF. But Let's make this PR in at first? Later if we want to change, I will change the TF also. Thanks!

aviatrix/resource_aviatrix_dcf_ips_rule_feed.go
xc-6: It won't be too large. Less than 16 MB.
The user will upload the file then UI will put it in json.
On TF, I try we can use the file path already
```
resource "aviatrix_dcf_ips_rule_feed" "malware_rules" {
  feed_name    = "update_rules-new"
  file_content = file("${path.module}/test.rules")
}
```

aviatrix/resource_aviatrix_dcf_ips_rule_feed.go
xc-6: Thanks! updated.

aviatrix/resource_aviatrix_dcf_ips_profile.go
xc-6: Thanks! It should be required. I updated the code from `optional` to `required` In this resource, rule_feeds is a required user input, not computed by the provider.

If omitted, Terraform should show a diff since the field is required.

aviatrix/resource_aviatrix_dcf_ips_profile_vpc.go
dnardo: Ok, I would think this is a lot of friction for the user, forcing the user to deal with things like UUIDs is generally not a great UX.

Changing this later IMO could cause more friction later. Usually when folks tell me "lets get this PR in  first" I never see the followup, so I'm a bit reluctant to see this go in this way. I think PM should give some feedback @Dennizz

aviatrix/resource_aviatrix_dcf_ips_rule_feed.go
dnardo: 16MB is pretty big right? As long a the customer has a way to manage this externally that's fine. we should make sure its well documented.

aviatrix/resource_aviatrix_dcf_ips_profile_vpc.go
shinrich: The problem with flipping to name here is that all the other DCF API's use UUIDs for cross references. We provide data sources in some cases to fetch the UUID from objects created in different TF workspaces.
By using UUIDs for references instead of names means that the customer has more freedom to update names later if they see fit.

dnardo (approved):

aviatrix/resource_aviatrix_dcf_ips_rule_feed.go
xc-6: Yes, I have already mentioned how to use the file in the dcf_ips_fule_feed README.

PR #2356, opened 2025-09-29
AVX-65624: The fix to have ext_args region separate to the input region

avtx-akonnur:
We need to remove the extArgs prefix used in terraform since this is not how external group arguments are seen on the backend API.

If we see an external key in the Terraform, then all the other keys must be set to the extArgs values instead of normal smart group expressions. The normal smart group expressions must all be set to null.
This is now handled correctly, and removing the extArgs prefix for all fields make sure the backend works well.


goaviatrix/smart_group.go
```diff
@@ -169,7 +169,7 @@ func smartGroupFilterToMapBasic(filter *SmartGroupMatchExpression, keepMaps bool
 		}
 		if len(filter.ExtArgs) > 0 {
 			for key, value := range filter.ExtArgs {
-				filterMap[key] = value
+				filterMap[fmt.Sprintf("%s.%s", ExtArgsPrefix, key)] = value
```
dnardo: Should there a test a change? Or do we have no tests for this?

avtx-akonnur: We don't have any particular tests for this. It was tested by running the TF and seeing if changes were applied correctly.

dnardo (dismissed):

kasunkarunaratne (dismissed):
lgtm

rpanzer-aviatrix (approved):

PR #2378, opened 2025-10-09
AVX-67564: fix tag state persistence for gw launch

sw-avtx:
Tags were not being saved in TF state properly for GWs. Seems to be some issue with the return structure from list_vpcs_summary either not returning them or putting them under a different key but I swapped it to existing client calls used in the data sources instead. Works just fine.

Validated with spoke GW deployment, no diff in state when running `terraform plan` afterwards.

### DIFF
--- a/aviatrix/resource_aviatrix_gateway.go
+++ b/aviatrix/resource_aviatrix_gateway.go
@@ -1473,7 +1473,10 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("insane_mode_az", "")
 	}
 
-	setGatewayTags(d, client, gw.CloudType, ignoreTagsConfig)
+	err = setGatewayTags(d, client, gw.CloudType, ignoreTagsConfig)
+	if err != nil {
+		return fmt.Errorf("failed to set tags for gateway %s: %v", gw.GwName, err)
+	}
 
 	if gw.VpnStatus == "enabled" && gw.SplitTunnel == "yes" {
 		d.Set("name_servers", gw.NameServers)
--- a/aviatrix/resource_aviatrix_gateway.go
+++ b/aviatrix/resource_aviatrix_gateway.go
@@ -1474,9 +1474,21 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.AWSRelatedCloudTypes|goaviatrix.AzureArmRelatedCloudTypes) {
-		tags := goaviatrix.KeyValueTags(gw.Tags).IgnoreConfig(ignoreTagsConfig)
-		if err := d.Set("tags", tags); err != nil {
-			log.Printf("[WARN] Error setting tags for (%s): %s", d.Id(), err)
+		tags := &goaviatrix.Tags{
+			ResourceType: "gw",
+			ResourceName: d.Get("gw_name").(string),
+			CloudType:    gw.CloudType,
+		}
+
+		_, err := client.GetTags(tags)
+		if err != nil {
+			log.Printf("[WARN] Failed to get tags for gateway %s: %v", tags.ResourceName, err)
+		}
+		if len(tags.Tags) > 0 {
+			tagsMap := goaviatrix.KeyValueTags(tags.Tags).IgnoreConfig(ignoreTagsConfig)
+			if err := d.Set("tags", tagsMap); err != nil {
+				log.Printf("[WARN] Error setting tags for gateway %s: %v", tags.ResourceName, err)
+			}
 		}
 	}
 
--- a/aviatrix/resource_aviatrix_gateway.go
+++ b/aviatrix/resource_aviatrix_gateway.go
@@ -1473,24 +1473,7 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
 		d.Set("insane_mode_az", "")
 	}
 
-	if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.AWSRelatedCloudTypes|goaviatrix.AzureArmRelatedCloudTypes) {
-		tags := &goaviatrix.Tags{
-			ResourceType: "gw",
-			ResourceName: d.Get("gw_name").(string),
-			CloudType:    gw.CloudType,
-		}
-
-		_, err := client.GetTags(tags)
-		if err != nil {
-			log.Printf("[WARN] Failed to get tags for gateway %s: %v", tags.ResourceName, err)
-		}
-		if len(tags.Tags) > 0 {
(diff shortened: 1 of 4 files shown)


jpeach (commented):
> Seems to be some issue with the return structure from list_vpcs_summary either not returning them or putting them under a different key

Is there a related cloudn fix for this controller bug?

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1474,9 +1474,21 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
...
+
+		_, err := client.GetTags(tags)
+		if err != nil {
+			log.Printf("[WARN] Failed to get tags for gateway %s: %v", tags.ResourceName, err)
+		}
+		if len(tags.Tags) > 0 {
+			tagsMap := goaviatrix.KeyValueTags(tags.Tags).IgnoreConfig(ignoreTagsConfig)
+			if err := d.Set("tags", tagsMap); err != nil {
+				log.Printf("[WARN] Error setting tags for gateway %s: %v", tags.ResourceName, err)
+			}
```
jpeach: We have three copies of the same code here. It would be good to consolidate into one function that is called when the relevant API doesn't return the flags. Comment and link to the cloudn tracking issues. Ideally, this code should adapt to when cloudn fixes the bugs in new releases.

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1474,9 +1474,21 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if goaviatrix.IsCloudType(gw.CloudType, goaviatrix.AWSRelatedCloudTypes|goaviatrix.AzureArmRelatedCloudTypes) {
```
jpeach: Why is fetching tags gated on these particular cloud types? IIUC google supports tags too. Suggest we always record the flags if we get them.

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1474,9 +1474,21 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
...
+		}
+
+		_, err := client.GetTags(tags)
+		if err != nil {
+			log.Printf("[WARN] Failed to get tags for gateway %s: %v", tags.ResourceName, err)
+		}
+		if len(tags.Tags) > 0 {
+			tagsMap := goaviatrix.KeyValueTags(tags.Tags).IgnoreConfig(ignoreTagsConfig)
+			if err := d.Set("tags", tagsMap); err != nil {
+				log.Printf("[WARN] Error setting tags for gateway %s: %v", tags.ResourceName, err)
```
jpeach: This seems like an actual error. If we tried to persist something in state but couldn't, that seems a lot like we wrote the code wrong.

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1474,9 +1474,21 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
...
-		tags := goaviatrix.KeyValueTags(gw.Tags).IgnoreConfig(ignoreTagsConfig)
-		if err := d.Set("tags", tags); err != nil {
-			log.Printf("[WARN] Error setting tags for (%s): %s", d.Id(), err)
+		tags := &goaviatrix.Tags{
+			ResourceType: "gw",
+			ResourceName: d.Get("gw_name").(string),
+			CloudType:    gw.CloudType,
+		}
+
+		_, err := client.GetTags(tags)
```
jpeach: This API puts its result in its argument? Who thought that was a good idea?!?!?!

aviatrix/resource_aviatrix_gateway.go
```diff
@@ -1474,9 +1474,21 @@ func resourceAviatrixGatewayRead(d *schema.ResourceData, meta interface{}) error
...
+		if err != nil {
+			log.Printf("[WARN] Failed to get tags for gateway %s: %v", tags.ResourceName, err)
```
jpeach: Comment to explain why it's ok to ignore this error.

sw-avtx:
> > Seems to be some issue with the return structure from list_vpcs_summary either not returning them or putting them under a different key
> 
> Is there a related cloudn fix for this controller bug?

I don't think theres a bug on controller but I didn't look very deep at the response as I found the direct alternative API beforehand. This behavior has been around in terraform for a long time I heard from multiple people and list_vpcs_summary is one of the most heavily used read APIs on the controller so if it had a bug we would have heard about it. Most likely the original implementation on provider side was wrong.

aviatrix/resource_aviatrix_gateway.go
sw-avtx: Don't know this was the original behavior. Would prefer not to update as it could cause unknown behavior outside the scope of the bug fix without further testing.

aviatrix/resource_aviatrix_gateway.go
sw-avtx: Agreed but previous behavior only logs a warning which is why I preserved it. Fetching tags is not a real failure case for GW launch it will just cause the state to be out of sync. Might be better to leave as is but let me know what you think. Happy to update it.

aviatrix/resource_aviatrix_gateway.go
sw-avtx: Do you mean a function with all the code in the if statement + the if statement? Like a "set tags to data" func which would be passed `tags` and `d`? Just wasnt sure based on your wording.

sw-avtx: ah nvm you mean each full diff. had to double check what i wrote its been a few days and a few PRs

aviatrix/utils.go
```diff
@@ -559,3 +559,33 @@ func expandIntList(list []interface{}) []int {
...
+		log.Printf("[WARN] Failed to get tags for gateway %s: %v", tags.ResourceName, err)
+		return
+	}
+
+	if len(tags.Tags) > 0 {
+		tagsMap := goaviatrix.KeyValueTags(tags.Tags).IgnoreConfig(ignoreTagsConfig)
+		if err := d.Set("tags", tagsMap); err != nil {
+			log.Printf("[WARN] Error setting tags for gateway %s: %v", tags.ResourceName, err)
+		}
+	}
```
jpeach, line 589:
```suggestion
		if err := d.Set("tags", goaviatrix.KeyValueTags(tags.Tags).IgnoreConfig(ignoreTagsConfig)); err != nil {
			return err
		}
```

IIUC, (*ResourceData).Set can only fail if we wrote the provider wrong, so propagating the error seems appropriate.

The tags could become empty after the filtering, so checking the len would not prevent storing empty tags. If the tags are actually empty, is seems OK to store that they are empty (otherwie maybe we could have stale tags)?

jpeach (commented):
Just a nit about the resource data, otherwise looks good.

jpeach (approved):

PR #2387, opened 2025-10-11
AVX-68592 ipv6 cidr info is missing from terraform.tfstate dump

avx-pkandoi:
Added IPv6 cidr in subnet info for VPC.

controller PR - https://github.com/AviatrixDev/cloudn/pull/49641

### DIFF
--- a/aviatrix/resource_aviatrix_vpc.go
+++ b/aviatrix/resource_aviatrix_vpc.go
@@ -132,6 +132,11 @@ func resourceAviatrixVpc() *schema.Resource {
 							Computed:    true,
 							Description: "Subnet ID.",
 						},
+						"ipv6_cidr": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "IPv6 CIDR of the subnet.",
+						},
 					},
 				},
 			},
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_vpc.go
```diff
@@ -132,6 +132,11 @@ func resourceAviatrixVpc() *schema.Resource {
 							Computed:    true,
 							Description: "Subnet ID.",
 						},
+						"ipv6_cidr": {
```
dnardo, line 135: Since we are updating this schema - do any docs need to change? I'm thinking about `resources/aviatrix_vpc.md`

avx-pkandoi: This is internal to write in tf state file but we are not looking user to configure this. Hence for this doc updates are not required.

rpanzer-aviatrix: I think it still makes sense to document this so that customers can query this property in Terraform.

avx-pkandoi: Thank you @dnardo and @rpanzer-aviatrix to remind it again.
I overlooked it and it was present in document.

I updated the doc now.

rpanzer-aviatrix (approved):

PR #2394, opened 2025-10-15
AVX-69413: Added a data source for DCF trust bundles

avtx-akonnur:
Adding a data source for DCF trust bundle which can be queried by display_name to get the bundle_uuid and other properties that can be used when creating a TLS profile.


tommy-carver (approved):
Can we also update the resource readme? Seems relevant

PR #2405, opened 2025-10-22
AVX-69599: Adding TF for DCF trust bundles

avtx-akonnur:
This adds TF support for DCF trust bundles, the bundle content must be used as file("bundle") and input as a string.

### DIFF
--- a/aviatrix/resource_aviatrix_dcf_trustbundle.go
+++ b/aviatrix/resource_aviatrix_dcf_trustbundle.go
@@ -28,7 +28,6 @@ func resourceAviatrixDCFTrustBundle() *schema.Resource {
 			"bundle_content": {
 				Type:         schema.TypeString,
 				Required:     true,
-				Sensitive:    true,
 				ValidateFunc: goaviatrix.ValidateTrustbundle,
 				Description:  "The CA bundle content in PEM format.",
 			},
--- a/aviatrix/resource_aviatrix_dcf_trustbundle.go
+++ b/aviatrix/resource_aviatrix_dcf_trustbundle.go
@@ -2,6 +2,7 @@ package aviatrix
 
 import (
 	"context"
+	"errors"
 	"strings"
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
@@ -75,7 +76,7 @@ func resourceAviatrixDCFTrustBundleRead(ctx context.Context, d *schema.ResourceD
 
 	trustBundle, err := client.GetDCFTrustBundle(ctx, bundleID)
 	if err != nil {
-		if err == goaviatrix.ErrNotFound {
+		if errors.Is(err, goaviatrix.ErrNotFound) {
 			d.SetId("")
 			return nil
 		}
--- a/aviatrix/resource_aviatrix_dcf_trustbundle.go
+++ b/aviatrix/resource_aviatrix_dcf_trustbundle.go
@@ -78,7 +78,7 @@ func resourceAviatrixDCFTrustBundleRead(ctx context.Context, d *schema.ResourceD
 	if err != nil {
 		if errors.Is(err, goaviatrix.ErrNotFound) {
 			d.SetId("")
-			return nil
+			return diag.Errorf("DCF Trust Bundle not found: %s", err)
 		}
 		return diag.Errorf("failed to read DCF Trust Bundle: %s", err)
 	}
@@ -118,7 +118,7 @@ func resourceAviatrixDCFTrustBundleDelete(ctx context.Context, d *schema.Resourc
 
 	err := client.DeleteDCFTrustBundle(ctx, bundleID)
 	if err != nil {
-		return diag.Errorf("failed to delete DCF Trust Bundle: %v", err)
+		return diag.Errorf("failed to delete DCF Trust Bundle: %s", err)
 	}
 
 	return nil
--- a/aviatrix/resource_aviatrix_dcf_trustbundle.go
+++ b/aviatrix/resource_aviatrix_dcf_trustbundle.go
@@ -29,7 +29,7 @@ func resourceAviatrixDCFTrustBundle() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				Sensitive:    true,
-				ValidateFunc: validation.StringIsNotWhiteSpace,
+				ValidateFunc: goaviatrix.ValidateTrustbundle,
 				Description:  "The CA bundle content in PEM format.",
 			},
 			"bundle_id": {
--- a/aviatrix/resource_aviatrix_dcf_trustbundle.go
+++ b/aviatrix/resource_aviatrix_dcf_trustbundle.go
@@ -1,0 +1,124 @@
+package aviatrix
+
+import (
+	"context"
+	"strings"
+
+	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func resourceAviatrixDCFTrustBundle() *schema.Resource {
+	return &schema.Resource{
+		CreateWithoutTimeout: resourceAviatrixDCFTrustBundleCreate,
+		ReadWithoutTimeout:   resourceAviatrixDCFTrustBundleRead,
+		UpdateWithoutTimeout: resourceAviatrixDCFTrustBundleUpdate,
+		DeleteWithoutTimeout: resourceAviatrixDCFTrustBundleDelete,
+
+		Schema: map[string]*schema.Schema{
+			"display_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
+				Description:  "Display name for the DCF trust bundle.",
+			},
+			"bundle_content": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Sensitive:    true,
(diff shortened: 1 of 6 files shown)


docs/resources/aviatrix_dcf_trustbundle.md
```diff
@@ -0,0 +1,65 @@
...
+subcategory: "Distributed Cloud Firewall"
+layout: "aviatrix"
+page_title: "Aviatrix: aviatrix_dcf_trustbundle"
+description: |-
+  Creates and manages an Aviatrix DCF Trust Bundle
+---
+
+# aviatrix_dcf_trustbundle
+
+The **aviatrix_dcf_trustbundle** resource handles the creation and management of DCF Trust Bundles for verifying origin certificates in Distributed Cloud Firewall Man-in-the-Middle (MITM) inspection. Available as of Provider R3.3.0+.
```
jpeach: Not sure that `R3.3.0+.` is correct?

aviatrix/resource_aviatrix_dcf_trustbundle.go
```diff
@@ -0,0 +1,125 @@
...
+
+		Schema: map[string]*schema.Schema{
+			"display_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
+				Description:  "Display name for the DCF trust bundle.",
+			},
+			"bundle_content": {
+				Type:         schema.TypeString,
```
jpeach, line 29: This is a collection of PEM strings, so should we use TypeList and set the Elem to string?

aviatrix/resource_aviatrix_dcf_trustbundle.go
```diff
@@ -0,0 +1,125 @@
...
+				Required:     true,
+				Sensitive:    true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
```
jpeach: validate func should at least verify we can PEM decode the certificates

goaviatrix/dcf_trustbundle.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// GetDCFTrustBundle retrieves a DCF trust bundle by UUID
+func (c *Client) GetDCFTrustBundle(ctx context.Context, bundleUUID string) (*TrustBundle, error) {
+	endpoint := fmt.Sprintf("dcf/trustbundle/%s", bundleUUID)
+
+	var trustBundle TrustBundle
+	err := c.GetAPIContext25(ctx, &trustBundle, endpoint, nil)
+	if err != nil {
+		if strings.Contains(err.Error(), "does not exist") || strings.Contains(err.Error(), "not found") {
```
jpeach, line 56: This is not great, but the `GetAPIContext25` doesn't really give you much option. do you mind taking an action to improve the error propagation so that callers can get the HTTP status?

aviatrix/resource_aviatrix_dcf_trustbundle.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func resourceAviatrixDCFTrustBundleRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*goaviatrix.Client)
+
+	bundleID := d.Id()
+
+	trustBundle, err := client.GetDCFTrustBundle(ctx, bundleID)
+	if err != nil {
+		if errors.Is(err, goaviatrix.ErrNotFound) {
```
jpeach, line 78: Why is it the right behavior to not error if we don't find the bundle? If the terraform state is current, then it should be there. Silently returning nil is a worry.

aviatrix/resource_aviatrix_dcf_trustbundle.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func resourceAviatrixDCFTrustBundleDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*goaviatrix.Client)
+
+	bundleID := d.Id()
+
+	err := client.DeleteDCFTrustBundle(ctx, bundleID)
+	if err != nil {
+		return diag.Errorf("failed to delete DCF Trust Bundle: %v", err)
```
jpeach: Everywhere else uses `%s`?

aviatrix/resource_aviatrix_dcf_trustbundle.go
avtx-akonnur: The input is taken as one file where the list of certs are passed in. So we have to keep it as a single input of string.

aviatrix/resource_aviatrix_dcf_trustbundle.go
avtx-akonnur: This validation is done on the API end, so I don't think we need to verify this again here.

aviatrix/resource_aviatrix_dcf_trustbundle.go
avtx-akonnur: Updated to return error

goaviatrix/dcf_trustbundle.go
avtx-akonnur: ErrNotFound is what the TF expects to be returned, so that it knows the resource is not found

aviatrix/resource_aviatrix_dcf_trustbundle.go
jpeach: What input are you talking about? The API?

aviatrix/resource_aviatrix_dcf_trustbundle.go
jpeach: We should validate here because it's trivial, gives a better UX and people maintaining the provider can know what is required.

aviatrix/resource_aviatrix_dcf_trustbundle.go
jpeach: If we return error on both paths why do we need special handling here?

docs/resources/aviatrix_dcf_trustbundle.md
jpeach: Why was this resolved with no change?

goaviatrix/dcf_trustbundle.go
jpeach: That wasn't my concern. The issue is that contents of strings is not an API we can rely on. We need to clean this up so that we can inspect the HTTP response status in order to return the designated error.

aviatrix/resource_aviatrix_dcf_trustbundle.go
avtx-akonnur: No I mean how the user passes the file in TF, they are expected to pass it as one file with all the certs inside it. So it is read in TF as one string.

docs/resources/aviatrix_dcf_trustbundle.md
avtx-akonnur: Missed this, removed the line

goaviatrix/dcf_trustbundle.go
avtx-akonnur: You are right, the GetAPIContext25 has the error code as 404 but it doesn't return it. So most of tests just do string check like here, which is not a good idea. Can you file a ticket for that? Or I can do it, but I think it is out of scope of this PR to fix it, as that change will require me to update all the other tests as well to use the new return code. I'd be happy to do it as part of another PR in a different ticket, and we can merge this one in. This test can be addressed along with the others.

aviatrix/resource_aviatrix_dcf_trustbundle.go
avtx-akonnur: I think we talked about this above, if the error is not found or not existing for a trust bundle we should still return 404 which is ErrNotFound, if we return ErrNotFound the unit test passes well, as TF expects this type when checking if the resource exists or not after destroy.

aviatrix/resource_aviatrix_dcf_trustbundle.go
avtx-akonnur: Updated to add validation check

aviatrix/resource_aviatrix_dcf_trustbundle.go
```diff
@@ -0,0 +1,125 @@
...
+			"display_name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringIsNotWhiteSpace,
+				Description:  "Display name for the DCF trust bundle.",
+			},
+			"bundle_content": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Sensitive:    true,
```
jpeach: Public certificates are not sensitive (private keys would be but we dn't have those)

goaviatrix/dcf_trustbundle.go
```diff
@@ -0,0 +1,125 @@
...
+	for _, cert := range certs {
+		if !cert.IsCA {
+			return nil, []error{fmt.Errorf("certificate %q is not a CA", cert.Subject.CommonName)}
+		}
+	}
+	return nil, nil
+}
+
+func ParseCertificates(remain []byte) ([]*x509.Certificate, error) {
+	return parseCertificatesNoBom(remain)
```
jpeach, line 111: Need to strip BOM if it exists.

avtx-akonnur: Updated

aviatrix/resource_aviatrix_dcf_trustbundle.go
avtx-akonnur: makes sense, removed sensitive

goaviatrix/dcf_trustbundle.go
avtx-akonnur: Created this ticket as a follow up item: https://aviatrix.atlassian.net/browse/AVX-70284

jpeach (approved):
LGTM

PR #2409, opened 2025-10-23
AVX-68895: Adding data source for DCF Log profile

avtx-akonnur:
DCF Log profiles, need a data source so the user can query existing resource on the controller to get the profile, they can query the resource by name and get the UUID and other properties of the log profile


docs/data-sources/aviatrix_dcf_log_profile.md
```diff
@@ -0,0 +1,34 @@
...
+
+* `profile_name` - (Required) (String) Name of the Log Profile.
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `profile_id` - (String) The unique identifier for the Log Profile.
+* `session_end` - (Boolean) Toggle to enable logging of session end.
+* `session_start` - (Boolean) Toggle to enable logging of session start.
```
stewartbutler: This doesn't sound right? If it is part of a data source you aren't toggling it, the backend is.

docs/data-sources/aviatrix_dcf_log_profile.md
```diff
@@ -0,0 +1,34 @@
...
+page_title: "Aviatrix: aviatrix_dcf_log_profile"
+description: |-
+  Gets details about a specific DCF log profile.
+---
+
+# aviatrix_dcf_log_profile
+
+The **aviatrix_dcf_log_profile** data source provides details about a specific Distributed Cloud Firewall (DCF) log profile created by the Aviatrix Controller.
+
+## Example Usage
```
stewartbutler, line 19: This should also include an example of how to use this in a real resource. If I understand right this produces the UUID which is used for another resource, so include an example of that

docs/data-sources/aviatrix_dcf_log_profile.md
```diff
@@ -0,0 +1,34 @@
...
+
+The following arguments are supported:
+
+* `profile_name` - (Required) (String) Name of the Log Profile.
+
+## Attribute Reference
+
+In addition to all arguments above, the following attributes are exported:
+
+* `profile_id` - (String) The unique identifier for the Log Profile.
```
stewartbutler: Specify that this UUID can be used as an argument for other DCF APIs

goaviatrix/dcf_log_profile.go
```diff
@@ -0,0 +1,28 @@
...
+
+type LogProfile struct {
+	ProfileName  string `json:"profile_name"`
+	ProfileID    string `json:"profile_id"`
+	SessionEnd   bool   `json:"session_end"`
+	SessionStart bool   `json:"session_start"`
+}
+
+func (c *Client) GetLogProfileByName(ctx context.Context, profileName string) (*LogProfile, error) {
+	endpoint := fmt.Sprintf("dcf/log-profile/name/%s", profileName)
```
stewartbutler: profileName needs escaping via https://pkg.go.dev/net/url#QueryEscape

stewartbutler: This also needs tests - mock up the API endpoint, send through various URLs with malformed profileName entries and make sure they are correctly escaped and handled by the helper.

avtx-akonnur: Added unit tests and using URL escape

aviatrix/data_source_aviatrix_dcf_log_profile_test.go
```diff
@@ -0,0 +1,60 @@
...
+				),
+			},
+		},
+	})
+}
+
+func testAccDataSourceAviatrixDcfLogProfileConfigBasic(profileName string) string {
+	return fmt.Sprintf(`
+data "aviatrix_dcf_log_profile" "test" {
+	profile_name = "%s"
```
stewartbutler: nit: You can use %q instead

stewartbutler (dismissed):

aviatrix/data_source_aviatrix_dcf_log_profile_test.go
avtx-akonnur: updated

stewartbutler (approved):

PR #2423, opened 2025-10-31
AVX-68721 Add support for ipv6 gateway creation with ipv6 subnet cidr

avx-rsalim:
## Summary
This PR adds **IPv6 gateway creation support** using IPv6 subnet CIDRs in Terraform.

---

## Changes
- Added two new fields in the Terraform schema for **spoke** and **transit** resources:
  - `subnet_ipv6_cidr`
  - `ha_subnet_ipv6_cidr`
- These fields are required when creating **IPv6-enabled gateways** (both **HA** and **HPE** types).
- The IPv6 CIDR is appended to the end of the subnet string using the following format:
`ipv4~~zone~~ipv6`

---

## Implementation Details
- For **Azure**, when a `zone` is present, Terraform currently sends the subnet as:
`ipv4~~zone~~`
To maintain compatibility with this format, any trailing `~~` is trimmed before appending `~~ipv6_cidr`.

- The IPv6 CIDR is **required only during gateway creation**.  
During **edit flows** (IPv6 enable/disable), it is **not needed** in the API call.

- **Behavior summary:**
- Changing `subnet_ipv6_cidr` when IPv6 is **enabled** → forces **gateway recreation** (similar to IPv4 subnet changes).
- Changing `subnet_ipv6_cidr` when IPv6 is **disabled** → **no effect** (similar to how `allocate_new_eip` behaves when `private_mode` is enabled).

---

## Tests Performed
- ✅ Verified **spoke** and **transit HA** gateway creation (with and without zones) on **Azure** and **AWS**.  
- ✅ Verified **HPE spoke** and **transit HA** gateway creation (with and without zones) on **Azure** and **AWS**.  
- ✅ Verified **IPv6 enable/disable via edit flow** works correctly.


aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -1555,6 +1642,12 @@ func resourceAviatrixSpokeGatewayRead(d *schema.ResourceData, meta interface{})
 	d.Set("enable_bgp_over_lan", goaviatrix.IsCloudType(gw.CloudType, goaviatrix.AzureArmRelatedCloudTypes) && gw.EnableBgpOverLan)
 	d.Set("enable_ipv6", gw.EnableIPv6)
 	d.Set("insertion_gateway", gw.InsertionGateway)
+	// set ipv6 subnet cidr if ipv6 is enabled on the gateway
+	if gw.EnableIPv6 {
```
avx-pkandoi: nit:
Maybe we don't need this if condition since gw.SubnetIPv6Cidr will be bydefault ""

avx-pkandoi (dismissed):

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -85,6 +87,17 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 				ValidateFunc: validation.IsCIDR,
 				Description:  "Public Subnet Info.",
 			},
+			"subnet_ipv6_cidr": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateIPv6CIDR,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
avx-pkandoi: Maybe we can put this check in customizeDiff?

avx-rsalim: customizeDiff is resouce level and mostly deals with validation and forcenew (Not with diff suppression)

DiffSuppressFunc works at field level and is designed to supress feild if required

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: addressed

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -85,6 +87,17 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 				ValidateFunc: validation.IsCIDR,
 				Description:  "Public Subnet Info.",
 			},
+			"subnet_ipv6_cidr": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateIPv6CIDR,
+				DiffSuppressFunc: func(_, _, _ string, d *schema.ResourceData) bool {
```
sairane10, line 95: Why do you need a diffSuppress here?
It's a computed field so it should not throw a diff.

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -121,6 +134,17 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 				ValidateFunc: validation.IsCIDR,
 				Description:  "HA Subnet. Required if enabling HA for AWS/AWSGov/AWSChina/Azure/AzureChina/OCI/Alibaba Cloud. Optional if enabling HA for GCP.",
 			},
+			"ha_subnet_ipv6_cidr": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ValidateFunc: validateIPv6CIDR,
+				DiffSuppressFunc: func(_, _, _ string, d *schema.ResourceData) bool {
```
sairane10, line 142: Same as above, computed diff don't need diff suppress

aviatrix/resource_aviatrix_transit_gateway.go
```diff
@@ -966,6 +990,18 @@ func resourceAviatrixTransitGateway() *schema.Resource {
 	}
 }
 
+func resourceAviatrixTransitGatewayCustomizeDiff(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
```
sairane10, line 993: why do you need a customize diff here when you have already added diffSuppress to the fields?

aviatrix/utils_test.go
```diff
@@ -223,6 +223,104 @@ func TestValidateCIDRRule(t *testing.T) {
 	}
 }
 
+func TestValidateIPv6CIDR(t *testing.T) {
```
sairane10, line 226: Can we also add unit tests for the new fields you added in the spoke and transit file?

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: the field is computed only in the edit/modify flow , 
If we are creating a Ipv6 gateway , Ipv6 CIDR is mandatory 
```
The IPv6 CIDR is required only during gateway creation.
During edit flows (IPv6 enable/disable), it is not needed in the API call.

Behavior summary:

Changing subnet_ipv6_cidr when IPv6 is enabled → forces gateway recreation (similar to IPv4 subnet changes).

Changing subnet_ipv6_cidr when IPv6 is disabled → no effect (similar to how allocate_new_eip behaves when private_mode is enabled).
```

avx-rsalim: We give Optional + computed 
as we need it during gateway creation , 
but during 
ipv6 edit it is computed

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: same

aviatrix/resource_aviatrix_transit_gateway.go
avx-rsalim: we need it , such that if customer changes the subnet_ipv6_cidr when ipv6 is enabled , 
we should recreate the gateway , 
similar to how if we change `subnet` which contains the ipv4 subnet is changed , we recreate the gateway , basically enoforce ForceNew if subnet_ipv6_cidr is changed when ipv6 is enabled

aviatrix/utils_test.go
avx-rsalim: sure , will add them ,

avx-rsalim: done

saileerane (approved):

sairane10 (approved):

aviatrix/utils.go
```diff
@@ -20,6 +20,28 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
 
+func validateIPv6CIDR(i interface{}, k string) (warnings []string, errors []error) {
```
jpeach, line 23:
```suggestion
func validateIPv6CIDR(i any, k string) (warnings []string, errors []error) {

```

aviatrix/utils.go
```diff
@@ -20,6 +20,28 @@ import (
...
+	v, ok := i.(string)
+	if !ok {
+		errors = append(errors, fmt.Errorf("expected type of %q to be string", k))
+		return warnings, errors
+	}
+
+	ip, _, err := net.ParseCIDR(v)
```
jpeach, line 30: IIUC, this is meant to be a network CIDR, so we should validate that the ipmask result has a Size() of more than 0?

i.e like validation.IsCIDRNetwork does.

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -85,6 +87,17 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 				ValidateFunc: validation.IsCIDR,
```
jpeach, line 87: Should this be validation.IsCIDRNetwork?

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -26,6 +26,8 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 			State: schema.ImportStatePassthrough,
 		},
 
+		CustomizeDiff: resourceAviatrixSpokeGatewayCustomizeDiff,
```
jpeach, line 29: What is this doing? Please explain in a block comment.

aviatrix/resource_aviatrix_spoke_gateway.go
jpeach, line 95: I'm also confused by this, so I  think you need a big block comment here. what I expected to see was code that validates the IPv6-related values are not set unless "ipv6_enabled" is true.

Perhaps the schema `RequiredWith` and `ConflictsWith` fields are helpful here?

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -85,6 +87,17 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
 				ValidateFunc: validation.IsCIDR,
 				Description:  "Public Subnet Info.",
 			},
+			"subnet_ipv6_cidr": {
```
jpeach, line 90: Does this field conflict with the "subnet" field?

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -85,6 +87,17 @@ func resourceAviatrixSpokeGateway() *schema.Resource {
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
```
jpeach, line 93: Why are the ipv6 subnet fields computed when the ipv4 ones are not?

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -1085,6 +1128,25 @@ func resourceAviatrixSpokeGatewayCreate(d *schema.ResourceData, meta interface{}
 		gateway.InsertionGateway = true
 	}
 
+	enableIpv6 := d.Get("enable_ipv6").(bool)
+	if enableIpv6 {
```
jpeach, line 1132:
```suggestion
	d.Get("enable_ipv6").(bool)
	if  d.Get("enable_ipv6").(bool) {
```

avx-pkandoi (approved):

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -1085,6 +1128,25 @@ func resourceAviatrixSpokeGatewayCreate(d *schema.ResourceData, meta interface{}
...
+		if !goaviatrix.IsCloudType(gateway.CloudType, goaviatrix.AzureArmRelatedCloudTypes|goaviatrix.AWSRelatedCloudTypes) {
```
jpeach, line 1133: Please add a IPv6SupportedOnCloudType function and use it to replace all instances of this check.

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -1191,6 +1253,18 @@ func resourceAviatrixSpokeGatewayCreate(d *schema.ResourceData, meta interface{}
 			spokeHaGw.InsertionGateway = true
 		}
 
+		enableIpv6 := d.Get("enable_ipv6").(bool)
+		if enableIpv6 {
```
jpeach, line 1257:
```suggestion
		if d.Get("enable_ipv6").(bool) {
```

jpeach: And please update all the other places that do this.

aviatrix/resource_aviatrix_spoke_gateway.go
```diff
@@ -2260,6 +2337,18 @@ func resourceAviatrixSpokeGatewayUpdate(d *schema.ResourceData, meta interface{}
 			spokeHaGw.InsertionGateway = true
 		}
 
+		enableIpv6 := d.Get("enable_ipv6").(bool)
```
jpeach, line 2340: The ghastly conditional that gets you into this block checks whether `ha_subnet` changed. Do you need to update that to also check for `ha_subnet_ipv6_cidr` changes?

jpeach (commented):
Since this merged, I won't continue the review, but please make a cleanup pass. There are a number of issue here.

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: No , subnet only takes the IPV4 network CIDR 
while subnet_ipv6_cidr deals with ipv6 cidr

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: will add it in the cleanup PR

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: Ideally it should be IsCIDRNetwork , bacause we are dealing with network block rather than hosts , 
Bit it looks we dont use it now in our terraform database for any subnet fields

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: Ipv6 fields are mandatary when we create a Ipv6 gateway 
(For normal gateways where the dual stack subnet exists ), 
and HPE gateways where we create the dual stack subnet using the ipv4 subnet and ipv6 subnet provided
but 

in edit flow for ipv6,
we do not need any cidrs, (we take the CIDR from the subnet if it has ipv6 cidr else we error out )

For context , 

For HPE , since we created the subnet in the first place, we assign the first free ipv6 cidr in edit flow , we do not do the same for normal gateways as customers might have created the subnet and do not want add any CIDR there (This is as per PM ask)

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: The intent here is , if customer changes `subnet_ipv6_cidr` when ipv6 is disabled , to not generate a diff , 

This case arises when customer 
1. created gateway with enable_ipv6 and subnet_ipv6_cidr
2. disabled ipv6 with enable_ipv6=false
3.customer for some reason changed the value of `subnet_ipv6_cidr`
with the help of DiffSuppressFunc , a diff will not be calculated

 If 'subnet_ipv6_cidr' is changed when 'enable_ipv6' is True ,
 CustomizeDiff: resourceAviatrixSpokeGatewayCustomizeDiff,
  will make sure the gateway is recreated ,as user intends to change the subnet where gateway is hosted

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: will address in  clean up PR

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: will address in clean up PR

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: will address in clean up PR

aviatrix/resource_aviatrix_spoke_gateway.go
avx-rsalim: Yes , we would need it , 
will address in clean up PR

aviatrix/utils.go
avx-rsalim: will address in clean up PR

avx-rsalim (commented):
Have raised a new Task for addressing these comments.
AVX-70667, 
have added some details for some comments which were of type query

aviatrix/utils.go
avx-rsalim: will address in clean up PR

PR #2427, opened 2025-11-05
AVX-68013: client: Check HTTP status codes not strings

DolceTriade:
Don't check random strings for errors and instead check the actual HTTP status codes for more accurate checking.


rpanzer-aviatrix (approved):
Looks good to me.
I am not sure I understand the logic, but it just documents how it works today.

PR #2432, opened 2025-11-07
AVX-70561: enable_firenet_for_edge knob fails to trigger API call

cjenn-aviatrix:
This pull request adds support for updating the `enable_firenet_for_edge` property in the Edge Spoke Transit Attachment resource and introduces comprehensive tests to ensure the update logic works as expected. The main focus is on enabling or disabling FireNet for edge attachments through the update operation.

**Enhancements to Edge Spoke Transit Attachment resource:**

* Added logic to `resourceAviatrixEdgeSpokeTransitAttachmentUpdate` to detect changes to the `enable_firenet_for_edge` property and send the appropriate update API request when this property is modified.

**Testing improvements:**

* Introduced a new test, `TestResourceAviatrixEdgeSpokeTransitAttachmentUpdate_enableFirenetForEdge`, to verify that changes to the `enable_firenet_for_edge` property are correctly detected and that the update form is constructed as expected for various scenarios (change from false to true, true to false, and unchanged).


aviatrix/resource_aviatrix_edge_spoke_transit_attachment.go
```diff
@@ -422,6 +422,21 @@ func resourceAviatrixEdgeSpokeTransitAttachmentUpdate(ctx context.Context, d *sc
...
 	}
 
+	if d.HasChange("enable_firenet_for_edge") {
+		form := map[string]interface{}{
+			"CID":                     client.CID,
+			"action":                  "edit_inter_transit_gateway_peering",
+			"gateway1":                spokeGwName,
+			"gateway2":                transitGwName,
+			"enable_firenet_for_edge": d.Get("enable_firenet_for_edge").(bool),
+		}
```
cjenn-aviatrix, line 432: Two options that I can see here, the first is to add EnableFirenetForEdge to the TransitGatewayPeering struct in goaviatrix/transit_gateway_peering.go, and this one. 
I didn't chose the former because this knob does not apply to other Transit Peering types, only EaS to Transit.

stewartbutler (approved):

aviatrix/resource_aviatrix_edge_spoke_transit_attachment.go
stewartbutler: I don't know enough about the domain to critique it -- may want a team member to double check it. Looks fine quality-wise, so I've approved it.

cjenn-aviatrix: @sairane10 could you comment on this design decision before I merge? Thanks

sairane10: Looks good to me. I would have moved the api call to the attachment.go file where all other api calls are present

sairane10 (approved):

PR #2434, opened 2025-11-10
AVX-69637 Terraform ipv6 cleanup

avx-rsalim:
Added some code clean up for ipv6 terraform and also addressed some pending comments from https://github.com/AviatrixSystems/terraform-provider-aviatrix/pull/2423


jpeach (approved):
Thanks!

PR #2436, opened 2025-11-11
AVX-70625: Adding a custom diff compare function to compare content of DCF Trust bundles

avtx-akonnur:
DCF Trustbundle API, expects a string as input, but returns a list of certs, the certs when joined for comparison, might not have the exact space or newline characters as the input, so we need a custom function that normalizes both input before comparing for actual change in cert value.


kasunkarunaratne (dismissed):
lgtm

jpeach (approved):

PR #2439, opened 2025-11-12
AVX-70603 IPS Profile VPC update

xc-6:
- Update the release notes for applying IPS profiles to VPC. (only one IPS profile per VPC is supported in controller version 8.2)

- Update the code to handle destroying the IPS-Profile-VPC resource. (If the user manually deletes the Gateway or VPC before disassociating the IPS profile from it, Terraform can still destroy the resource.)

### Test
```
Plan: 0 to add, 0 to change, 2 to destroy.
Do you really want to destroy all resources?
  Terraform will destroy all your managed infrastructure, as shown above.
  There is no undo. Only 'yes' will be accepted to confirm.
  Enter a value: yes
aviatrix_dcf_ips_profile.ips_profile: Destroying... [id=142ca577-fe67-4f42-9bc2-18ee603a23d5]
aviatrix_vpc.vpc_1: Destroying... [id=tf-ips-new-vpc3]
aviatrix_dcf_ips_profile.ips_profile: Destruction complete after 0s
aviatrix_vpc.vpc_1: Destruction complete after 6s
Destroy complete! Resources: 2 destroyed.
```

### DIFF
--- a/aviatrix/resource_aviatrix_dcf_ips_profile_vpc.go
+++ b/aviatrix/resource_aviatrix_dcf_ips_profile_vpc.go
@@ -4,6 +4,7 @@ package aviatrix
 import (
 	"context"
 	"errors"
+	"strings"
 
 	"github.com/AviatrixSystems/terraform-provider-aviatrix/v3/goaviatrix"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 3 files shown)


aviatrix/resource_aviatrix_dcf_ips_profile_vpc.go
```diff
@@ -100,3 +106,11 @@ func resourceAviatrixDCFIpsProfileVpcDelete(ctx context.Context, d *schema.Resou
 	d.SetId("")
 	return nil
 }
+
+func isVpcNotFoundError(err error) bool {
```
jpeach: We should use a designated AVXERR here if we need to know that something specific happened. Make a follow-up jira to fix and add a note here with the ID.

jpeach (dismissed):

jpeach (approved):
Thanks for the AVXERR update 👍

PR #2443, opened 2025-11-17
AVX-70621: Adding metadata fields for TF created policy objects.

avtx-akonnur:
The TF when calling the MWP APIs will now attach a metadata field so the UI can know that this object was created by TF.


kspoelstra-aviatrix (approved):
LGTM, in the future we need to think about what we do when we change the policy blocks, if any additional metadata is needed or we need to ensure that the metadata is still there.

stewartbutler (approved):

PR #2470, opened 2026-01-05
AVX-72203 Updating the spoke HA gw name when not provided by the user

sairane10:
- Adding a new post async api to poll the HA GW name from the response. This will create a a correct state file entry for the spoke HA GW when the gateway name is not provided by the user

```
# Primary Spoke Gateway Creation on Zone 1
resource "aviatrix_spoke_gateway" "Aws-Gwy-VPC-1-AZa" {
  cloud_type            = 1
  account_name          = "sailee-aws"
  gw_name               = "aws-vpc-1-gw-1"
  vpc_id                = aviatrix_vpc.aws_vpc_1.vpc_id
  vpc_reg               = "us-west-1"
  gw_size               = "t3.medium"
  subnet                = aviatrix_vpc.aws_vpc_1.public_subnets[0].cidr  # Use actual subnet
  insane_mode           = false
  manage_ha_gateway     = false
  single_az_ha          = false
}

# HA Spoke Gateway Creation on Zone 2
resource "aviatrix_spoke_ha_gateway" "Aws-Gwy-VPC-1-AZb" {
  primary_gw_name = "aws-vpc-1-gw-1"
  subnet = "10.0.32.0/20" # Deploy on the second public subnet
}
```

### DIFF
--- a/goaviatrix/client.go
+++ b/goaviatrix/client.go
@@ -407,11 +407,12 @@ func (c *Client) PostAsyncAPIHaGw(action string, i interface{}, checkFunc CheckA
 	return c.PostAsyncAPIContextHaGw(context.Background(), action, i, checkFunc)
 }
 
+//nolint:cyclop,funlen
 func (c *Client) PostAsyncAPIContextHaGw(ctx context.Context, action string, i interface{}, checkFunc CheckAPIResponseFunc) (string, error) {
 	log.Printf("[DEBUG] Post AsyncAPI HaGw %s: %v", action, i)
 	resp, err := c.PostContext(ctx, c.baseURL, i)
 	if err != nil {
-		return "", fmt.Errorf("HTTP POST %s failed: %v", action, err)
+		return "", fmt.Errorf("HTTP POST %s failed: %w", action, err)
 	}
 
 	// Response struct that includes ha_gw_name
@@ -428,7 +429,7 @@ func (c *Client) PostAsyncAPIContextHaGw(ctx context.Context, action string, i i
 	bodyString := buf.String()
 	bodyIoCopy := strings.NewReader(bodyString)
 	if err = json.NewDecoder(bodyIoCopy).Decode(&data); err != nil {
-		return "", fmt.Errorf("Json Decode %s failed %v\n Body: %s", action, err, bodyString)
+		return "", fmt.Errorf("json decode %s failed: %w\n Body: %s", action, err, bodyString)
 	}
 	if !data.Return {
 		return "", fmt.Errorf("rest API %s POST failed to initiate async action: %s", action, data.Reason)
@@ -455,13 +456,14 @@ func (c *Client) PostAsyncAPIContextHaGw(ctx context.Context, action string, i i
 		}
 		buf = new(bytes.Buffer)
 		buf.ReadFrom(resp.Body)
+		resp.Body.Close()
 		err = json.Unmarshal(buf.Bytes(), &data)
 		if err != nil {
 			if resp.StatusCode == http.StatusBadGateway || resp.StatusCode == http.StatusServiceUnavailable {
 				time.Sleep(sleepDuration)
 				continue
 			}
-			return "", fmt.Errorf("decode check_task_status failed: %v\n Body: %s", err, buf.String())
+			return "", fmt.Errorf("decode check_task_status failed: %w\n Body: %s", err, buf.String())
 		}
 
 		// Capture ha_gw_name from polling response if present
--- a/goaviatrix/client.go
+++ b/goaviatrix/client.go
@@ -328,7 +328,8 @@ func (c *Client) PostFileAPIContext(ctx context.Context, params map[string]strin
 	return checkAPIResp(resp, params["action"], checkFunc)
 }
 
-// StartResponseHook is called with the raw start response, allowing callers to extract custom fields.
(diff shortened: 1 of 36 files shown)


goaviatrix/client.go
```diff
@@ -402,6 +402,92 @@ func (c *Client) PostAsyncAPIContext(ctx context.Context, action string, i inter
 	return fmt.Errorf("waited %s but upgrade never finished. Please manually verify the upgrade status", maxPoll*sleepDuration)
 }
 
+// PostAsyncAPIHaGw is similar to PostAsyncAPI but also returns the ha_gw_name from the response
```
dnardo: seems strange that we would have a one-off function like this just for the HA GW?  why is that special?  I feel this shouldn't be necessary

goaviatrix/client.go
```diff
@@ -328,20 +328,26 @@ func (c *Client) PostFileAPIContext(ctx context.Context, params map[string]strin
 	return checkAPIResp(resp, params["action"], checkFunc)
 }
 
-func (c *Client) PostAsyncAPI(action string, i interface{}, checkFunc CheckAPIResponseFunc) error {
+// PostAsyncAPI submits an async request and waits for completion.
+// Returns (ha_gw_name, error) - ha_gw_name is populated if the response contains it.
+func (c *Client) PostAsyncAPI(action string, i interface{}, checkFunc CheckAPIResponseFunc) (string, error) {
 	return c.PostAsyncAPIContext(context.Background(), action, i, checkFunc)
 }
```
dnardo, line 362: `PostAsyncAPI` / `PostAsyncAPIContext` read to me like generic async helpers (start a task, poll until completion, then run checkFunc). This change makes them implicitly gateway/HA-specific by assuming the start response includes `ha_gw_name` and then threading that through polling.

A couple concerns/questions:

1) Is `ha_gw_name` guaranteed to be present for all actions that use this helper?
If it’s not guaranteed, the generic helper is now brittle.

2) If the intent is that this helper is only for HA gateways, then I agree the implementation might be fine, but the current name is misleading. Something like `PostAsyncGatewayContext` (or your choosing) would make that coupling explicit.

3) If the intent is to keep `PostAsyncAPIContext` generic, I think we should avoid baking `HaGwName` into its response struct. 

Instead. we should keep the helper generic, and let callers customize the poll payload / decode behavior (e.g. via an injected pollPayload(requestID) func or option hook), or add a gateway specific wrapper that captures `ha_gw_name` and calls the generic helper.

Yes that is more work but its the right way if this is truly generic.

goaviatrix/client.go
sairane10: Updated the existing async api function to return the gateway name.

goaviatrix/client.go
sairane10: Updated the async api with opt's and changes we spoke about

goaviatrix/client.go
dnardo, line 419: Right above this line I think you also want
`form := cfg.pollPayload(requestID)`

So lets move it into the loop here. (instead of above where you have it)

goaviatrix/client.go
```diff
@@ -328,41 +328,90 @@ func (c *Client) PostFileAPIContext(ctx context.Context, params map[string]strin
...
+type asyncCfg struct {
+	onStartResponse StartResponseHook
+	pollPayload     AsyncPollPayloadFunc
+}
+
+// AsyncOption configures async API behavior.
+type AsyncOption func(*asyncCfg)
+
+// WithStartResponseHook sets a hook to be called with the raw response.
+// The hook is called on both the initial response and each poll response,
```
dnardo, line 348: I didn't realize it was called on initial and each poll - lets just call this `ResponseHook` then, my bad.

goaviatrix/client.go
```diff
@@ -328,41 +328,90 @@ func (c *Client) PostFileAPIContext(ctx context.Context, params map[string]strin
...
+		},
+	}
+	for _, o := range opts {
+		o(&cfg)
+	}
+
 	log.Printf("[DEBUG] Post AsyncAPI %s: %v", action, i)
 	resp, err := c.PostContext(ctx, c.baseURL, i)
 	if err != nil {
 		return fmt.Errorf("HTTP POST %s failed: %v", action, err)
```
dnardo: not your change but lets change this to:
`return fmt.Errorf("HTTP POST %s failed: %w", action, err)`

goaviatrix/client.go
```diff
@@ -375,16 +424,26 @@ func (c *Client) PostAsyncAPIContext(ctx context.Context, action string, i inter
 		}
 		buf = new(bytes.Buffer)
 		buf.ReadFrom(resp.Body)
```
dnardo, line 426: Lets close the body here also after the read:
resp.Body.Close()

goaviatrix/client.go
sairane10: updated

goaviatrix/client.go
sairane10: done

goaviatrix/client.go
sairane10: updated

goaviatrix/client.go
sairane10: added this change

dnardo (approved):

